PR #3, opened 2018-04-19
add willingc to owners/reviewers list

willingc:


jlewi:
/approve

Although it looks like tide isn't enabled #4

PR #6, opened 2018-04-27
optimize css

lqj679ssn:
Some css is not valid like: `xx: 1rem/4` or `overflow: none`, and some can be optimized.
Now fixed these small bugs.


ankushagarwal:
/lgtm
/approve

Thanks @lqj679ssn

PR #8, opened 2018-05-01
optimize css - make pagination looks nicer

lqj679ssn:
Former pagination heights are not consistent. Now make it looks nicer.


gaocegege:
Thanks for your contribution!

Some people in the community are in Kubecon Europe 2018 thus the review progress may be a little slow. Sorry for that.

lqj679ssn:
@gaocegege 233333 thanks for telling...

gaocegege:
https://5ae92b70dd6a542d2c48fc41--kubeflow-website-dev.netlify.com/

This is the preview for this PR.

And the preview of master is here: https://5ae92a6911b73b6792f76812--kubeflow-website-dev.netlify.com/

I am not sure if we want it.

willingc:
HI @lqj679ssn, Thanks for the PR. I'm not a CSS expert so I'm not sure what in this PR caused the GitHub button size to change. It would be great to keep the GitHub button at its original (master) size.

@gaocegege Thanks for the links to PR and master rendering. Very helpful.

lqj679ssn:
@willingc @gaocegege 
As you can see, the former one is:
image
The height of 2 buttons are different, and `page 1 of 2` are not in the vertical middle of the button.

gaocegege:
@lqj679ssn 

I prefer your change on the button, but it seems that the GitHub link is also changed:

screenshot from 2018-05-02 14-49-04

themes/kube/static/css/kube.css
```diff
@@ -1012,24 +1015,30 @@ button:hover,
...
+  padding: 0 20px;
   border-radius: 3px; }
 
 .button.big {
   font-size: 17px;
   min-height: 48px;
-  padding: 13px 24px;
+  line-height: 48px;
+  vertical-align: middle;
+  padding: 24px;
```
gaocegege: Maybe we do not need to change the big button, WDYT

lqj679ssn: Oh sorry for that, and thanks for telling!
The former `button small` CSS used `padding` feature and managed to make text in the vertical middle, same as `button large` and `button big`, but that's not a good way as you can see in my comment. So I change those CSS and use `line-height` & `vertical-align` feature. But I forgot to make `padding-top` & `padding-small` as `0`.
Now I fixed this bug.

Issue #7, opened 2018-04-28
[discussion] Setup Prow


gaocegege:
Maybe we could refer to https://github.com/kubernetes/website for details. I think they do not setup presubmit or postsubmit since it is a website repo.

Should we remove them from test-infra?

/cc @jlewi @ewilderj

jlewi:
/priority p1
/assign jlewi
/unassign @gaocegege 
@gaocegege tried but asked me to take over since he's hitting some test failures with empty prow_config.yaml.

jlewi:
Fixed by #11 
/close

PR #9, opened 2018-05-02
Netlify: Add support

gaocegege:
- Added `netlify.toml` to auto-config netlify app
- Ignore public (If we decided to use netlify for prod)

Preview in Netlify is: https://kubeflow-website-dev.netlify.com/


gaocegege:
/hold

We need to come to an agreement in #5 before the PR.

jlewi:
Is the test failure just because we don't have any tests? Would adding an empty prow_config.yaml fix that?

gaocegege:
ref https://github.com/kubeflow/website/issues/7

I think we do not need to presubmit and postsublime. WDYT?

Of course, Adding a empty prow.yaml also works for me.

jlewi:
For consistency with other repositories I'd prefer to have an empty prow file rather than removing the prow jobs all together.

gaocegege:
SGTM. I will take a look.

jlewi:
@gaocegege Can you sync? We added a prow_config.yaml file so I think the tests should pass.

gaocegege:
Yeah, I will rebase the master ASAP.

gaocegege:
@jlewi 

Please see the deploy preview https://deploy-preview-9--competent-brattain-de2d6d.netlify.com/

jlewi:
/hold cancel
I think we've agreed to go with Netify.

PR #10, opened 2018-05-05
prow_config.yaml: Add

gaocegege:
Signed-off-by: Ce Gao <gaoce@caicloud.io>

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -1,0 +1,1 @@
+workflows:


gaocegege:
/assign @jlewi 

I think ksonnet is really a big overhead for maintenance. :worried:

jlewi:
Is the workflow actually testing anything relevant for this repository? What if in prow_config.yaml we just make workflows an empty array? Then I think we won't run any Argo workflows but the test should be reported as a success; so we don't need to check in a ksonnet app.

gaocegege:
Do you mean

```
workflows:
```

I tried to leave the file empty but the test failed.

jlewi:
@gaocegege correct. Make the array workflows in prow_config.yaml an empty array. We loop over that array [here](https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py#L95) so if workflows is empty it should be a nullop.

So if there are no workflows the test should just be reported as successful.

gaocegege:
I will update it, thanks!

gaocegege:
@jlewi It seems that it does not work for us.

Could I unassign myself and assign you to setup the prow for the repo?  You are familiar with our test infra :smile:

jlewi:
Sure

jlewi:
Fixed by #11 PTAL.

PR #28, opened 2018-06-05
Placeholders for Getting Started documents

abhi-g:
Added placeholders for Getting Started documents. The actual content in those needs to be filled in or updated. 

Some more Table to Contents layout cleanup.


abhi-g:
screen shot 2018-06-05 at 4 36 52 pm

abhi-g:
- All the deleted files are the boilerplate content that I have removed.

 - You might want to go through all the added files. content/docs/ are the placeholder docs. Couple of them have content copied over from kubeflow readmes.

- Modified the css for TOC to remove numbering and compact the listing.

- Changed the toclocator plugin to respond to window resize rather than scroll.

pdmack:
@abhi-g actually I'm unable to navigate to the new content on local serve. `/docs/` gives 404.

abhi-g:
@pdmack I left the new content in "Draft" mode, therefore its not yet available via the server. If we think its okay to have the content be accessible, I can switch the draft mode off. Also, until this PR gets merged, it would not be accesible on the site.
If you are trying locally, run "hugo server -D" to view the Draft pages.

Issue #15, opened 2018-05-24
Define and update the top nav elements on the landing page.


abhi-g:
Need to identify the top level navigational items and update the page accordingly. Here is a potential suggestion for the list items:
  - Getting Started
  - Deployment
  - Examples
  - Community
  - News/Blog
  - FAQ/Getting Help
Possibly use a document to define the sub-entries for each of those.

Issue #14, opened 2018-05-24
Get the design/layout approved and pushed to the repo.


abhi-g:
Identify any remaining blockers on the landing page design template. If all approvals are there, push it to the website repo.

PR #38, opened 2018-06-13
Incorporate new Kubeflow site design into the hugo theme

abhi-g:
- Created a new theme (kf) out of the design templates for the site.
- Updated templates according to the new design
- Some fixes for navigational links
- restructured to have all getting started documents in a section of their own.


abhi-g:
resolve #14

abhi-g:
fixes #15

PR #63, opened 2018-06-22
Bump ksonnet reference to 0.11.0

pdmack:
Should be OK to point users at the latest.
<fingers-crossed>

/assign @abhi-g 
/assign @ewilderj

### DIFF
--- a/content/docs/user_guide.md
+++ b/content/docs/user_guide.md
@@ -14,7 +14,7 @@ For an end to end example illustrating in details how to deploy kubeflow and run
 
 ## Requirements
  * Kubernetes >= 1.8 [see here](https://github.com/kubeflow/tf-operator#requirements)
- * ksonnet version [0.9.2](https://ksonnet.io/#get-started). (See [below](#why-kubeflow-uses-ksonnet) for an explanation of why we use ksonnet)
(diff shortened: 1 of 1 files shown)


content/docs/user_guide.md
```diff
@@ -14,7 +14,7 @@ For an end to end example illustrating in details how to deploy kubeflow and run
 
 ## Requirements
  * Kubernetes >= 1.8 [see here](https://github.com/kubeflow/tf-operator#requirements)
- * ksonnet version [0.9.2](https://ksonnet.io/#get-started). (See [below](#why-kubeflow-uses-ksonnet) for an explanation of why we use ksonnet)
+ * ksonnet version [0.11.0](https://ksonnet.io/#get-started). (See [below](#why-kubeflow-uses-ksonnet) for an explanation of why we use ksonnet)
```
abhi-g: @kunmingg Can you please take a look. 
We could also say ksonnet >= 0.x.x depending on what's the bare minimum required. If it's not critical, I am okay with switching to 0.11.0

kunmingg: 0.11.0 should be good, thanks!

PR #64, opened 2018-06-22
Add instructions for copying the ksonnet application to your machine.

jlewi:


abhi-g:
@jlewi Can you please also add a line to route users to the [User guide](https://www.kubeflow.org/docs/user_guide/#bringing-up-a-jupyter-notebook)  after the basic setup is done.

jlewi:
/assign @ankushagarwal PTAL since @abhi-g is now on vacation.

Do you know the proper way to add links to the user_guide.md so that they render properly (I just added a full link to the HTTP URL but I suspect there is a way to use a relative link).

PR #66, opened 2018-06-22
Fix some markdown issues in the UG

pdmack:
/assign @abhi-g 
/assign @ewilderj

### DIFF
--- a/content/docs/user_guide.md
+++ b/content/docs/user_guide.md
@@ -34,7 +34,7 @@ Install the Kubeflow packages into your application.
 ```
 # For a list of releases see:
 # https://github.com/kubeflow/kubeflow/releases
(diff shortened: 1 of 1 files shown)


content/docs/user_guide.md
```diff
@@ -514,26 +514,28 @@ ks delete ${KF_ENV} -c ${MNIST_JOB_NAME}
 
 ## Advanced Customization
 
+### Persistent Disks
+
 * Often times data scientists require a POSIX compliant filesystem
```
abhi-g: Can you please also reformat the first couple of bullets before line 524 to be a simple paragraph, rather than a bulleted list.

pdmack: Yep, that felt weird to me too. ;-)

PR #74, opened 2018-06-29
Added Kubeflow events calendar.

amsaha:


jlewi:
/ok-to-test

Thank you so much.

content/blog/future_events.md
```diff
@@ -0,0 +1,26 @@
++++
```
jlewi: I would suggest moving the content to content/about as opposed to blogs.

I'm not sure blog post is a good fit since we will want to continually update this.

I would also suggest having a single page and not distinguishing between future and past events but just listing events in reverse chronological order. 

This way we won't have to update the page every time an event actually happens; so it will just reduce the toil of keeping it up to date.

content/blog/future_events.md
```diff
@@ -0,0 +1,26 @@
...
+description = "Events where Kubeflow will be presented"
+weight = 10
+toc = true
+bref= ""
++++
+
+List of events where talks and workshops about Kubeflow will be presented. Please note that
+this might not be an exhaustive list.
+
+* [Paris Container Day, Paris](https://www.nginx.com/resources/events/paris-container-day/), 26 June, 2018
```
jlewi: Can you add links to Kubecon Shanghai and Kubecon Seattle?

amsaha:
I am going to include all the feedback later this week and do a new pull request. Busy with some other pressing stuff.

jlewi:
@amsaha Thanks. You could also update this PR rather than creating a new one.

content/blog/future_events.md
amsaha: This would mean that a top level "Content" page needs to be created, right? Currently, there is none.

amsaha: @jlewi The blog page currently has the title "Blogs, Talks, and Videos". We should probably change that to "Blogs and Events" and under that an Events page. "about" does not seem to be an appropriate title.

amsaha:
/hold

I am merging with upstream changes.

content/blog/future_events.md
amsaha: @jlewi Got your comment. My fork was outdated :-) A lot has changed :-)

amsaha:
/assign @jlewi. 
The changes have been made and PR updated. Please review.

screen shot 2018-07-05 at 8 13 38 pm

PR #86, opened 2018-07-12
Update the docs for v0.2.1

jlewi:
* The getting started process now uses deploy scripts and
  doesn't rely on bootstrapper.


jlewi:
There is an indent error in the Create Kubeflow deployment section that I can't figure out how to fix.
https://deploy-preview-86--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-gke/

indenterror

PR #87, opened 2018-07-13
Fixing some typos and formatting on the GKE instructions page

richardsliu:
Fixed a few typos and formatting issues in the Kubeflow GKE instructions. Also added some missing steps during credentials creation.

### DIFF
--- a/content/docs/started/getting-started-gke.md
+++ b/content/docs/started/getting-started-gke.md
@@ -17,7 +17,7 @@ Instructions for optimizing and deploying Kubeflow on GKE.
 Running on Kubeflow on GKE comes with the following advantages:
 
   * We use [Google Cloud Deployment Manager](https://cloud.google.com/deployment-manager/docs/) to 
-    declaratively manage all non K8s resources (incuding the GKE cluster), which is easy to customize for your particular use case
+    declaratively manage all non K8s resources (including the GKE cluster), which is easy to customize for your particular use case
   * You can take advantage of GKE autoscaling to scale your cluster horizontally and vertically
     to meet the demands of ML workloads with large resource requirements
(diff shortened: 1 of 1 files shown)


richardsliu:
@jlewi Please take a look.

jlewi:
Is this ready for review? "/assign" it to someone when it is.
Looks like you are removing some of the code boxes; is that intentional?

richardsliu:
I removed some of the newlines - I believe these cause Markdown to treat the lines in between as <p>, which is causing the bullet markers and list numbers to misalign.

jlewi:
Should we
  * Put the URL http://<hostname>/_gcp_gatekeeper/authenticate in a code box?
  * Put <hostname> and <name> in bold?

screenshot

jlewi:
Step 4. in Create OAuth Client id; can we put the environment variables in a code box so its easy to copy and paste?

jlewi:
In the Create Kubeflow Deployment can you add back the code boxes. The code boxes make it easy to see the commands you need to run and copy and paste them.

Compare the following in your PR
create_deployment

To the current 
current

richardsliu:
@jlewi Done. Netlify and github renders Markdown differently, so the preview versions didn't match before.

content/docs/started/getting-started-gke.md
```diff
@@ -14,57 +14,55 @@ bref = "The Kubeflow project is dedicated to making deployments of machine learn
 
 Instructions for optimizing and deploying Kubeflow on GKE.
 
-Running on Kubeflow on GKE comes with the following advantages
+Running on Kubeflow on GKE comes with the following advantages:
```
jlewi: delete "on"

jlewi: -> Running Kubeflow on GKE

richardsliu:
Done.

Issue #88, opened 2018-07-13
Publish versions of the website corresponding to Kubeflow versions


jlewi:
How do we publish version of the website that map to versions of Kubeflow?

If you look at Kubernetes website I think they cut a version of the website for every version of K8s and then flip it over when the new version launches.

jlewi:
Fixed by #136

PR #91, opened 2018-07-14
Split kubeflow-core into individual prototypes

ankushagarwal:
Update to kubeflow/kubeflow :  https://github.com/kubeflow/kubeflow/pull/1207

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/91)


jlewi:
We should keep the website in sync with the current release.

Presumably kubeflow/kubeflow#1207 won't go out until 0.3.0 or possibly 0.2.3.

So we shouldn't commit this until then.

Alternatively there is an open issue #88 to version our website.

Ideally we would cut branches for every release and then kubeflow.org would point to the branch corresponding to the current release. This way we could continue to update docs at head.

jlewi:
@ankushagarwal The docs are now versioned so this can be submitted.

ankushagarwal:
@jlewi : Ready for review

ankushagarwal:
Self LGTM-ed because @k8s-ci-robot removed LGTM label after a rebase.

PR #94, opened 2018-07-17
Moved pytorch doc from 'About' to 'Getting Started' - issue #82

### LINKED ISSUE #82: The Pytorch doc belongs in the "Getting Started" section rather than the "About" section
The Pytorch doc is currently in the "About" section of the docs
https://www.kubeflow.org/docs/about/pytorch/

It probably belongs rather in the "Getting Started" section:
https://www.kubeflow.org/docs/started/

amsaha:
Fix #82

### DIFF
--- a/content/docs/started/pytorch.md
+++ b/content/docs/started/pytorch.md
@@ -6,7 +6,7 @@ toc = true
 bref= "This guide will walk you through using PyTorch with Kubeflow"
 aliases = ["/docs/pytorch/"]
 [menu.docs]
-  parent = "about"
+  parent = "started"
   weight = 3
 +++
 


amsaha:
Could either of @ewilderj @aronchick review this?

jlewi:
It looks like the Kubernetes page has a section "Tasks"

sarahmaddox:
Good point @jlewi. I suggested moving the doc into the "getting started" section because that was the only section we currently have, apart from the "about" section. I agree that it'd be better to create a new section called "Guides" - or "Tasks", to align with Kubernetes as you've pointed out. (Personally I prefer "Guides", but alignment is important.)

I'm planning to look at the information architecture of the docs in more detail as soon as time allows, and to think about new guides/tutorials we'll need and where to put them. I'll create a doc with some design ideas for discussion, and circulate the doc in the community.

In the meantime, my 2c is that it's good to move the Pytorch guide into "getting started", where the Minikube and GKE docs currently also reside. Another point in favour of the move is that the Pytorch guide is currently a basic getting-started guide too.

jlewi:
/lgtm
/approve

/hold 

@amsaha I'll leave it up to you to decide whether to leave in Getting started or to create a new guides sections.

amsaha:
/hold cancel

I suggest we go ahead with the change now and wait for @sarahmaddox to come back with her suggestions.

PR #96, opened 2018-07-19
Add redirects to GitHub content.

jlewi:
Fix kubeflow/kubeflow#1156

After this PR you could something like

```
curl -L https://kubeflow.org/version/stable/gke/scripts/deploy.sh | bash
```

We can then update the version that stable points to as part of our release process.

### DIFF
--- a/content/_redirects
+++ b/content/_redirects
@@ -1,0 +1,4 @@
+# Redirects from what the browser requests to what we serve
+/version/stable/gke/scripts/deploy.sh              https://raw.githubusercontent.com/kubeflow/kubeflow/v0.2.2/scripts/gke/deploy.sh
+/version/stable/scripts/deploy.sh      https://raw.githubusercontent.com/kubeflow/kubeflow/v0.2.2/scripts/deploy.sh
+/version/stable/source.tar.gz          https://github.com/kubeflow/kubeflow/archive/v0.2.2.tar.gz


pdmack:
Are we documenting this technique outside this PR?
`curl -L https://kubeflow.org/version/stable/gke/scripts/deploy.sh | bash`

jlewi:
@pdmack I was going to wait until after #1239 to update the instructions but I guess we could also update the instructions to be

```
export KUBEFLOW_VERSION=0.2.2
curl -L https://kubeflow.org/version/stable/gke/scripts/deploy.sh | bash
```

It seemed a little weird to me that you specify two different versions "0.2.2" and "stable".

Although I guess we could also create the link 0.2.2 so it would be

```
export KUBEFLOW_VERSION=0.2.2
curl -L https://kubeflow.org/version/${KUBEFLOW_VERSION}/gke/scripts/deploy.sh | bash
```

I'm a little reluctant though to go down the road of maintaining duplicate URLs to the extent GitHub will already give us URLs corresponding to each version.

jlewi:
It looks like we can maybe move Git tags.
http://blog.iqandreas.com/git/how-to-move-tags/

So maybe we should be going that route and not messing with redirects.

pdmack:
Maybe redirects are better. Force pushing tags might get a little squirrelly.

Issue #97, opened 2018-07-20
Split user guide into a section guides


jlewi:
Right now we have an uber [user_guide](https://www.kubeflow.org/docs/about/user_guide/) with different sections for TFJob, Seldon, Jupyter, etc....

This won't scale.

I think we should create a new section either "Guides" that would be at the same level as the About and Getting Started section. We could then have individual pages for things like

- TFJob
- PyTorch
- Seldon
- JupyterHub

Kubernetes has a section called "Tasks" but @sarahmaddox (Technical Writer who will be contributing to Kubeflow) suggested Guides might be better. Lets go with her recommendation.

@sarahmaddox will be evaluate overall layout of the docs at some point. However, I think splitting up the user guide will significantly improve user's ability to find relevant information. So I'd prefer not to wait on that.

sarahmaddox:
+1 to the idea of a "Guides" section.

+1 to the idea of splitting the current user guide into smaller guides, each of which tackles a specific component. The content of the user guide is well suited to such a split. As well as the sections jlewi@ has mentioned, the user guide contains info on setting up a persistent disk, upgrading Kubeflow deployments, troubleshooting, ksonnet.

As a goal for the future, we can add some end-to-end tutorials that help people deploy Kubeflow and integrate specific components (installation on Kubernetes Engine or local, TensorFlow or Pytorch training, Seldon or TensorFlow Serving, and so on). We'll need to choose a few typical workflows and create the tutorials that match them.

amsaha:
@jlewi  is **priority/1** a typo and was intended to be **priority/p1**?

amsaha:
Does this look okay?

sarahmaddox:
@amsaha Thanks, that's a great start towards addressing this issue. Your PR creates the "Guides" section and moves the PyTorch guide into the section, which is good.

This issue #97 also proposes that we split the "User Guide" doc (https://www.kubeflow.org/docs/about/user_guide/) into separate docs, and put those separate docs into the "Guides" section. That's a bigger task than is addressed by your current PR.

amsaha:
@sarahmaddox Yes, I think the separation of user guides is a larger task and has to be taken up via a separate issue. Else the change might will be too big.

amsaha:
Should we simply move the different sections of the current [user_guide](https://www.kubeflow.org/docs/about/user_guide/)? Maybe that is what was suggested. Sorry for the confusion. Let me do that.

sarahmaddox:
[...]

PR #102, opened 2018-07-24
Adds a "Guides" section to document

amsaha:
Fixes issue #97


jlewi:
This is great; thank you.

/lgtm
/approve

Any reason for the hold? Just remove it if you think its good to go.

amsaha:
/hold cancel
@jlewi not sure why there is a conflict in "content/docs/started/pytorch.md".

amsaha:
@jlewi rebasing it fixed it :-) So just neet /lgtm label.

amsaha:
@jlewi need lgtm on this one.

content/docs/guides/usage_reporting.md
```diff
@@ -0,0 +1,36 @@
++++
```
sarahmaddox: This is great, thanks for doing this! I have just one suggestion: Change the file name to use a hyphen instead of an underscore:
usage-reporting.md

Using a hyphen rather than an underscore in URLs is common practice. I see we use hyphens for all the file names in the getting-started section, and it'd be good to standardise.
https://github.com/kubeflow/website/tree/master/content/docs/started

Now that you're removing the old "user_guide.md", this is a good time to start the standardisation.  :)

jlewi:
This looks great.

One more comment can we sort the Guides Alphabetically in the Guides Navigation section? I think that will be make them easier to find. I don't know if there's a way to do that automatically or if we have to manually adjust the order.

amsaha:
@jlewi  can you take a look at this? This has been pending for sometime now.

jlewi:
Sorry; was busy with next
/lgtm
/approve

/hold
This is a significant improvement so if you want to go ahead and submit it just remove the hold.

Your point about alphabetical not making sense is well taken.

The problem with the current layout is that I think a common mode is people will want to quickly find information about a particular component "e.g. seldon". In that case lack of Alphabetical ordering makes it really hard to find the information you want.

What if we had a subsection components and within that we alphabetized things

e.g.

Guides
   Requirements
   Components
      jupyter notebooks
      ksonnet
      pytorch
      seldon
      tensorflow serving
      tensorflow  training
   Troubleshooting     
   Upgrading Kubeflow
   Usage Reporting
  
So within the components section we alphabetize but we order the other sections according to user flow

amsaha:
@jlewi I agree. Will send another patch later in the day :-)

amsaha:
/hold off

@jlewi I played around with a second level inside "Guides" and it is not trivial to do. I need to play around for sometime to do it. I suggest we handle it via a separate issue (issue #105)

Issue #51, opened 2018-06-19
Doc on TF serving - document Istio and other features like prometheus


lluunn:
migrating [this](https://github.com/kubeflow/kubeflow/blob/master/components/k8s-model-server/istio-integration.md) to website

jlewi:
See also #97 we should have a new guide page for just TFServing and put all the information about TFServing there.

amsaha:
/assign

Will get this done once #97 is closed (currently pr #102 in review).

amsaha:
Waiting on #105.

PR #113, opened 2018-07-31
Move istio related doc from kubeflow/kubeflow to this repo

amsaha:
Issue #51

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/113)


content/docs/guides/components/istio.md
```diff
@@ -0,0 +1,161 @@
...
+### Metrics
+The istio sidecar reports data to [Mixer](https://istio.io/docs/concepts/policy-and-control/mixer.html).
+We can view the istio dashboard by [installing Grafana](https://istio.io/docs/tasks/telemetry/using-istio-dashboard.html#viewing-the-istio-dashboard).
+Execute the command:
+```
+kubectl -n istio-system port-forward $(kubectl -n istio-system get pod -l app=grafana -o jsonpath='{.items[0].metadata.name}') 3000:3000 &
+```
+Visit http://localhost:3000/dashboard/db/istio-dashboard in your web browser.
+Send some requests to the TF serving service, then there should be some data (QPS, success rate, latency) like
+![istio dashboard](istio-dashboard.png)
```
lluunn: Do you want to copy the image here?

amsaha: Oops! should not have missed that. Thanks @lluunn.
Anyway, fixed it in the PR.

content/docs/guides/components/istio.md
```diff
@@ -0,0 +1,162 @@
...
+The istio sidecar reports data to [Mixer](https://istio.io/docs/concepts/policy-and-control/mixer.html).
+We can view the istio dashboard by [installing Grafana](https://istio.io/docs/tasks/telemetry/using-istio-dashboard.html#viewing-the-istio-dashboard).
+Execute the command:
+```
+kubectl -n istio-system port-forward $(kubectl -n istio-system get pod -l app=grafana -o jsonpath='{.items[0].metadata.name}') 3000:3000 &
+```
+Visit http://localhost:3000/dashboard/db/istio-dashboard in your web browser.
+Send some requests to the TF serving service, then there should be some data (QPS, success rate, latency)
+like
+![istio dashboard](../istio-dashboard.png)
```
lluunn: should be "./"? They are both under components

amsaha: Hugo has a problem displaying images if they are kept in the same level. With "./istio-dashboard.png" the generated URL becomes ```components/istio/istio-dashboard.png``` whereas we want ```components/istio-dashboard.png```. 

Check https://deploy-preview-113--competent-brattain-de2d6d.netlify.com/docs/about/kubeflow/ on how it looks now (with this current patch).

amsaha: https://discourse.gohugo.io/t/keep-images-content-together/51/11

jlewi:
Should we just add this information to the existing TFServing page rather than creating a new page?

content/docs/guides/components/istio.md
abhi-g: Based on what is supported in the current version of Hugo, we have following possibilities:

1. Create a static/images/ directory at the top level and put all referenced images there. This is what we did for the theme referenced images. Pros: All images out of content sections, and not cluttering space there. Cons: Hard to associate images with which post they are referenced by without following an explicit naming scheme.

2. Do what Amit has done, and put the images referenced by the document in the same parent directory. Then it could be referenced by the document with a relative URL of ../image.jpg. Pros: Slightly easier to organize images. Cons: Still need some naming convention to identify images with posts, as all docs in that section directory will have to store referenced images. 
  Better than option 1, but still not ideal.

3. Create a subdir for each document and put .md and referenced images in that dir.
   Pros: Content and images together. 
   Cons: a. The Url for the page will become components/itsio/itsio/itsio-dashboard.png.
              b. Creating a subdir for each document this way also creates a new Section Nav entry which would be completely wrong.

So in the end looks like we are stuck with **option 2** until Hugo improves to support content bundles which is currently expected in 0.26 release. Not happening in another year...

amsaha: Thanks @abhi-g. I went through these options and I agree completely with you :-)

amsaha:
@jlewi I thought about it but then considered that relative newcomers to the site would think in terms of keywords; if the keyword they are thinking of is **Istio**, it is easier for them to search if we have Istio as a different page rather than having it inside TFServing (though I agree that at this point the Istio documentation is TFServing specific).

content/docs/guides/components/istio.md
```diff
@@ -0,0 +1,162 @@
...
+  In addition, a default routing rule is created and routes all requests to `v1`.
+  2. When we want to rollout the new model, use the same name X and set a different version, e.g. `v2`.
+  ```
+  ks param set --env=$ENV $MODEL_COMPONENT version v2
+  ks param set --env=$ENV $MODEL_COMPONENT firstVersion false
+  ks apply $ENV -c $MODEL_COMPONENT
+  ```
+  This deploys the new deployment `X-v2`, but the traffic will still go to `v1`
+  3. Create the new routing rule. For example, the following sends 5% traffic to `v2`.
+  ```
```
abhi-g: This pre section is not rendering properly. Similarly the one above lines 138-142. You might want to try playing around by leaving an empty line before/after or both.

amsaha: Thanks @abhi-g. This Hugo creates a lot of markdown incompatibility :-) updated PR.

jlewi:
@amsaha Can you add a link on the TFServing page to the ISTIO page? In particular on the TFServing page we should have instructions for how to collect telemetry data using istio.

amsaha:
@jlewi done.

amsaha:
Can one of you (@jlewi @lluunn) review this?

PR #131, opened 2018-08-06
Add doc on AVX2 and Haswell

lluunn:
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/131)

### DIFF
--- a/content/docs/guides/troubleshooting.md
+++ b/content/docs/guides/troubleshooting.md
@@ -17,6 +17,16 @@ Linux:
 grep -ci avx /proc/cpuinfo
 ```
 
+### AVX2
+Some components requirement AVX2 for better performance, e.g. TF Serving.
+To ensure the nodes support AVX2, we added minCpuPlatform arg in our deployment
+[config](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L105).
+
+This will fail in regions (e.g. us-central1-a) that do not have Intel Haswell.
+In that case, please choose another region, or change the config to other 
+[platform](https://en.wikipedia.org/wiki/List_of_Intel_CPU_microarchitectures)
+newer than Haswell.
+
 ## Minikube
 
 On [Minikube](https://github.com/kubernetes/minikube) the Virtualbox/VMware drivers for Minikube are recommended as there is a known
--- a/content/docs/guides/troubleshooting.md
+++ b/content/docs/guides/troubleshooting.md
@@ -19,10 +19,13 @@ grep -ci avx /proc/cpuinfo
 
 ### AVX2
 Some components requirement AVX2 for better performance, e.g. TF Serving.
-To ensure the nodes support AVX2, we added minCpuPlatform arg in our deployment
+To ensure the nodes support AVX2, we added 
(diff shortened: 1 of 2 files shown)


jlewi:
Can you add a note on the GKE getting started page in the troubleshooting page as well?

https://deploy-preview-131--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-gke/

It would be good to explain how users actually change the zone as well with GKE. I think they have to change the value of the ZONE environment variable.

content/docs/guides/troubleshooting.md
```diff
@@ -17,6 +17,16 @@ Linux:
 grep -ci avx /proc/cpuinfo
 ```
 
+### AVX2
+Some components requirement AVX2 for better performance, e.g. TF Serving.
+To ensure the nodes support AVX2, we added minCpuPlatform arg in our deployment
+[config](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L105).
+
+This will fail in regions (e.g. us-central1-a) that do not have Intel Haswell.
```
jlewi: "On GCP this will fail"

content/docs/guides/troubleshooting.md
```diff
@@ -17,6 +17,16 @@ Linux:
...
+In that case, please choose another region, or change the config to other 
+[platform](https://en.wikipedia.org/wiki/List_of_Intel_CPU_microarchitectures)
+newer than Haswell.
```
jlewi: Are you saying that we won't  run in us-central1-a with Haswell; but if we pick a newer platform then we can run in us-central1a?

content/docs/guides/troubleshooting.md
lluunn: done

content/docs/guides/troubleshooting.md
lluunn: I think so. https://cloud.google.com/compute/docs/regions-zones/#available
us-central1-a has skylake which is newer than haswell

lluunn:
Added some words in getting started with GKE. PTAL, thanks!

content/docs/guides/troubleshooting.md
```diff
@@ -17,6 +17,16 @@ Linux:
...
 ```
 
+### AVX2
+Some components requirement AVX2 for better performance, e.g. TF Serving.
+To ensure the nodes support AVX2, we added minCpuPlatform arg in our deployment
+[config](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L105).
+
+On GCP this will fail in regions (e.g. us-central1-a) that do not have Intel Haswell.
+In that case, please choose another region, or change the config to other 
+[platform](https://en.wikipedia.org/wiki/List_of_Intel_CPU_microarchitectures)
```
jlewi: How will using a newer platform fix this issue? If a region doesn't have haswell but it has a newer platform then wouldn't it already get scheduled there?

I thought mincpuplatform can be satisfied by newer platforms.

lluunn: No it needs to explicitly have that platform.

jlewi: Lets add a link to 
https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform
as well

PR #165, opened 2018-08-16
WIP Add private cluster docs

ankushagarwal:


content/docs/started/getting-started-gke.md
```diff
@@ -352,16 +352,40 @@ means the GKE nodes won't have public IP addresses. This can improve security by
...
+    ```
+
+1. Update the various ksonnet components to use gcr.io images instead of dockerhub images
+
+    ```
+    cd ${KFAPP}/ks_app
+    ${KUBEFLOW_REPO}/scripts/gke/use_gcr_for_all_images.sh
+    ```
+
+1. Remove components which are not useful in private clusters
```
jlewi: Do we need to add a doc or reference to explain how they can generate an SSL certificate? 
They should be able to do this 
1. By creating/uploading their own cert
1. Use the DNS Acme challenge and Cloud DNS

content/docs/started/getting-started-gke.md
```diff
@@ -352,16 +352,40 @@ means the GKE nodes won't have public IP addresses. This can improve security by
...
+
+    ```
+    cd ${KFAPP}/ks_app
+    ks component rm cloud-endpoints
+    ks component rm cert-manager
+    ks component rm iap-ingress
```
jlewi: Doesn't IAP still work with private clusters?If not what's the issue and what's the alternative ingress story?

ankushagarwal: I haven't tried IAP because cert-manager cannot obtain certificate from letsencrypt in a private cluster. I can play around with it and send a follow-up PR with how to make IAP work.

ankushagarwal: As of this doc, the ingress story is to do port-forwarding

content/docs/started/getting-started-gke.md
ankushagarwal: I haven't tried this. Can I do this in a follow-up PR?

Issue #167, opened 2018-08-17
Apply content reuse for oft-used snippets like version numbers


sarahmaddox:
It'd be nice to be able to refer to minimum versions of various components/frameworks throughout the docs, without causing a maintenance nightmare. In other words, we need some sort of content reuse, so that we can define a snippet of information in one place, and reuse it wherever we need it.

Hugo has the concept of "shortcodes", which are similar to Django variables. 
https://gohugo.io/templates/shortcode-templates/
https://gohugo.io/content-management/shortcodes/

To get this pattern going, let's use shortcodes for:
* the minimum required version of Kubernetes
* the minimum required version of ksonnet

If anyone has any other suggestions for snippets of information we should reuse, please comment on this issue.

sarahmaddox:
area/docs

PR #168, opened 2018-08-17
Added shortcodes for minimum ksonnet and Kubernetes versions.

sarahmaddox:
I've added 2 shortcodes (like Django variables) so that we can refer to the minimum versions of Kubernetes and ksonnet throughout the docs, without causing a maintenance nightmare. Refer to issue #167 for more context.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/168)


themes/kf/layouts/shortcodes/kubernetes-min-version.html
```diff
@@ -0,0 +1 @@
+1.8
```
richardsliu: Should this also be a link?

sarahmaddox: I've changed the K8s shortcode to a link, and I've also added a bit about shortcodes in the docs guide (the website README). You can see the README updates staged here: https://github.com/sarahmaddox/website/tree/shortcodes

Issue #175, opened 2018-08-22
[gcp] Document the service accounts used by Kubeflow


jlewi:
We should explain the different service accounts that we create for Kubeflow

Right now we create 3

https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L32

jlewi:
```
{% set KF_ADMIN_NAME = NAME_PREFIX + '-admin' %}
{% set KF_USER_NAME = NAME_PREFIX + '-user' %}
{% set KF_VM_SA_NAME = NAME_PREFIX + '-vm' %}
```

Admin is used for some admin tasks like configuring the load balancers. The idea here is that its needed to deploy Kubeflow but not needed to actually run jobs.

User is intended to be used by training jobs, models to access GCP resources (GCS, BigQuery, etc...).

VM is used only for the VM service account.

We have 3 different service accounts because we want to follow the principle of least privilege.

PR #176, opened 2018-08-22
Document the service accounts used by Kubeflow

ankushagarwal:
Fixes https://github.com/kubeflow/website/issues/175

/assign @jlewi

### DIFF
--- a/content/docs/started/getting-started-gke.md
+++ b/content/docs/started/getting-started-gke.md
@@ -345,6 +345,20 @@ cd ${KFAPP}
 ${KUBEFLOW_REPO}/scripts/kfctl.sh apply k8s
 ```
 
+### GCP Service Accounts
+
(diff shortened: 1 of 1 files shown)


content/docs/started/getting-started-gke.md
```diff
@@ -345,6 +345,20 @@ cd ${KFAPP}
 ${KUBEFLOW_REPO}/scripts/kfctl.sh apply k8s
 ```
 
+### GCP Service Accounts
+
+Creating a deployment using `kfctl.sh` creates three service accounts in the GCP Project. These service accounts are created using the principle of least privilege. The three service accounts are:
+
+* kubeflow-admin
```
jlewi: Won't the service account names depend on the KFAPP(deployment) name?

ankushagarwal: Updated.

PR #198, opened 2018-08-31
Updating instructions for installing KVM2 on Ubuntu & CentOS

abhi-g:
The current instructions on the website are already out-of-date as we noticed recently. Updating to reflect current valid procedure.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/198)

### DIFF
--- a/content/docs/started/getting-started-minikube.md
+++ b/content/docs/started/getting-started-minikube.md
@@ -21,22 +21,52 @@ Install [Virtual Box](https://www.virtualbox.org/wiki/Downloads) or [VMware Fusi
 ##### Ubuntu
 Install [Virtual Box](https://www.virtualbox.org/wiki/Downloads) or [KVM](http://www.linux-kvm.org/).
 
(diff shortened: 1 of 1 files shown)


content/docs/started/getting-started-minikube.md
```diff
@@ -21,22 +21,52 @@ Install [Virtual Box](https://www.virtualbox.org/wiki/Downloads) or [VMware Fusi
...
+
+# Add yourself to the libvirt group so you don't need to sudo
+# NOTE: For older Debian/Ubuntu versions change the group to `libvirtd`
+$ sudo usermod -a -G libvirt $(whoami)
+
+# Update your current session for the group change to take effect
+# NOTE: For older Debian/Ubuntu versions change the group to `libvirtd`
+$ newgrp libvirt
+```
+Then install the driver itself:
```
richardsliu: I remember there was a reboot step. Do we need to mention it here?

abhi-g: I believe the reboot was needed only to ensure that user's new group membership takes effect. That is actually achieved here by the 'newgrp libvirt' command.

PR #247, opened 2018-10-04
Improve the instructions for setting up the OAuth client. 

### LINKED ISSUE #177: getting-started-gke.md discrepancy
When I am looking at the OAuth consent screen, I do not see an option to enter `Authorized domains`.  Is this a deprecated option?

jlewi:
Improve the instructions for setting up the OAuth client. 
* Add screenshots
* Explain that Authorized Domains may not be an option and explain they
  don't need to set it in that case.
Fix #177


content/docs/started/getting-started-gke.md
```diff
@@ -33,15 +33,18 @@ Create an OAuth client ID to be used to identify Cloud IAP when requesting acces
...
-   * Under **Authorized domains**, enter
+   * If you see **Authorized domains**, enter
 
      ```
      <project>.cloud.goog
      ```
-
-     where \<project\> is your Google Cloud Platform (GCP) project ID.
-
+     * where \<project\> is your Google Cloud Platform (GCP) project ID.
```
sarahmaddox: The indentation is off here. I think indentation by 8 characters should work - similar to what's on line 57 onwards of this file.

content/docs/started/getting-started-gke.md
```diff
@@ -33,15 +33,18 @@ Create an OAuth client ID to be used to identify Cloud IAP when requesting acces
...
+     * If you are using your own domain e.g. **acme.com** you should add that as well
+     * Depending on your project **Authorized domains** might not be an option
    * Click Save.
+   * Here's a screenshot
+   
+     ![consent-screen](/docs/images/consent-screen.png)
```
sarahmaddox: The screenshot shows an authorized domain = "kubeflow.org" but our instructions say it must follow the pattern "<project>.cloud.goog". If "kubeflow.org" is an example of "using your own domain", then we should use something other than "kubeflow.org" as the example. It's confusing as it is, because people may justifiably think they need to enter "kubeflow.org".

content/docs/started/getting-started-gke.md
```diff
@@ -33,15 +33,18 @@ Create an OAuth client ID to be used to identify Cloud IAP when requesting acces
...
 
      ```
      <project>.cloud.goog
      ```
-
-     where \<project\> is your Google Cloud Platform (GCP) project ID.
-
+     * where \<project\> is your Google Cloud Platform (GCP) project ID.
+     * If you are using your own domain e.g. **acme.com** you should add that as well
+     * Depending on your project **Authorized domains** might not be an option
```
sarahmaddox: Suggested rewording, to avoid sounding like folks are missing out on something:

The **Authorized domains** option appears only for certain project configurations. If you don't see the option, that's fine.

content/docs/started/getting-started-gke.md
jlewi: Good point.

jlewi:
Thanks. PTAL.

abhi-g:
Looks good to me. I'll let @sarahmaddox have the final word.

sarahmaddox:
Thanks!
/lgtm
/approve

PR #269, opened 2018-10-24
Add instructions to setup kubeflow using ArgoCD

dthomson25:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/269)


content/docs/guides/gitops-for-kubeflow.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Argo CD and GitOps
+Argo CD is a Kubernetes-native Declarative Continuous Delivery tool that follows the GitOps methodology.  Along with all the benefits of using GitOps, Argo CD offers:
```
jlewi: Add a link to ArgoCD?

content/docs/guides/gitops-for-kubeflow.md
```diff
@@ -0,0 +1,147 @@
...
+brew install argoproj/tap/argocd
+```
+
+1. Set extra permissions for Argo CD
+Since the GKE cluster has RBAC enabled, you will need to grant your account the ability to create new cluster roles by running:
+```shell
+kubectl create clusterrolebinding YOURNAME-cluster-admin-binding --clusterrole=cluster-admin --user=YOUREMAIL@gmail.com
+```
+
+1. Expose the Argo CD API server
```
jlewi: Is it possible to connect via kubectl port-forward?
We generally try to avoid telling people to expose services via LoadBalancer because we don't want to assume they understand the security implications of doing so.

jlewi:
/ok-to-test
This is great; just a few comments.

content/docs/guides/gitops-for-kubeflow.md
dthomson25: That's a great point! I just changed the directions to take that into account.

content/docs/guides/gitops-for-kubeflow.md
dthomson25: 👍

dthomson25:
@jlewi Thanks, and I just pushed some changes to address your comments! Can you take another look when you have a chance?

PR #273, opened 2018-10-26
KUBEFLOW_REPO should be full path

connected-bsamadi:
Making it explicit that KUBEFLOW_REPO should be the full path not just a name. I also added `cd ..` so that the users can run the following commands to deploy Kubeflow into GKE:
```
mkdir ${KUBEFLOW_REPO}
    cd ${KUBEFLOW_REPO}
    export KUBEFLOW_TAG={{% kf-stable-tag %}}
    curl https://raw.githubusercontent.com/kubeflow/kubeflow/${KUBEFLOW_TAG}/scripts/download.sh | bash
    cd ..
    ${KUBEFLOW_REPO}/scripts/kfctl.sh init ${KFAPP} --platform gcp --project ${PROJECT}
    cd ${KFAPP}
    ${KUBEFLOW_REPO}/scripts/kfctl.sh generate platform
    ${KUBEFLOW_REPO}/scripts/kfctl.sh apply platform
    ${KUBEFLOW_REPO}/scripts/kfctl.sh generate k8s
    ${KUBEFLOW_REPO}/scripts/kfctl.sh apply k8s
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/273)

### DIFF
--- a/content/docs/started/getting-started-gke.md
+++ b/content/docs/started/getting-started-gke.md
@@ -84,8 +84,9 @@ Run the following steps to deploy Kubeflow:
     cd ${KUBEFLOW_REPO}
     export KUBEFLOW_TAG={{% kf-stable-tag %}}
     curl https://raw.githubusercontent.com/kubeflow/kubeflow/${KUBEFLOW_TAG}/scripts/download.sh | bash
+    cd ..
      ```
-   * **KUBEFLOW_REPO** directory where you want to download the source to
+   * **KUBEFLOW_REPO** full path to the directory where you want to download the source to
    * **KUBEFLOW_TAG** a tag corresponding to the version to checkout such as `master` for latest code.
    * **Note** you can also just clone the repository using git.
 1. Run the following scripts to set up and deploy Kubeflow:


content/docs/started/getting-started-gke.md
```diff
@@ -84,8 +84,9 @@ Run the following steps to deploy Kubeflow:
     cd ${KUBEFLOW_REPO}
     export KUBEFLOW_TAG={{% kf-stable-tag %}}
     curl https://raw.githubusercontent.com/kubeflow/kubeflow/${KUBEFLOW_TAG}/scripts/download.sh | bash
+    cd ..
      ```
-   * **KUBEFLOW_REPO** directory where you want to download the source to
+   * **KUBEFLOW_REPO** full path to the directory where you want to download the source to
```
jlewi: Can we change this environment variable KUBEFLOW_SRC?

Otherwise it interferes with kfctl.sh which uses KUBEFLOW_REPO?

jlewi: I think the problem here is that we are using environment variable KUBEFLOW_REPO but we should be using a different variable like KUBEFLOW_SRC.

The commands here shouldn't depend on KUBEFLOW_REPO being a full path; i.e. mkdir and cd work with the whatever directory you specify.

download.sh will then download into the current directory.

KUEFLOW_REPO is used by kfctl.sh. It will automatically be set based on the location of kctl.sh.
The environment variable KUBEFLOW_REPO should only be set if you want to override the default for some reason.

sarahmaddox: Note that PR [#277](https://github.com/kubeflow/website/pull/277) changes KUBEFLOW_REPO to KUBEFLOW_SRC in this and other files.

abhi-g:
Thanks @connected-bsamadi for fixing the commands on the doc. But as @sarahmaddox mentions this is now resolved with #277

PR #280, opened 2018-10-31
Add doc for rollout model

lluunn:
https://github.com/kubeflow/kubeflow/issues/1309
https://github.com/kubeflow/kubeflow/issues/1844

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/280)

### DIFF
--- a/content/docs/guides/components/istio.md
+++ b/content/docs/guides/components/istio.md
@@ -15,28 +15,32 @@ quota, rollout and A/B testing.
 ## Install Istio
 We assume Kubeflow is deployed in the `kubeflow` namespace.
 
+### 1. Download Istio
 Follow the istio [doc](https://istio.io/docs/setup/kubernetes/quick-start/)
-to install istio.
+to download istio.
 
-1. download or clone the istio source. cd to the directory.
-2. Install Istio's CRD,
+- download or clone the istio source. cd to the directory.
+- Install Istio's CRD,
     ```
     kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml
     ```
 
+### 2. Configure sidecar injection
 Before installing Istio components, we will change some configuration.
 We will use isio's automatic sidecar injection. However, we want to inject only to those services with annotation.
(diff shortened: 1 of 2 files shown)


content/docs/guides/components/istio.md
```diff
@@ -45,14 +49,27 @@ gcloud container clusters describe XXXXXXX --zone=XXXXXX | grep -e clusterIpv4Ci
...
 servicesIpv4Cidr: 10.7.240.0/20
 ```
-5. Set the outbound IP ranges (direcly in the configmap, or use helm). The value should be like
+
+- Set the outbound IP ranges (direcly in the configmap). The value should be like
    `"10.32.0.0/14\,10.35.240.0/20"`.
 
-6. Install istio (without mTLS)
+### 4. Configure istio-ingressgateway
+We are using a regular ingress instead of Istio's ingressgateway as the entry to cluster, so this service should not
```
jlewi: typo "ingressgateway" -> "ingress gateway"

Should we clarify that on GCP we recommend using NodePort and GCLB but on other platforms you might want to use ISTIO ingress?

content/docs/guides/components/istio.md
```diff
@@ -85,6 +99,9 @@ like
 See istio [doc](https://istio.io/docs/tasks/telemetry/metrics-logs.html).
 
 #### Expose Grafana dashboard behind ingress/IAP
+
+TODO(lunkai): update this section
```
jlewi: Can we put TODOs in comments so that won't be rendered? Can we also link TODOs to issues instead of people?

content/docs/guides/components/istio.md
lluunn: done

content/docs/guides/components/istio.md
lluunn: link to issue instead of person now.

I think render the TODO is better so that people know this part is not up-to-date.

lluunn:
cc @ellis-bigelow

Issue #268, opened 2018-10-23
The "What is Kubeflow" information should be in master as well as v0.2


sarahmaddox:
https://v0-2.kubeflow.org/docs/about/kubeflow/

```
## What is Kubeflow?

Kubeflow is *the machine learning toolkit for Kubernetes*. 

It's a collection of configuration files and scripts that you download,
customize, and run. The config files and scripts activate the Kubeflow
[components](/docs/guides/components/components) to do the work. 

The basic workflow is:

* Download the Kubeflow scripts and configuration files.
* Customize the configuration.
* Run the scripts to deploy your containers to your chosen environment.

You adapt the configuration to choose the platforms and services that you want
to use for each stage of the ML workflow: data preparation, model training, 
prediction serving, and service management.

You can choose to deploy your workloads locally or to a cloud environment.
```

PR #285, opened 2018-11-03
Change "What is Kubeflow" information from v0.2

andreyvelich:
Fixes: https://github.com/kubeflow/website/issues/268

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/285)

### DIFF
--- a/content/docs/about/kubeflow.md
+++ b/content/docs/about/kubeflow.md
@@ -29,22 +29,8 @@ Ultimately, we want to have a set of simple manifests that give you an easy to
 use ML stack _anywhere_ Kubernetes is already running, and that can self 
 configure based on the cluster it deploys into.
 
-## What is Kubeflow?
-
 Kubeflow is *the machine learning toolkit for Kubernetes*. 
 
-It's a collection of configuration files and scripts that you download,
-customize, and run. The config files and scripts activate the Kubeflow
-[components](/docs/guides/components/components) to do the work. 
+## What is Kubeflow?
 
-The basic workflow is:
-
-* Download the Kubeflow scripts and configuration files.
-* Customize the configuration.
-* Run the scripts to deploy your containers to your chosen environment.
-
-You adapt the configuration to choose the platforms and services that you want
-to use for each stage of the ML workflow: data preparation, model training, 
-prediction serving, and service management.
-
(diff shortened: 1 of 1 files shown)


content/docs/about/kubeflow.md
```diff
@@ -29,11 +29,25 @@ Ultimately, we want to have a set of simple manifests that give you an easy to
 use ML stack _anywhere_ Kubernetes is already running, and that can self 
 configure based on the cluster it deploys into.
 
+## What is Kubeflow?
+
 Kubeflow is *the machine learning toolkit for Kubernetes*. 
 
-## What is Kubeflow?
+It's a collection of configuration files and scripts that you download,
```
abhi-g: I prefer the existing description which provides a much higher level description of what Kubeflow is. That description allows underlying details of Kubeflow as a platform to evolve without requiring changes to what defines it.

richardsliu: +1. Also Kubeflow has more components than the main repository (which is where the config files and scripts are). For example we also have TF training operators and serving.

andreyvelich: @abhi-g @richardsliu thank you for your review. 
Should we add this information (from v0.2) bellow or just leave "What is Kubeflow" as it is?

abhi-g: @andreyvelich I looked at the git issue from Sarah that you referenced. It seems like a more appropriate change would be to make a PR in the v0.2 branch of the docs to make this section identical to this paragraph in master.

abhi-g: @andreyvelich I looked at the git issue from Sarah that you referenced. It seems like a more appropriate change would be to make a PR in the v0.2 branch of the docs to make this section identical to this paragraph in master.

andreyvelich: @abhi-g Change it in v0.2 branch

andreyvelich:
I signed it!

andreyvelich:
@abhi-g I will create a new PR because here I got problem with CLA. 
It might be because I changed branch from master to v0.2-branch.

abhi-g:
Makes sense. Plus making this change in v0.2 directly might be easier to
avoid a lot of conflicts arising due to other changes getting pulled in etc.

Issue #304, opened 2018-11-24
Upgrade to latest version of Hugo (extended)


sarahmaddox:
Please upgrade the website generator server to the latest version of Hugo *extended*. We need the latest, extended version to build the Docsy theme (see issue #291). The extended version supports SCSS, which the theme requires. (I have "v0.51/extended" running locally, and can successfully build the site using the Docsy theme.)

sarahmaddox:
Note for posterity: I've been having trouble with the Netlify preview for PR #303 due to a known Netlify issue: Netlify doesn't support *extended* Hugo versions. See [netlify/build-image#182](https://github.com/netlify/build-image/issues/182).

The magic fix for the Netlify problem was that I needed to install PostCSS (`npm install -D --save autoprefixer` then `npm install -D --save postcss-cli`), before building the site and checking in the `resources` folder as recommended in the above Netlify issue.

I'm closing this issue, as it'll be fixed by PR #303 if we decide to go ahead with Docsy.

/close

PR #308, opened 2018-11-26
Upgrade hugo version to v0.51

abhi-g:
upgrading hugo to allow for docsy theme to function..

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/308)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
   command = "hugo"
 
 [context.deploy-preview.environment]
-  HUGO_VERSION = "0.51"
+  HUGO_VERSION = "0.51/extended"
 
 [context.production.environment]
-  HUGO_VERSION = "0.51"
+  HUGO_VERSION = "0.51/extended"
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
   command = "hugo"
 
 [context.deploy-preview.environment]
-  HUGO_VERSION = "0.40"
+  HUGO_VERSION = "0.51"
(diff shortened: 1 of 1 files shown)


abhi-g:
fixes #304

abhi-g:
This is all I changed so far and it seems to pass the checks. Looking at your PR, the error from netlify deployment was:
11:42:26 AM: Building sites …
11:42:26 AM: ERROR 2018/11/23 19:42:26 error: failed to transform resource: TOCSS: failed to transform "scss/main.scss" (text/x-scss): this feature is not available in your current Hugo version
11:42:27 AM: Total in 272 ms

So, either my change of version itself is okay, but mixed with your changes to docsy theme, might still fail?

Based on https://gohugo.io/hosting-and-deployment/hosting-on-netlify/ I assumed this is all I needed to do.

@richardsliu Do you know if upgrading the version of hugo in this config is enough?

abhi-g:
Also from your PR's log @sarahmaddox 
"11:42:25 AM: Installing Hugo 0.51" before it failed.
So I am assuming this deployment will fail as well with the docsy theme. We might need "0.51/extended". Not sure yet how to get it with netlify.

sarahmaddox:
I think the error is due to the fact that we need the *extended* version of Hugo, which supports SCSS.

abhi-g:
haha, was worth a try. Doesn't work by just adding /extended in there.

abhi-g:
sounds good.

PR #319, opened 2018-11-30
Update istio doc

lluunn:
- add routing path information
- update grafana section

The link for grafana virtual service manifest is depending on https://github.com/kubeflow/kubeflow/pull/2025

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/319)

### DIFF
--- a/content/docs/guides/components/istio.md
+++ b/content/docs/guides/components/istio.md
@@ -13,65 +13,24 @@ toc = true
 quota, rollout and A/B testing.
 
 ## Install Istio
-We assume Kubeflow is deployed in the `kubeflow` namespace.
-
-### 1. Download Istio
-Follow the istio [doc](https://istio.io/docs/setup/kubernetes/quick-start/)
-to download istio.
-
-- download or clone the istio source. cd to the directory.
-- Install Istio's CRD,
-    ```
-    kubectl apply -f install/kubernetes/helm/istio/templates/crds.yaml
-    ```
-
-### 2. Configure sidecar injection
-Before installing Istio components, we will change some configuration.
-We will use isio's automatic sidecar injection. However, we want to inject only to those services with annotation.
-According to this [table](https://github.com/istio/istio/issues/6476#issuecomment-399219937), we will
-label the namespace as `enabled`, and set policy in `istio-sidecar-injector` (a configmap) as `disabled`.
-
-- label the namespace
-   ```
-   kubectl label namespace kubeflow istio-injection=enabled
-   ```
-- Set the policy. This can be done by editing `install/kubernetes/istio-demo.yaml` directly.
-
-
-### 3. Configure egress traffic
-Also, we need to allow egress traffic, e.g. read models from GCS. This is control by the `-i` flag of the args of 
-`istio-init` container in the same configmap `istio-sidecar-injector`. We can also edit it directly.
-To determine the IP values, follow the instructions in the Istio
-[doc](https://istio.io/docs/tasks/traffic-management/egress/#calling-external-services-directly). For example, for GKE you can do
+We assume Kubeflow is already deployed in the `kubeflow` namespace.
 
 ```
-gcloud container clusters describe XXXXXXX --zone=XXXXXX | grep -e clusterIpv4Cidr -e servicesIpv4Cidr
-clusterIpv4Cidr: 10.4.0.0/14
-servicesIpv4Cidr: 10.7.240.0/20
+kubectl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/master/dependencies/istio/install/crds.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/master/dependencies/istio/install/istio-noauth.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/master/dependencies/istio/kf-istio-resources.yaml
+
+kubectl label namespace kubeflow istio-injection=enabled
 ```
 
-- Set the outbound IP ranges (direcly in the configmap). The value should be like
-   `"10.32.0.0/14\,10.35.240.0/20"`.
+The first command installs Istio's CRDs.
 
-### 4. Configure istio-ingressgateway
-We recommend using a regular ingress instead of Istio's ingress gateway as the entry to cluster on GCP,
-so this service should not be of type `LoadBalancer`.
+The second command installs Istio's core components (without mTLS), with some customization:
+1. sidecar injection configmap policy is changed from `enabled` to `disabled`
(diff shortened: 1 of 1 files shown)


content/docs/guides/components/istio.md
```diff
@@ -82,6 +82,11 @@ ks param set ${MODEL_COMPONENT} injectIstio true
 
 This will inject an istio sidecar in the TF serving deployment.
 
+#### Routing with Istio v.s. Ambassador
```
sarahmaddox: s/v.s./vs

content/docs/guides/components/istio.md
```diff
@@ -60,12 +60,12 @@ so this service should not be of type `LoadBalancer`.
...
-### 4. Install istio (without mTLS)
+### 5. Install istio (without mTLS)
     ```
     kubectl apply -f install/kubernetes/istio-demo.yaml
     ```
 
-### (optional) 5. Deploy the Gateway
+### 6. Deploy the Gateway
 
 This is for rolling out model and doing traffic split. See more detail below.
```
sarahmaddox: There's still a "TODO" here - can we add instructions for deploying the gateway now, since we've added line 109 which says "Requirement: Gateway is deployed, see above".

lluunn: updated, thanks

content/docs/guides/components/istio.md
lluunn: done

jlewi:
@lluunn Lets ask Ellis to take a look.

content/docs/guides/components/istio.md
```diff
@@ -60,16 +60,18 @@ so this service should not be of type `LoadBalancer`.
 - Edit `install/kubernetes/istio-demo.yaml` and change the service type of `istio-ingressgateway` from
   `LoadBalancer` to `NodePort`.
 
-### 4. Install istio (without mTLS)
+### 5. Install istio (without mTLS)
     ```
     kubectl apply -f install/kubernetes/istio-demo.yaml
```
ellis-bigelow: This is a pretty brittle and confusing customer experience. I followed these instructions myself and it's nontrivial to install istio, especially for inexperienced users. It also relies on a brittle and strange filename (demo?) and if istio changes anything, these docs will become incorrect.

Instead, I think it would be much more robust to snapshot istio's install. Knative has taken this approach and bundled their default istio install into a file you can apply straight from the web: https://raw.githubusercontent.com/knative/serving/v0.2.1/third_party/istio-1.0.2/istio.yaml

content/docs/guides/components/istio.md
```diff
@@ -60,16 +60,18 @@ so this service should not be of type `LoadBalancer`.
...
     ```
 
-### (optional) 5. Deploy the Gateway
```
ellis-bigelow: Why is deploying the gateway a separate step? It's all one dependency, right?

content/docs/guides/components/istio.md
```diff
@@ -60,16 +60,18 @@ so this service should not be of type `LoadBalancer`.
...
+### 6. Deploy the Gateway
 
-This is for rolling out model and doing traffic split. See more detail below.
+This is for rolling out model and doing traffic split.
 
-TODO(https://github.com/kubeflow/kubeflow/issues/1309): update this
+```
+kubectl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/master/istio/gateway.yaml
```
ellis-bigelow: +1

content/docs/guides/components/istio.md
```diff
@@ -82,6 +84,11 @@ ks param set ${MODEL_COMPONENT} injectIstio true
 
 This will inject an istio sidecar in the TF serving deployment.
 
+#### Routing with Istio vs Ambassador
```
ellis-bigelow: What's the rationale to allow two types of routing? It's confusing from a users perspective, and multiple ingress paths make creating a secure perimeter more difficult.

content/docs/guides/components/istio.md
```diff
@@ -101,24 +108,12 @@ See istio [doc](https://istio.io/docs/tasks/telemetry/metrics-logs.html).
...
-  getambassador.io/config: |
-    ---
-    apiVersion: ambassador/v0
-    kind:  Mapping
-    name:  grafana_dashboard_mapping
-    prefix: /grafana/
-    service: grafana.istio-system:3000
-  ```
+  - Deploy the VirtualService for Grafana service,
+    [grafana-virtual-service](https://github.com/kubeflow/kubeflow/blob/master/istio/grafana-virtual-service.yaml)
```
ellis-bigelow: Can you make this a copy pastable 
```kubectl apply -f```

File: content/docs/guides/components/istio.md (hunk 3 above)
Comment (ellis-bigelow) [reply content/docs/guides/components/istio.md]: Looking at https://github.com/kubeflow/kubeflow/pull/2020/files, It looks like you've done exactly this. Great!

Comment (lluunn) [reply content/docs/guides/components/istio.md]: Yeah, that's the plan

File: content/docs/guides/components/istio.md (hunk 4 above)
Comment (lluunn) [reply content/docs/guides/components/istio.md]: gateway it's a client resource, not part of the installation.
The installation comes with a dummy gateway in istio's namespace.
Here we are creating one that will be used by kubeflow in kubeflow namespace

File: content/docs/guides/components/istio.md (hunk 6 above)
Comment (lluunn) [reply content/docs/guides/components/istio.md]: We want to move to istio, but we can't get rid of ambassador at one time.
So it's a temporary state

**Comment by lluunn** [comment]
@ellis-bigelow @jlewi updated the doc with new version of istio manifest, PTAL

File: content/docs/guides/components/istio.md [hunk 8]
```diff
@@ -13,63 +13,24 @@ toc = true
...
+The first command installs Istio's CRDs.
 
-This is for rolling out model and doing traffic split. See more detail below.
+The second command installs Istio's core components (without mTLS), with some customization:
+1. sidecar injection configmap policy is changed from `enabled` to `disabled`
+1. istio-ingressgateway is of type `NodePort` instead of `LoadBalancer`
 
-TODO(https://github.com/kubeflow/kubeflow/issues/1309): update this
+The third command deploys some resources for Kubeflow.
+The fourth command label the kubeflow namespace for sidecar injector.
```
ellis-bigelow: Maybe better to say, "enables the sidecar for pods in the kubeflow namespace".

Thinking about this, do you want all pods in that namespace to have the sidecar. Would it be better to make a kubeflow-serving namespace?

lluunn: No it's not injecting all pods in kubeflow namespace.
It only injects when your deployment has the proper annotation.
I think this is the behavior we want.

PR #326, opened 2018-12-04
Update GKE getting started docs.

IMBurbank:
Fixes [kubeflow/kubeflow #1898](https://github.com/kubeflow/kubeflow/issues/1898)

Updated [GKE quickstart](https://www.kubeflow.org/docs/started/getting-started-gke/) and [troubleshooting](https://www.kubeflow.org/docs/guides/gke/troubleshooting-gke/) docs to address the outstanding issues in [kubeflow/kubeflow #1898](https://github.com/kubeflow/kubeflow/issues/1898).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/326)

### DIFF
--- a/content/docs/started/getting-started-gke.md
+++ b/content/docs/started/getting-started-gke.md
@@ -25,6 +25,19 @@ brings the following advantages:
     and troubleshooting
   * GPUs and [TPUs](https://cloud.google.com/tpu/) can be used to accelerate your work
 
+## Prerequisites
+
+  * Ensure [ksonnet](https://ksonnet.io/#get-started) version meets Kubeflow [requirements](/docs/guides/requirements)
+  * Enable [boost mode](https://cloud.google.com/shell/docs/features#boost_mode) to deploy using the [cloud shell](https://cloud.google.com/shell/)
+  * Ensure the following APIs are enabled:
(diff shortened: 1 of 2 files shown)


content/docs/started/getting-started-gke.md
```diff
@@ -56,7 +69,7 @@ Create an OAuth client ID to be used to identify Cloud IAP when requesting acces
         ```
         * `<name>` and `<project>` must have the same values as set in the next
           step when you run the deployment script.
-        * The deployment uses "kubeflow" by default for `<name>` but you can
+        * The deployment uses "ks-kubeflow" by default for `<name>` but you can
```
abhi-g: Is that true? I believe the default is "kubeflow"

IMBurbank: Let me correct this. The correct answer is that it defaults to the `KFAPP` value.

IMBurbank: I've clarified the relationship between `<name>`, `KFAPP`, and `DEPLOYMENT_NAME`. Namely that when `kftcl.sh init $KFAPP ...` is run, `DEPLOYMENT_NAME` is assigned to the `KFAPP` value, saved in `${KFAPP}/env.sh` and subsequently used as the `<name>` value for the envoy-ingress endpoint.

PR #364, opened 2018-12-18
Docs for using ModelDB

garganubhav:
Fixes https://github.com/kubeflow/kubeflow/issues/1861

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/364)

### DIFF
--- a/content/docs/guides/components/modeldb.md
+++ b/content/docs/guides/components/modeldb.md
@@ -13,13 +13,13 @@ toc = true
 
 After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running - 
 
-1. Port forward the modeldb-backend port to port 6543 - 
+ Port forward the modeldb-backend pod to port 6543 - 
 
 ```
 kubectl get pods -n kubeflow # Find your modeldb-backend pod
 kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
 ``` 
-2. Create a ModelDB syncer
+1. Create a ModelDB syncer
 
 ModelDBSyncer is the object that logs models and operations to the ModelDB backend. You can initialize the Syncer with your specified configurations as shown below. Explore the ModelDBSyncer here for more details on the Syncer object and the different ways to initialize it.
 
@@ -39,9 +39,9 @@ syncer_obj = Syncer(
     NewExperimentRun("", "sha_A1B2C3D4"))
 ```
 
-3. Sync Information
+1. Sync Information
 
-Method 1 :
+*Method 1 :
 
 Load all model information from a JSON or a YAML file. The expected key names can be found [here](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/utils/MetadataConstants.py). There are also samples JSON and YAML
  files in samples/basic.
@@ -51,7 +51,7 @@ syncer_obj.sync_all(filepath)
 syncer_obj.sync()
 ```
 
-Method 2 :
+*Method 2 :
 
 Initialize the Dataset, Model, ModelConfig, ModelMetrics classes with the needed information as arguments then call the sync methods on the Syncer object. Finally, call syncer_obj.sync().
 ```
@@ -82,14 +82,14 @@ syncer_obj.sync()
 ```
 The code for the API can be found in [ModelDbSyncerBase.py](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/basic/ModelDbSyncerBase.py), where the Syncer, Dataset, Model, ModelConfig, ModelMetrics classes and their methods are declared.
 
-4.Port-forward the modeldb-frontend pod to port 3000.
+1. Port-forward the modeldb-frontend pod to port 3000.
 ```
 kubectl get pods -n kubeflow # Get modeldb-frontend pod
 kubectl port-forward [modeldb-frontend pod] 3000 -n kubeflow 
 ```
-5. Run your model in the browser at https://localhost:3000/.
+1. Run your model in the browser at https://localhost:3000/.
 
-##Samples
+## Samples
 
 
 [BasicWorkflow.py](https://github.com/mitdbg/modeldb/blob/master/client/python/samples/basic/BasicWorkflow.py) and [BasicSyncAll.py](https://github.com/mitdbg/modeldb/blob/master/client/python/samples/basic/BasicSyncAll.py)
--- a/content/docs/guides/components/modeldb.md
+++ b/content/docs/guides/components/modeldb.md
@@ -13,13 +13,13 @@ toc = true
 
 After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running - 
 
- Port forward the modeldb-backend pod to port 6543 - 
+1)  Port forward the modeldb-backend pod to port 6543 - 
 
 ```
 kubectl get pods -n kubeflow # Find your modeldb-backend pod
 kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
 ``` 
-1. Create a ModelDB syncer
+2)  Create a ModelDB syncer
 
 ModelDBSyncer is the object that logs models and operations to the ModelDB backend. You can initialize the Syncer with your specified configurations as shown below. Explore the ModelDBSyncer here for more details on the Syncer object and the different ways to initialize it.
 
@@ -39,9 +39,9 @@ syncer_obj = Syncer(
     NewExperimentRun("", "sha_A1B2C3D4"))
 ```
 
-1. Sync Information
+3)  Sync Information
 
-*Method 1 :
+* Method 1 :
 
 Load all model information from a JSON or a YAML file. The expected key names can be found [here](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/utils/MetadataConstants.py). There are also samples JSON and YAML
  files in samples/basic.
@@ -51,7 +51,7 @@ syncer_obj.sync_all(filepath)
 syncer_obj.sync()
 ```
 
-*Method 2 :
+* Method 2 :
 
 Initialize the Dataset, Model, ModelConfig, ModelMetrics classes with the needed information as arguments then call the sync methods on the Syncer object. Finally, call syncer_obj.sync().
 ```
@@ -82,12 +82,13 @@ syncer_obj.sync()
 ```
 The code for the API can be found in [ModelDbSyncerBase.py](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/basic/ModelDbSyncerBase.py), where the Syncer, Dataset, Model, ModelConfig, ModelMetrics classes and their methods are declared.
 
-1. Port-forward the modeldb-frontend pod to port 3000.
+4)  Port-forward the modeldb-frontend pod to port 3000.
+
 ```
 kubectl get pods -n kubeflow # Get modeldb-frontend pod
 kubectl port-forward [modeldb-frontend pod] 3000 -n kubeflow 
 ```
-1. Run your model in the browser at https://localhost:3000/.
+5)  Run your model in the browser at https://localhost:3000/.
 
 ## Samples
 
--- a/content/docs/guides/components/modeldb.md
+++ b/content/docs/guides/components/modeldb.md
@@ -1,0 +1,97 @@
++++
+title = "ModelDB"
+description ="ModelDB - A system to manage machine learning models"
+weight = 10
+toc = true
+[menu]
+[menu.docs]
+  parent = "components"
+  weight = 5
++++
+
+## Using ModelDB
+
+After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running - 
+
+1. Port forward the modeldb-backend port to port 6543 - 
+
+```
+kubectl get pods -n kubeflow # Find your modeldb-backend pod
+kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
+``` 
+2. Create a ModelDB syncer
+
+ModelDBSyncer is the object that logs models and operations to the ModelDB backend. You can initialize the Syncer with your specified configurations as shown below. Explore the ModelDBSyncer here for more details on the Syncer object and the different ways to initialize it.
+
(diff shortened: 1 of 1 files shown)


content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+toc = true
+[menu]
+[menu.docs]
+  parent = "components"
+  weight = 5
++++
+
+## Using ModelDB
+
+After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running -
```
jlewi: ModelDB isn't installed by default. People will need to install it by doing
ks generate ..
ks apply ...

Can you add the appropriate instructions please?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Using ModelDB
+
+After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running - 
+
+1)  Port forward the modeldb-backend pod to port 6543 - 
+
+```
+kubectl get pods -n kubeflow # Find your modeldb-backend pod
+kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
```
jlewi: With the latest kubectl you should be able to forward to a service
kubectl port-forward service/argocd-server 8080:443

But there will be an ambassador route
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/modeldb/modeldb.libsonnet#L141
(The route will be /modeldb /katib is a bug). 

So we could point folks at those directions.

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+``` 
+2)  Create a ModelDB syncer
```
jlewi: Can you make this a new section with a descriptive heading e.g. "Loading metrics from Python"

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+## Using ModelDB
+
+After ModelDB is deployed using kfctl.sh and the modeldb-db, modeldb-frontend and modeldb-backend pods are running - 
+
+1)  Port forward the modeldb-backend pod to port 6543 - 
+
+```
+kubectl get pods -n kubeflow # Find your modeldb-backend pod
+kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
+```
```
jlewi: Maybe add some information about the data model in ModelDB; e.g. what are projects, experiments and runs?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+syncer_obj.sync_model("train", model_config1, mdb_model1)
+
+# sync the metrics to the model and also specify which dataset tag to use for it
+syncer_obj.sync_metrics("test", mdb_model1, model_metrics1)
+
+syncer_obj.sync()
+```
+The code for the API can be found in [ModelDbSyncerBase.py](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/basic/ModelDbSyncerBase.py), where the Syncer, Dataset, Model, ModelConfig, ModelMetrics classes and their methods are declared.
+
+4)  Port-forward the modeldb-frontend pod to port 3000.
```
jlewi: Is this needed so that the python code can connect to modeldb? Can we add a section describing this?
I assume if we are running in cluster e.g. in a notebook running on Kubeflow we don't need to port-forward we could just specify the service hostname and port when configuring modeldb?

content/docs/guides/components/modeldb.md
mpvartak: You shouldn't need to port-forward for modeldb-backend; that's only required to access the frontned.

content/docs/guides/components/modeldb.md
mpvartak: This can be:

ModelDB organizes model data in a 3-level model hierarchy. At the bottom of the hierarchy, is an ExperimentRun: every execution of a script/program creates an ExperimentRun. Next, related ExperimentRuns can be grouped into an Experiment (e.g., "running hyperparameter optimization for the Neural Network"). Finally, all Experiments and ExperimentRuns belong to a Project (e.g., "churn prediction").

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+kubectl get pods -n kubeflow # Find your modeldb-backend pod
+kubectl port-forward [modeldb-backend pod] 6543:6543 -n kubeflow
+``` 
+2)  Create a ModelDB syncer
+
+ModelDBSyncer is the object that logs models and operations to the ModelDB backend. You can initialize the Syncer with your specified configurations as shown below. Explore the ModelDBSyncer here for more details on the Syncer object and the different ways to initialize it.
+
+You can initialize the syncer either from a config file ( [see sample config file](https://github.com/mitdbg/modeldb/blob/master/client/syncer.json) ) or explicitly via arguments.
+```
+# Initialize syncer from a JSON or YAML config file
```
mpvartak: It would be easiest to create the syncer directly in code, so let's give that example and then point to the ModelDB website for other ways of instantiating.

# Create a syncer using a convenience API
syncer_obj = Syncer.create_syncer("Project Name", \
                                  "test_user", \
                                  "project description", \
                                  host="modeldb-backend")

content/docs/guides/components/modeldb.md
mpvartak: - To connect to the ModelDB backend, you can just specify the host and port as I commented above (during syncer init)
- The port-forward is to be able to access the modeldb UI

content/docs/guides/components/modeldb.md
mpvartak: FYI @garganubhav this changed in the latest PR. So users now have to run ks commands manually to setup.

content/docs/guides/components/modeldb.md
mpvartak: To use the ModelDB lib on the client machine, they'll first have to do a "pip install modeldb" to get the lib

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,98 @@
...
+syncer_obj = Syncer(
+    NewOrExistingProject("Samples Project", "test_user",
+    "using modeldb light logging"),
+    DefaultExperiment(),
+    NewExperimentRun("", "sha_A1B2C3D4"))
+```
+
+3)  Sync Information
+
+* Method 1 :
```
mpvartak: Can we just keep Method 2 since that's used more frequently? We can say "For other methods of logging please refer to the ModelDB docs on https://github.com/mitdbg/modeldb)

content/docs/guides/components/modeldb.md
mpvartak: I'd recommend we say something like: "Open up a Jupyter notebook you will use for model training. Make sure that the modeldb libraries have been installed already or install via 'pip install modeldb'."

content/docs/guides/components/modeldb.md
garganubhav: are you sure, bcoz when I run my model python script without it gives me this error - thrift.transport.TTransport.TTransportException: Could not connect to any of [('::1', 6543, 0, 0), ('127.0.0.1', 6543)]

mpvartak:
See the syncer setup I mentioned in the PR: you can specify the backend
host there which is just “modeldb-backend” in this case. Then you don’t
need to port forward.

content/docs/guides/components/modeldb.md
garganubhav: ok

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+title = "ModelDB"
+description ="ModelDB - A system to manage machine learning models"
+weight = 10
+toc = true
+[menu]
+[menu.docs]
+  parent = "components"
+  weight = 5
++++
+## Deploying ModelDB
```
jlewi: Can you add an overview section?

This should contain a a couple sentences explaining what ModelDB is and why people might want to use it.

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+
+Use the below commands to deploy ModelDB.
+
+```
+ks generate modeldb modeldb
+ks apply default -c modeldb
+```
+
+## Using ModelDB
```
jlewi: Maybe rename this section to Data Model?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+ 
+ModelDB organizes model data in a 3-level model hierarchy. At the bottom of the hierarchy, is an ExperimentRun: every execution of a script/program creates an ExperimentRun.
+Next, related ExperimentRuns can be grouped into an Experiment (e.g., "running hyperparameter optimization for the Neural Network"). Finally, all Experiments and ExperimentRuns belong to a Project (e.g., "churn prediction").
+
+After ModelDB is deployed and modeldb-db, modeldb-backend and modeldb-frontend pods are running -
```
jlewi: Put this in a new section "Loading Data From Python"

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+## Using ModelDB
+ 
+ModelDB organizes model data in a 3-level model hierarchy. At the bottom of the hierarchy, is an ExperimentRun: every execution of a script/program creates an ExperimentRun.
+Next, related ExperimentRuns can be grouped into an Experiment (e.g., "running hyperparameter optimization for the Neural Network"). Finally, all Experiments and ExperimentRuns belong to a Project (e.g., "churn prediction").
+
+After ModelDB is deployed and modeldb-db, modeldb-backend and modeldb-frontend pods are running - 
+
+1)  Loading Metrics from Python
+
+Open up a Jupyter notebook you will use for model training. Make sure that the ModelDB libraries have been installed already or install via 'pip install modeldb'.
```
jlewi: Can you clarify that this can be a notebook or python file.

Can you make installing modeldb package its own bullet item?

Could also file an issue to add modeldb to our Jupyter runtime images?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+
+ModelDBSyncer is the object that logs models and operations to the ModelDB backend.
+You can initialize the Syncer with your specified configurations as shown below.
+
+* Create a syncer using a convenience API
```
jlewi: Should this be a numbered item in the list?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## Using ModelDB
+ 
+ModelDB organizes model data in a 3-level model hierarchy. At the bottom of the hierarchy, is an ExperimentRun: every execution of a script/program creates an ExperimentRun.
+Next, related ExperimentRuns can be grouped into an Experiment (e.g., "running hyperparameter optimization for the Neural Network"). Finally, all Experiments and ExperimentRuns belong to a Project (e.g., "churn prediction").
+
+After ModelDB is deployed and modeldb-db, modeldb-backend and modeldb-frontend pods are running - 
+
+1)  Loading Metrics from Python
```
jlewi: Can we use markdown to format the numbering rather than doing it manually?

e.g

1. item 1
1. item 2.

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+```
+syncer_obj = Syncer.create_syncer("Project Name", 
+"test_user", 
+"project description", 
+host="modeldb-backend")
+```
+
+For other ways of initializing the syncer see [here](https://github.com/mitdbg/modeldb/blob/master/client/python/light_api.md#b-create-a-modeldb-syncer).
+
+3)  Sync Information
```
jlewi: Can we use automating numbering of the items here?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+
+Initialize the Dataset, Model, ModelConfig, ModelMetrics classes with the needed information as arguments then call the sync methods on the Syncer object. Finally, call syncer_obj.sync().
```
jlewi: Should we define Dataset, Model, ModelConfig, and ModelMetrics in the previous section where we define project, experiment, experiment run.

content/docs/guides/components/modeldb.md
jlewi: Actually maybe just call it concepts.

You can then add other concepts like Dataset, Model, Metrics, etc... see comment below

content/docs/guides/components/modeldb.md
jlewi: I think if the python code isn't running in the K8s cluster then we need to make the backend available via kubect port-forward or some other mechanism.

If the code is running in the cluster (e.g. in Jupyter) then I think we can just configure the hostname and port to match that used by the backend service.

@garganubhav Is this worth capturing in the docs? Can we explain how users configure the hostname and port the modeldb library uses to talk to the backend?

content/docs/guides/components/modeldb.md
```diff
@@ -0,0 +1,94 @@
...
+The code for the API can be found in [ModelDbSyncerBase.py](https://github.com/mitdbg/modeldb/blob/master/client/python/modeldb/basic/ModelDbSyncerBase.py), where the Syncer, Dataset, Model, ModelConfig, ModelMetrics classes and
+their methods are declared.
+
+For other methods of logging please refer to the ModelDB docs [here](https://github.com/mitdbg/modeldb/blob/master/client/python/light_api.md#c-sync-information)
+
+4)  Port-forward the modeldb-frontend pod to port 3000.
+
+```
+kubectl get pods -n kubeflow # Get modeldb-frontend pod
+kubectl port-forward [modeldb-frontend pod] 3000 -n kubeflow
```
jlewi: here and for every port-forward command can we use the name of the service which will be known so we can just give the user a command they can copy and paste so they don't have to fetch the logs?

jlewi:
@garganubhav Will you be able to update this PR next week? It would be great to get this merged.

garganubhav:
@jlewi yes, I will do it today.

PR #390, opened 2019-01-02
Update getting started guide by adding zone.

TimZaman:
See: https://github.com/kubeflow/kubeflow/issues/1914

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/390)

### DIFF
--- a/content/docs/started/getting-started.md
+++ b/content/docs/started/getting-started.md
@@ -66,14 +66,14 @@ Download, set up, and deploy (If you prefer to work from source code, feel free 
 1. Run the following commands to setup and deploy Kubeflow:
     
     ```
-    ${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform none --zone ${ZONE}
+    ${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform none
     cd ${KFAPP}
     ${KUBEFLOW_SRC}/scripts/kfctl.sh generate k8s
     ${KUBEFLOW_SRC}/scripts/kfctl.sh apply k8s
     ```
-   * **${ZONE}** is the deployment zone, make sure this [zone has GPUs](https://cloud.google.com/compute/docs/regions-zones/).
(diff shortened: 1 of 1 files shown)


content/docs/started/getting-started.md
```diff
@@ -66,11 +66,12 @@ Download, set up, and deploy (If you prefer to work from source code, feel free
 1. Run the following commands to setup and deploy Kubeflow:
     
     ```
-    ${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform none
+    ${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform none --zone ${ZONE}
```
sarahmaddox: Is it always essential to add a zone? Looking at issue [kubeflow/kubeflow/1914](https://github.com/kubeflow/kubeflow/issues/1914) I think it may be OK to use the default zone in most cases.

If use of the default zone is good for most cases, then it's best to keep the init command simple. You could mark the zone parameter as optional, and explain the circumstances under which people need to include the parameter.

content/docs/started/getting-started.md
```diff
@@ -66,11 +66,12 @@ Download, set up, and deploy (If you prefer to work from source code, feel free
...
     cd ${KFAPP}
     ${KUBEFLOW_SRC}/scripts/kfctl.sh generate k8s
     ${KUBEFLOW_SRC}/scripts/kfctl.sh apply k8s
     ```
+   * **${ZONE}** is the deployment zone, make sure this [zone has GPUs](https://cloud.google.com/compute/docs/regions-zones/).
```
sarahmaddox: Not everyone needs GPUs, so we should note that they should check the availability GPUs in the zone only if they need GPUs for their workload.

TimZaman: That's true, but the install script does give an error on inavailability.

TimZaman:
Thanks @sarahmaddox. I have changed the PR to reflect your comments.

Issue #428, opened 2019-01-20
current year


Arhell:
Hello
On footer --> Copyright © **2018** old date

PR #429, opened 2019-01-20
update date

Arhell:
issue #428

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/429)

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -17,7 +17,7 @@
         {{ end }}
       </div>
       <div class="col-12 col-sm-4 text-center py-2 order-sm-2">
-        {{ with .Site.Params.copyright }}<small class="text-white">Copyright &copy; {{ 2018 }} {{ .}} <br>{{ "Documentation Distributed under CC BY 4.0" }}</small>{{ end }}
+        {{ with .Site.Params.copyright }}<small class="text-white">Copyright &copy; 2018-{{ now.Year }} {{ .}} <br>{{ "Documentation Distributed under CC BY 4.0" }}</small>{{ end }}
         {{ with .Site.Params.privacy_policy }}<small class="ml-1"><a href="{{ . }}">{{ T "footer_privacy_policy" }}</a></small>{{ end }}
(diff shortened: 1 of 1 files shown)


layouts/partials/footer.html
```diff
@@ -17,7 +17,7 @@
         {{ end }}
       </div>
       <div class="col-12 col-sm-4 text-center py-2 order-sm-2">
-        {{ with .Site.Params.copyright }}<small class="text-white">Copyright &copy; {{ 2018 }} {{ .}} <br>{{ "Documentation Distributed under CC BY 4.0" }}</small>{{ end }}
+        {{ with .Site.Params.copyright }}<small class="text-white">Copyright &copy; {{ 2019 }} {{ .}} <br>{{ "Documentation Distributed under CC BY 4.0" }}</small>{{ end }}
```
sarahmaddox: This is an interesting one. The copyright date should actually *not* be automatically set to the current year for pre-existing content. Because, if we set the copyright date to this year, it means we can no longer claim copyright for last year. If someone published our content last year, they could now claim we didn't claim copyright for it.

That said, there's a problem with having a single copyright statement for the site as a whole, because we're still publishing new content on the site this year, so the copyright date needs to be this year for that content.

I think we should claim copyright from 2018 (fixed date) to current year (auto-updated date). Something like this (code not tested):

```
Copyright &copy; {{ 2018 }} - {{ now.Year}} {{ .}} <br>
```

Arhell: Hello @sarahmaddox 
You are right, updated PR
https://github.com/gohugoio/hugoDocs/blob/master/themes/gohugoioTheme/layouts/partials/site-footer.html#L38

sarahmaddox:
Thanks @Arhell, but in your latest change, I don't see the date range, I just see the current year.

Arhell:
@sarahmaddox 
Thank you, but I want to clarify
Currently there is only one date on the site.
or make as 2018-2019?

Arhell:
Understood thanks
I updated the PR

sarahmaddox:
Nice, thanks!

/ok-to-test
/lgtm
/approve

PR #431, opened 2019-01-22
clarify that KFAPP should not be a path

hougangliu:
KFAPP cannot be a path based on [kfctl.sh](https://github.com/kubeflow/kubeflow/blob/master/scripts/kfctl.sh#L325-L354). `cd ${KFAPP}` may mislead user to name it as a path. We should clarify what it is.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/431)


jlewi:
Thanks
/lgtm
/approve

PR #440, opened 2019-01-24
Added optional instructions on how to change the service type

rummens:
Added additional optional instructions on how to change the service type of the model service, should give some users a hint on how to expose the external IP. Per default this param seems to be set to ClusterIP, which only exposes internal IP.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/440)


rummens:
I signed it!

Issue #459, opened 2019-02-03
remove obsolete deployment info from ksonnet page


amygdala:
On the ksonnet page, https://www.kubeflow.org/docs/components/ksonnet/ , the instructions to deploy via:
`curl https://raw.githubusercontent.com/kubeflow/kubeflow/v${KUBEFLOW_VERSION}/scripts/deploy.sh | bash` 
are obsolete.  Probably we should just remove that section and direct them to the 'getting started' pages instead.

PR #472, opened 2019-02-11
Fixes the ksonnet guide

sarahmaddox:
Removes obsolete info from ksonnet guide and tidies up some text and links.

Fixes https://github.com/kubeflow/website/issues/459

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/472)

### DIFF
--- a/content/docs/components/ksonnet.md
+++ b/content/docs/components/ksonnet.md
@@ -70,7 +70,7 @@ For each environment, you can use ksonnet to deploy the same components
 with slightly different parameters to customize the deployment for a particular 
 environment. For example, this feature makes it easy to run a job locally 
 without GPUs for a small number of steps to make sure the code doesn't crash, 
-then move the deployment to the Cloud to run at scale with GPUs.
+then move the deployment to the cloud to run at scale with GPUs.
 
 ## More about ksonnet
 
--- a/content/docs/components/ksonnet.md
+++ b/content/docs/components/ksonnet.md
@@ -4,7 +4,7 @@ description = "Information about ksonnet as used in Kubeflow"
 weight = 20
 +++
 
-Kubeflow makes use of [ksonnet] to help manage deployments.
+Kubeflow makes use of [ksonnet](https://ksonnet.io/) to help manage deployments.
 
 ## Installing ksonnet
 
@@ -14,8 +14,8 @@ Make sure you have the minimum required version of ksonnet:
 Follow the steps below to install ksonnet:
 
 1. Follow the [ksonnet installation
-   guide][ksonnet-installation], choosing the relevant options for your
-   operating system. For example, if you're on Linux:
+   guide](https://ksonnet.io/get-started/), choosing the relevant options for
+   your operating system. For example, if you're on Linux:
 
     * Set some variables for the ksonnet version:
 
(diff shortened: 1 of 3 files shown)


sarahmaddox:
Preview: https://deploy-preview-472--competent-brattain-de2d6d.netlify.com/docs/components/ksonnet/

content/docs/components/ksonnet.md
```diff
@@ -14,8 +14,8 @@ Make sure you have the minimum required version of ksonnet:
 Follow the steps below to install ksonnet:
 
 1. Follow the [ksonnet installation
-   guide][ksonnet-installation], choosing the relevant options for your
-   operating system. For example, if you're on Linux:
+   guide](https://ksonnet.io/get-started/), choosing the relevant options for
+   your operating system. For example, if you're on Linux:
 
     * Set some variables for the ksonnet version:
```
alecglassford: This may be out of scope for this PR (GitHub literally won't let me comment on the correct line, since it's not part of the diff), but is there a reason you use the `--no-check-certificate` flag on line 31 for the `wget` command that downloads ksonnet? It exposes the user to extra risk of downloading malicious code.

content/docs/components/ksonnet.md
```diff
@@ -44,53 +44,33 @@ Follow the steps below to install ksonnet:
...
-easy to run a job locally without GPUs for a small number of steps to make sure the code doesn't crash, and then easily move that to the
-Cloud to run at scale with GPUs.
+ksonnet treats [environment](https://ksonnet.io/docs/concepts#environment)
+as a first class concept. For example, you can define separate
+environments for your development, test, staging, and production deployments.
+For each environment, you can use ksonnet to deploy the same components 
+with slightly different parameters to customize the deployment for a particular 
+environment. For example, this feature makes it easy to run a job locally 
+without GPUs for a small number of steps to make sure the code doesn't crash, 
+then move the deployment to the Cloud to run at scale with GPUs.
```
alecglassford: I don't think cloud should be capitalized, since it's not a proper name. cloud.google.com and kubernetes.io both use "the cloud" (lowercase). You could also consider "cloud infrastructure" or another term.

layouts/shortcodes/ksonnet-min-version.html
```diff
@@ -1 +1 @@
-<a href="https://ksonnet.io/#get-started">0.13.1</a>
\ No newline at end of file
+<a href="https://ksonnet.io/get-started/">0.13.1</a>
```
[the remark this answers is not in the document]
sarahmaddox: Thanks, that's an interesting comment. In the docs we use this variable under the assumption that it will link to the ksonnet installation instructions, so I'll leave it as it is for this CL. If you think it's worth changing, we'll need to provide the link to both the version info and the installation guide, in every place in the docs where this variable is used.

content/docs/components/ksonnet.md
sarahmaddox: I don't know why we use the `--no-check-certificate` flag. That may be worth an issue of its own. It _is_ out of scope for this CL, especially as we use the flag on other pages too.

content/docs/components/ksonnet.md
sarahmaddox: Done.

sarahmaddox:
Thanks @alecglassford!

/assign @abhi-g

PR #512, opened 2019-03-03
build-pipeline.md nits

ryan-williams:
- bump SDK release version
- clarify some instructions

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/512)

### DIFF
--- a/content/docs/pipelines/build-pipeline.md
+++ b/content/docs/pipelines/build-pipeline.md
@@ -114,7 +114,7 @@ export PATH=MINICONDA_PATH/bin:$PATH
 Run the following to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip3 install https://storage.googleapis.com/ml-pipeline/release/0.1.7/kfp.tar.gz --upgrade
+pip3 install https://storage.googleapis.com/ml-pipeline/release/0.1.11/kfp.tar.gz --upgrade
 ```
 
 After successful installation the command `dsl-compile` should be added to your 
@@ -122,17 +122,18 @@ PATH.
 
 ### Compile the samples
 
-The pipelines are written in Python, but they must be compiled to an 
+The pipelines are written in Python, but must be compiled to an
 intermediate representation before submitting to the Kubeflow Pipelines service:
 
 ```bash
 dsl-compile --py [path/to/py/file] --output [path/to/output/tar.gz]
 ```
 
-For example:
+For example, to compile [the `sequential.py` example from kubeflow/pipelines](https://github.com/kubeflow/pipelines/blob/0.1.11/samples/basic/sequential.py):
 
 ```bash
-dsl-compile --py [ML_REPO_DIRECTORY]/samples/basic/sequential.py --output [ML_REPO_DIRECTORY]/samples/basic/sequential.tar.gz
+DIR=[YOUR PIPELINES REPO DIRECTORY]/samples/basic
+dsl-compile --py $DIR/sequential.py --output $DIR/sequential.tar.gz
 ```
 
 ### Deploy the samples


content/docs/pipelines/build-pipeline.md
```diff
@@ -114,25 +114,25 @@ export PATH=MINICONDA_PATH/bin:$PATH
...
 ```bash
 dsl-compile --py [path/to/py/file] --output [path/to/output/tar.gz]
 ```
 
-For example:
+For example, to compile [the `sequential.py` example from kubeflow/pipelines](https://github.com/kubeflow/pipelines/blob/0.1.11/samples/basic/sequential.py):
 
 ```bash
-dsl-compile --py [ML_REPO_DIRECTORY]/samples/basic/sequential.py --output [ML_REPO_DIRECTORY]/samples/basic/sequential.tar.gz
+dsl-compile --py "$PIPELINES_REPO_DIRECTORY"/samples/basic/sequential.py --output "$PIPELINES_REPO_DIRECTORY"/samples/basic/sequential.tar.gz
```
sarahmaddox: Are you implying that there's an environment variable named PIPELINES_REPO_DIRECTORY? If so, to avoid confusion, we should set the environment variable before using it:

```
export PIPELINES_REPO_DIRECTORY=https://github.com/kubeflow/pipelines/blob/0.1.11
```

Also, I'm not sure that we need the double quotes. Would the following work just as well?

```
dsl-compile --py $PIPELINES_REPO_DIRECTORY/samples/basic/sequential.py --output $PIPELINES_REPO_DIRECTORY/samples/basic/sequential.tar.gz
```

ryan-williams: good q's… I mostly just thought we should `ML` → `PIPELINES` here. My impression is that kubeflow/pipelines used to be referred to as "ML", but that that's been phased out.

Using env vars is probably less clear than the existing square-brackets (which obviously need substituting by the user), so I'm fine to revert that. `dsl-compile` doesn't work with a hard-coded URL like the github.com one you used above, so we'll end up needing to tell the user to substitute something in square- or angle-brackets either way.

*Maybe* we could reduce boilerplate in the actual command like:

```bash
DIR=[PIPELINES REPO DIRECTORY]/samples/basic
dsl-compile --py "$DIR"/sequential.py --output "$DIR"/sequential.tar.gz
```

but the original structure wfm.

re: the quotes, they're generally advisable around env vars in bash (cf. [1](https://mywiki.wooledge.org/Quotes), [2](https://mywiki.wooledge.org/BashPitfalls#echo_.24foo)), but maybe not worth cluttering up the examples over. I'll defer to you / maybe it's moot in this case anyway.

Thanks for the review!

sarahmaddox: Thanks! I like your suggestion of simplifying the code like: 

```
DIR=[YOUR PIPELINES REPO DIRECTORY]/samples/basic
dsl-compile --py "$DIR"/sequential.py --output "$DIR"/sequential.tar.gz
```

(I added the word "YOUR" for even more clarity.)

sarahmaddox: Oh, and probably let's leave out the double quotation marks, as we don't have them elsewhere in the docs.

sarahmaddox: BTW, though I'm not suggesting it here, I've seen recommendations that we should use curly brackets like this:

```
${DIR}/sequential.py
```

That's the convention I'm used to. I hadn't seen the double quotation marks before. Interesting.

ryan-williams: Cool, just pushed the `$DIR` version

Yea "curlies" are sometimes necessary for correctness, and can arguably be better for readability. In general I don't use them unless I have to, for example if we wanted to put as much common-prefix as possible into an env var (all the way through `…/sequential.`, including the trailing dot), we'd need them to delineate where the variable name ends and the extension begins:

```bash
sequential=[YOUR PIPELINES REPO DIRECTORY]/samples/basic/sequential.
dsl-compile --py ${sequential}py --output ${sequential}tar.gz
```

Even that example is probably cleaner as:

```bash
sequential=[YOUR PIPELINES REPO DIRECTORY]/samples/basic/sequential
dsl-compile --py $sequential.py --output $sequential.tar.gz
```

and the lack of curlies will never cause a problem in this case.

otoh, the unquoted vars are liable to misbehave based on quirks of individual users' systems. For example, if someone cloned kf/p into a directory named "kubeflow pipelines", or if its parent directory is called "GitHub Repos", the command will fail cryptically. It's really hard to write safe bash 😞 

But anyway, yea I'm fine to leave out the quotes here, it will do the right thing for >99% of users and is cleaner, and this is technically just bash pseudocode anyway 😀

sarahmaddox: Thanks for the details. :)

PR #521, opened 2019-03-06
Add instructions for running API reference generation tool

richardsliu:
Also fixed an incorrect version name in the docs.

Fixes https://github.com/kubeflow/tf-operator/issues/731.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/521)

### DIFF
--- a/gen-api-reference/README.md
+++ b/gen-api-reference/README.md
@@ -27,7 +27,6 @@ at the root of the repo, and that your GOPATH is set properly. For example:
 ```
 cd $GOPATH/src/github.com/kubeflow/tf-operator/
 ```
-
 1. Run the `gen-tfjob-api.sh` script.
 
 1. Run `git diff` to verify the changes.
--- a/gen-api-reference/README.md
+++ b/gen-api-reference/README.md
@@ -1,0 +1,35 @@
+This folder contains scripts that automatically generate API reference pages for Kubeflow.
+The tool used here can be found at https://github.com/ahmetb/gen-crd-api-reference-docs.
+
+## Prerequisites
+* Clone this repository
+* Clone of the repository for which you are generating reference (e.g. kubeflow/tf-operator).
+
+## Installation
+
+1. Download the tool from https://github.com/ahmetb/gen-crd-api-reference-docs/releases.
+
+1. Extract the tool to a local directory, for example:
+```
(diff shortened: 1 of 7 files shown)


gen-api-reference/README.md
```diff
@@ -0,0 +1,34 @@
+This folder contains scripts that automatically generate API reference pages for Kubeflow.
+The tool used here can be found at https://github.com/ahmetb/gen-crd-api-reference-docs.
+
+## Prerequisites
+* Clone this repository
+* Clone of the repository for which you are generating reference (e.g. kubeflow/tf-operator).
```
sarahmaddox: s/Clone of/Clone

gen-api-reference/README.md
```diff
@@ -0,0 +1,34 @@
+This folder contains scripts that automatically generate API reference pages for Kubeflow.
+The tool used here can be found at https://github.com/ahmetb/gen-crd-api-reference-docs.
```
sarahmaddox: It's not clear whether this directory will contain various API reference generation tools, or just the tool for CRD reference. I think the directory will contain various, but this page currently only describes CRD because that's the only one we have. If that's a correct assumption, then I think you should add a heading between line 1 and line 2, something like this:

```
## Reference docs for Kubernetes custom resources
```

gen-api-reference/README.md
```diff
@@ -0,0 +1,34 @@
...
+1. Extract the tool to a local directory, for example:
+```
+tar -xvf gen-crd-api-reference-docs_linux_amd64.tar.gz -C gen-crd-api-reference-docs
+```
+
+## Usage
+
+1. Open up `gen-tfjob-api.sh` script.
+
+1. Set the `GEN_DOCS` variable to where you have gen-crd-api-reference extracted.
```
sarahmaddox: Should this:
`gen-crd-api-reference`
be this:
`gen-crd-api-reference-docs`
?

gen-api-reference/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. Set the `WEBSITE_ROOT` variable to where your website repository is cloned.
```
sarahmaddox: To clarify, please give an example of the value of this variable.

Issue #505, opened 2019-02-28
Update the docs for the new Kubeflow UI


sarahmaddox:
The main effort as far as docs are concerned will be to re-do the screenshots.
Also check the surrounding wording about UI text and behavior.

See design doc: https://bit.ly/kf-landing-page-cuj

sarahmaddox:
This issue is an umbrella issue. I've created sub-issues for each part of the docs that need updating, as some of the updates require running through a tutorial from start to finish.

sarahmaddox:
I've created separate issues for each doc or section that needs updating. I'll close this umbrella issue now, as I've finished the search and analysis part of the task.

/close

PR #536, opened 2019-03-13
corrected doc text to match what's in pipelines

kmarquardsen:
Reference: https://github.com/kubeflow/pipelines/blob/603597d65b1092d28d77111051e1175c40d77fea/frontend/src/lib/Buttons.ts#L134

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/536)

### DIFF
--- a/content/docs/pipelines/pipelines-quickstart.md
+++ b/content/docs/pipelines/pipelines-quickstart.md
@@ -46,7 +46,7 @@ workload:
     alt="Pipelines UI"
     class="mt-3 mb-3 border border-info rounded">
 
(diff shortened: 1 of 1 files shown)


kmarquardsen:
@sarahmaddox good call. let me clean it up and see if there's others

sarahmaddox:
@kmarquardsen I've just realized: It may be that this fix gets wrapped up in the larger effort when we update the docs for the new UI (https://github.com/kubeflow/website/issues/505).

kmarquardsen:
Sounds good @sarahmaddox let me close this and look through open issues 😄

PR #551, opened 2019-03-23
Sets apart component output for each component.

thedriftofwords:
- Pulls out the component output from each component to make it a tiny bit easier to find with a quick scan.
- Makes a guess at what the component output for the BigQuery query might be (needs verification).
- Removes the **Component:** preceding each component -- this seems redundant, unless other non-component items will be added to this list. Otherwise, the bolded URL / name gives each component sufficient visual distinction on the page.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/551)

### DIFF
--- a/content/docs/pipelines/reusable-components.md
+++ b/content/docs/pipelines/reusable-components.md
@@ -21,57 +21,64 @@ The following components submit jobs to
 [Cloud ML Engine](https://cloud.google.com/ml-engine/docs/)
 on Google Cloud Platform (GCP).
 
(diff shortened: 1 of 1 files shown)


sarahmaddox:
Preview:
https://deploy-preview-551--competent-brattain-de2d6d.netlify.com/docs/pipelines/reusable-components/

sarahmaddox:
Thanks for this update, @thedriftofwords. The update LGTM overall, but please verify the BQ component output as stated in your comment:
"Makes a guess at what the component output for the BigQuery query might be (needs verification)."

PR #555, opened 2019-03-25
Added troubleshooting for crash-looping envoy containers.

### LINKED ISSUE #489: for troubleshooting: document how a backend services quota issue manifests
Creating this so I don't forget, can add more detail later. Could be added to some troubleshooting section.
See this issue for context: https://github.com/kubeflow/kubeflow/issues/2310

Essentially, if the user does a GKE deploy and is seeing the envoy pods crash-looping, and needs to use IAP (the error is ignorable in the non-IAP case), the root cause may be that they have exceeded their backend services quota.  This will almost certainly be the case with the default quota & >1 deployment.
They can check this via 
`kubectl -n kubeflow describe ingress`
and look for something like this in the output
```
Events:
  Type     Reason  Age                  From                     Message
  ----     ------  ----                 ----                     -------
  Warning  Sync    14m (x193 over 19h)  loadbalancer-controller  Error during sync: googleapi: Error 403: Quota 'BACKEND_SERVICES' exceeded. Limit: 5.0 globally., quotaExceeded
```

sarahmaddox:
Fixes #489

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/555)


sarahmaddox:
Preview: https://deploy-preview-555--competent-brattain-de2d6d.netlify.com/docs/gke/troubleshooting-gke/#envoy-pods-crash-looping-root-cause-is-backend-quota-exceeded

sarahmaddox:
This PR is ready for review. Adding a hold to allow for multiple reviewers.

/hold

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
+### Fixing the problem
+
+If you have any redundant Kubeflow deployments, you can delete them using
+the [Deployment Manager](https://cloud.google.com/deployment-manager/docs/).
+
+Alternatively, you can request more backend services quota on the GCP Console.
+
+1. Go to the [quotas page on the 
+  GCP Console](https://console.cloud.google.com/iam-admin/quotas).
+1. Tick the box next to **Compute Engine API**
```
alecglassford: I was confused by this instruction. When I opened the quota page I saw many rows that all said "Compute Engine API," each of which has a tickable box next to it (you can see two in the screenshot that you've embedded). I didn't know which to tick.

Ultimately, I interpreted this instruction as clicking the box under **Service** and searching for "Compute Engine API" to filter to only that service. (By default, it shows "All services"). Is that what you wanted?

I may have misinterpreted, as I realize that there is only one "Backend Services" quota, so filtering by **Service** is not necessary to find it.

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
+1. Click the box under **Metric** and search for **Backend services**:
+    <img src="/docs/images/gcp-backend-quota-1.png" 
+        alt="Connection UI"
+        class="mt-3 mb-3 border border-info rounded">
+
+1. Select **Backend services** and make sure **Backend services** appears
```
alecglassford: Nit: I did get a bit tangled up in this process because, by default, all the metrics are selected. So if you click on "Backend services," it deselects that service and shows every metric *except* the one you want.

It wasn't so difficult for me to realize I had to start over, click "Select none," and *then* search for "Backend services" and select that service. But I just wanted to note that this UI is not a walk in the park.

I suppose my nit is that "Select **Backend services**" made me think it was as simple as just clicking a button. A possible, more open-ended alternative: "Filter the metrics to include only **Backend services**" (or you could describe the series of clicks required, but that seems like too much). However, this is pretty picky and I'm not sure it makes that much of a difference, so totally up to you.

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
+        alt="Connection UI"
+        class="mt-3 mb-3 border border-info rounded">
+
+1. Select **Backend services** and make sure **Backend services** appears 
+  under **Metric**:
+    <img src="/docs/images/gcp-backend-quota-2.png" 
+        alt="Connection UI"
+        class="mt-3 mb-3 border border-info rounded">
+
+1. Click **EDIT QUOTAS**. A quota editing form opens on the right of the
```
alecglassford: It's at this point where it seems necessary to tick the box next to "Compute Engine API (Backend Services" - otherwise, the quota editing form is empty. But it was not really feasible at step 1, since there were so many possible boxes to check. I think moving that instruction right above this one would probably improve the flow.

content/docs/gke/troubleshooting-gke.md
alecglassford: See my comment on line 297: my suggested fix is just to move this list item directly above that one.

(Also, it should have a period at the end.)

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
+the [Deployment Manager](https://cloud.google.com/deployment-manager/docs/).
+
+Alternatively, you can request more backend services quota on the GCP Console.
+
+1. Go to the [quotas page on the 
+  GCP Console](https://console.cloud.google.com/iam-admin/quotas).
+1. Tick the box next to **Compute Engine API**
+1. Click the box under **Metric** and search for **Backend services**:
+    <img src="/docs/images/gcp-backend-quota-1.png" 
+        alt="Connection UI"
```
alecglassford: I think you may have inadvertently copied this alt text from elsewhere; it doesn't seem to apply here.

Suggestion: "Screenshot of filtering for the Backend services metric in the GCP Console's quota page"

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
+        class="mt-3 mb-3 border border-info rounded">
+
+1. Select **Backend services** and make sure **Backend services** appears 
+  under **Metric**:
+    <img src="/docs/images/gcp-backend-quota-2.png" 
+        alt="Connection UI"
```
alecglassford: Again, this alt text seems wrong.

Suggestion: "Screenshot of the Compute Engine API Backend services quota in the GCP Console's quota page"

amygdala:
Overall, LGTM.  Thank you!

I like Alec's suggestion re: just deep linking to the quota settings (https://console.cloud.google.com/iam-admin/quotas?metric=Backend%20services), as that UI is pretty hard to use.

One thing to maybe add (rephrase as appropriate): the envoy pod crash looping should be ignorable if the cluster is not IAP-enabled, but rather the user is just connecting via a port-forward.

content/docs/gke/troubleshooting-gke.md
```diff
@@ -231,6 +231,74 @@ usually indicates the loadbalancer doesn't think any backends are healthy.
...
       * Check services are pointing at the points (look at the endpoints for the various services)
 
+## Envoy pods crash-looping: root cause is backend quota exceeded
+
+If your logs show the 
+[Envoy](https://istio.io/docs/concepts/what-is-istio/#envoy) pods crash-looping, 
+the root cause may be that you have exceeded your quota for some 
+backend services such as loadbalancers. 
+This is particularly likely if you have multiple, differently named deployments 
+in the same GCP project using [Cloud AIP](https://cloud.google.com/iap/).
```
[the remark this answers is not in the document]
sarahmaddox: Thanks!

sarahmaddox:
Thanks so much for your very helpful review comments @alecglassford. I've gone ahead and used the deep link, which dramatically simplified things.

Thanks for your review @amygdala :) I've added a bit about the error being ignorable. PTAL and LGTM if you're happy with the updates.

PR #559, opened 2019-03-26
Update install-sdk.md

joeliedtke:
As I understand it, pip3 is the Python3 version of pip. On systems with both Python 2 and 3, pip will use Python 2. This affects hosted notebook environments that have both Python 2 and 3 installed, such as Kubeflow notebooks or Google Cloud Notebooks. Since the SDK requires Python 3, I'd recommend changing this doc to use pip3.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/559)


sarahmaddox:
/assign
/ok-to-test
/lgtm
/approve

Thanks!

PR #589, opened 2019-04-04
Fix typo(s) in installation of kfp package via conda

chiefkarlin:
Most recent version of miniconda installs pip, not pip3 with Python 3.7.  When using GCP Cloud Shell, pip3 is in /usr/bin and will cause kfp package to end up in the users site packages directory and not in the mlpipelines env.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/589)


sarahmaddox:
Thanks!

/ok-to-test
/lgtm
/approve

PR #618, opened 2019-04-10
Add --zone option to kfctl generate step

ultrons:
It's a critical option to set at generate step.
The default value is us-east1-b.
If it's not specified at the generate step.  generated config files have the default zone hardcoded in multiple places.
Manually changing it in cluster*.yaml and doing the apply doesn't seem to work.
Even if it did, it's better that the users are aware of this option when they are creating the deployment.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/618)

### DIFF
--- a/content/docs/gke/deploy/deploy-cli.md
+++ b/content/docs/gke/deploy/deploy-cli.md
@@ -61,6 +61,7 @@ Follow these steps to deploy Kubeflow:
     ```bash
     # The following command is optional, to make kfctl binary easier to use.
     export PATH=$PATH:<path to kfctl in your kubeflow installation>
+    export ZONE=<your target zone> #where the deployment will be created
 
     export PROJECT=<your GCP project>
     export KFAPP=<your choice of application directory name>
@@ -70,7 +71,7 @@ Follow these steps to deploy Kubeflow:
     kfctl init ${KFAPP} --platform gcp --project ${PROJECT} --use_basic_auth -V
 
     cd ${KFAPP}
-    kfctl generate all -V
+    kfctl generate all -V --zone ${ZONE}
     kfctl apply all -V
     ```
    * **${KFAPP}** - the _name_ of a directory where you want Kubeflow 


content/docs/gke/deploy/deploy-cli.md
```diff
@@ -61,6 +61,7 @@ Follow these steps to deploy Kubeflow:
     ```bash
     # The following command is optional, to make kfctl binary easier to use.
     export PATH=$PATH:<path to kfctl in your kubeflow installation>
+    export ZONE=<your target zone> #where the deployment will be created
```
sarahmaddox: We should give a bit more of a clue that "zone" is a GCP concept:

```
export ZONE=<your target GCP zone>
```

content/docs/gke/deploy/deploy-cli.md
```diff
@@ -70,7 +71,7 @@ Follow these steps to deploy Kubeflow:
     kfctl init ${KFAPP} --platform gcp --project ${PROJECT} --use_basic_auth -V
 
     cd ${KFAPP}
-    kfctl generate all -V
+    kfctl generate all -V --zone ${ZONE}
```
sarahmaddox: @gabrielwen Can people specify the zone in the `kfctl init` command or only in the `generate` command?

gabrielwen: no - location is not needed during `init`

sarahmaddox: Great, thanks Gabriel.

content/docs/gke/deploy/deploy-cli.md
ultrons: I agree Sarah. 
Since the page in question is for GKE:
https://www.kubeflow.org/docs/gke/deploy/deploy-cli/

Probably it gives some context. But erring on being more explicit is always better.
Do I need to create another pull request or can it be accepted with your change?

sarahmaddox: Thanks @ultrons! You can use this PR. If you make the change in the same branch, GitHub automatically adds the change to this PR. Since it's a small change, you can do it in the GitHub UI if you prefer. If that doesn't work, let me know and I'll approve this PR then you can make another.

sarahmaddox: Friendly ping @ultrons

PR #686, opened 2019-05-08
Adding nuclio component doc

yaronha:
Adding nuclio component documentation and examples

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/686)

### DIFF
--- a/content/docs/components/nuclio.md
+++ b/content/docs/components/nuclio.md
@@ -1,0 +1,173 @@
++++
+title = "Nuclio functions"
+description = "Nuclio - High performance serverless for data processing and ML"
+weight = 40
++++
+
+## Nuclio Overview
+
+[nuclio](https://github.com/nuclio/nuclio) is a high performance serverless platform which runs over docker or kubernetes 
+and automate the development, operation, and scaling of code (written in 8 supported languages).
+Nuclio is focused on data analytics and ML workloads, it provides extreme performance and parallelism, supports stateful and data intensive 
+workloads, GPU resource optimization, check-pointing, and 14 native triggers/streaming protocols out of the box including HTTP, Cron, batch, Kafka, Kinesis, 
+Google pub/sub, Azure event-hub, MQTT, etc. additional triggers can be added dynamically (e.g. [Twitter feed](https://github.com/v3io/tutorials/blob/master/demos/stocks/read-tweets.ipynb)).
+  
+nuclio can run in the cloud as a [managed offering](https://www.iguazio.com/), or on any Kubernetes cluster (cloud, on-prem, or edge)<br>
+[read more about nuclio ...](https://github.com/nuclio/nuclio)
+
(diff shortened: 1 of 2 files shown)


content/docs/components/nuclio.md
```diff
@@ -0,0 +1,173 @@
...
+Nuclio is focused on data analytics and ML workloads, it provides extreme performance and parallelism, supports stateful and data intensive 
+workloads, GPU resource optimization, check-pointing, and 14 native triggers/streaming protocols out of the box including HTTP, Cron, batch, Kafka, Kinesis, 
+Google pub/sub, Azure event-hub, MQTT, etc. additional triggers can be added dynamically (e.g. [Twitter feed](https://github.com/v3io/tutorials/blob/master/demos/stocks/read-tweets.ipynb)).
+  
+nuclio can run in the cloud as a [managed offering](https://www.iguazio.com/), or on any Kubernetes cluster (cloud, on-prem, or edge)<br>
+[read more about nuclio ...](https://github.com/nuclio/nuclio)
+
+## Using Nuclio In Data Science Pipelines 
+
+Nuclio functions can be used in the following ML pipline tasks:
```
jlewi: The bulleted list isn't rendering correctly. You can see the preview here
https://deploy-preview-686--competent-brattain-de2d6d.netlify.com/docs/components/nuclio/

jlewi:
I wonder if we shouldn't create a separate section named something like "Integrating with other systems" or use cases?

We're starting to have more docs aimed at explaining how Kubeflow can connect with other systems and it might be worth distinguishing between docs describing a particular application and docs explaining how applications that play together.

This doc is really about how to connect Kubeflow Jupyter and (maybe later pipelines) with nuclio.

So this document isn't really describing an independent component or application. The applications would be

* jupyter
* pipelines
* nuclio

The main docs for the latter live elsewhere.

/cc @sarahmaddox

content/docs/components/nuclio.md
yaronha: @jlewi fixed

yaronha:
> I wonder if we shouldn't create a separate section named something like "Integrating with other systems" or use cases?
> 
> We're starting to have more docs aimed at explaining how Kubeflow can connect with other systems and it might be worth distinguishing between docs describing a particular application and docs explaining how applications that play together.
> 
> This doc is really about how to connect Kubeflow Jupyter and (maybe later pipelines) with nuclio.
> 
> So this document isn't really describing an independent component or application. The applications would be
> 
> * jupyter
> * pipelines
> * nuclio
> 
> The main docs for the latter live elsewhere.
> 
> /cc @sarahmaddox

right, i do assume that each app has its own web site w details about it and its usage, so this section is sort of an overview + how it plugs into the overall story.

BTW i added example of how it integrate w the nuclio new `pipelines` components + a link to nuclio pipelines `md` doc.

PR #698, opened 2019-05-13
Highligt how to install k8s cluster.

gyliu513:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/698)


sarahmaddox:
@jlewi Do you have any thoughts on recommending kind over Minikube, as this PR does? Also, note that the page that this PR updates assumes that the user already has a Kubernetes cluster set up. So, I'm thinking this change is out of place.

* We talk about using Minikube if you want to run Kubernetes locally in a virtual machine (VM):
  https://www.kubeflow.org/docs/started/getting-started/
* This guide describes the Minikube setup in detail:
  https://www.kubeflow.org/docs/started/getting-started-minikube/

Issue #669, opened 2019-04-30
Add docs for preemptible VMs/GPUs


sarahmaddox:
The draft content is available in a Google Doc written by the engineering team.

PR #699, opened 2019-05-14
Added Pipelines preemptible VMs and GPUs

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/669

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/699)

### DIFF
--- a/content/docs/pipelines/preemptible.md
+++ b/content/docs/pipelines/preemptible.md
@@ -1,6 +1,6 @@
 +++
-title = "Using Preemptible VMs and GPUs"
-description = "Configuring preemptible VMs and GPUs for Kubeflow Pipelines"
+title = "Using Preemptible VMs and GPUs on GCP"
+description = "Configuring preemptible VMs and GPUs for Kubeflow Pipelines on GCP"
 weight = 65
 +++
 
@@ -99,8 +99,8 @@ The above function works for both methods of generating the `ContainerOp`:
 
 **Note**: 
 
-+   Include `.set_retry(#NUM_RETRY)`in the `ContainerOp` to retry the
-    task after Compute Engine has preempted the VM.
++   Call `.set_retry(#NUM_RETRY)` on your `ContainerOp` to retry 
+    the task after the task is preempted.
 +   If you modified the
     [node taint](https://cloud.google.com/kubernetes-engine/docs/how-to/node-taints)
     when creating the node pool, pass the same node toleration to the
@@ -218,11 +218,11 @@ The above function works for both methods of generating the `ContainerOp`:
 
 **Note**: 
 
-+   Include `.set_gpu_limit(#NUM_GPUs, GPU_VENDOR)` in the `ContainerOp`
-    to specify the GPU limit (for example, `1`) and vendor (for example,
-    `'nvidia'`).
-+   Include `.set_retry(#NUM_RETRY)`in the `ContainerOp` to retry the task
-    after Compute Engine has preempted the VM.
++   Call `.set_gpu_limit(#NUM_GPUs, GPU_VENDOR)` on your 
+    `ContainerOp` to specify the GPU limit (for example, `1`) and vendor (for 
+    example, `'nvidia'`).
++   Call `.set_retry(#NUM_RETRY)` on your `ContainerOp` to retry 
+    the task after the task is preempted.
 +   If you modified the
     [node taint](https://cloud.google.com/kubernetes-engine/docs/how-to/node-taints)
     when creating the node pool, pass the same node toleration to the
--- a/content/docs/pipelines/preemptible.md
+++ b/content/docs/pipelines/preemptible.md
@@ -1,0 +1,314 @@
++++
+title = "Using Preemptible VMs and GPUs"
+description = "Configuring preemptible VMs and GPUs for Kubeflow Pipelines"
+weight = 65
++++
+
+This document describes how to configure preemptible virtual machines
+([preemptible VMs](https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms))
+and GPUs on preemptible VM instances
+([preemptible GPUs](https://cloud.google.com/compute/docs/instances/preemptible#preemptible_with_gpu))
+for your workflows running on Kubeflow Pipelines on Google Cloud Platform (GCP). 
+
+## Introduction
+
+Preemptible VMs are [Compute Engine VM 
+instances](https://cloud.google.com/compute/docs/instances/) that last a maximum 
+of 24 hours and provide no availability guarantees. The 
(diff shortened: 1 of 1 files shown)


sarahmaddox:
Preview: https://deploy-preview-699--competent-brattain-de2d6d.netlify.com/docs/pipelines/preemptible/

sarahmaddox:
This PR is ready for review. Adding a hold to allow for multiple reviewers.

/hold

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
+The above function works for both methods of generating the `ContainerOp`:
+
++   The `ContainerOp` generated from 
+[`kfp.components.func_to_container_op`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_python_op.py).
++   The `ContainerOp` generated from the task factory function, which is
+    loaded by [`components.load_component_from_url`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_components.py).
+
+**Note**: 
+
++   Include `.set_retry(#NUM_RETRY)`in the `ContainerOp` to retry the
```
aikidave: Space  between ` and in.

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
+
+In the [DSL code](https://www.kubeflow.org/docs/pipelines/sdk/dsl-overview/) for
+your pipeline, add the following to the `ContainerOp` instance:
+
+    .apply(gcp.use_preemptible_nodepool())
+
+The above function works for both methods of generating the `ContainerOp`:
+
++   The `ContainerOp` generated from 
+[`kfp.components.func_to_container_op`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_python_op.py).
```
gaoning777: Github can generate a link pointing to the specific line, in this example:
https://github.com/kubeflow/pipelines/blob/c3235d725eb1d1eb06b5600a8291967aa6cf518f/sdk/python/kfp/components/_python_op.py#L283

Here is how to get the link above.
When you click the line, there will be a 'three dot' on the left; Then click it, and choose 'Copy Permalink'. 
The other benefit for the generated link is its version fixed.

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
++   The `ContainerOp` generated from the task factory function, which is
+    loaded by [`components.load_component_from_url`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_components.py).
```
gaoning777: The same as above: https://github.com/kubeflow/pipelines/blob/c3235d725eb1d1eb06b5600a8291967aa6cf518f/sdk/python/kfp/components/_components.py#L66

content/docs/pipelines/preemptible.md
gaoning777: This is part of my mistake. What I tried to say is:
call .set_retry(#NUM_RETRY) of your ContainerOp instances such that the pipeline system will retry the task when the task gets preempted.

Would you mind rewording this sentence? I feel like it might be misinterpreted.

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
+
+In the [DSL code](https://www.kubeflow.org/docs/pipelines/sdk/dsl-overview/) for
+your pipeline, add the following to the `ContainerOp` instance:
+
+    .apply(gcp.use_preemptible_nodepool()
+
+The above function works for both methods of generating the `ContainerOp`:
+
++   The `ContainerOp` generated from 
+[`kfp.components.func_to_container_op`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_python_op.py).
```
gaoning777: The same as above: link

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
++   The `ContainerOp` generated from the task factory function, which is
+    loaded by [`components.load_component_from_url`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/_components.py).
```
gaoning777: The same as above: link

content/docs/pipelines/preemptible.md
```diff
@@ -0,0 +1,314 @@
...
+
+**Note**: 
+
++   Include `.set_gpu_limit(#NUM_GPUs, GPU_VENDOR)` in the `ContainerOp`
```
gaoning777: Thanks for adding this. This is great.

gaoning777:
This looks great. Thanks for pushing the preemptible VM/GPU documentation so fast.
One concern: Do you think we should somehow highlight that the preemptible VMs/GPUs are GCP only features?

content/docs/pipelines/preemptible.md
sarahmaddox: Wow, well spotted.

content/docs/pipelines/preemptible.md
sarahmaddox: If someone adds or removes code to the file, the line number will change. Also, I'm not sure we want to fix on a specific version in the docs, because we'll need to update the links when we move to the next version.

I'm thinking I'll leave the link relatively vague - what do you think?

content/docs/pipelines/preemptible.md
sarahmaddox: Thanks! Done.

content/docs/pipelines/preemptible.md
sarahmaddox: Thanks!

sarahmaddox:
> This looks great. Thanks for pushing the preemptible VM/GPU documentation so fast.
> One concern: Do you think we should somehow highlight that the preemptible VMs/GPUs are GCP only features?

Good point. AFAICT, this is the first doc in the Pipelines section (apart from the end-to-end GCP tutorial) that's GCP-specific. At a later date, we'll probably need to create a separate section for GCP-specific guides. For now, I've added "GCP" to the title and subtitle. (It was already in the intro, but that's already many words into the doc.)

sarahmaddox:
/hold cancel

PTAL @gaoning777

content/docs/pipelines/preemptible.md
gaoning777: If someone adds/removes code to the file, the line will not change in this version, though.
However, you have a good point. It does require some maintenance efforts in this case.
I'm OK with it being vague.
Thanks

sarahmaddox: Thanks!

PR #742, opened 2019-05-30
Add document for KFP API

IronPan:
The html is generated using 
https://github.com/bootprint/bootprint-openapi#javascript-usage

The Swagger definition is located in 
https://github.com/kubeflow/pipelines/blob/master/backend/api/swagger/kfp_api_single_file.swagger.json

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/742)

PR #772, opened 2019-06-05
Improves link and formatting of Pipelines API reference docs

sarahmaddox:
There's an empty bullet point on the Pipelines reference index page:
https://www.kubeflow.org/docs/pipelines/reference/

And the navigation of the website is broken by the Pipelines API reference (generated HTML):
https://www.kubeflow.org/docs/pipelines/reference/api/kubeflow-pipeline-api-spec/

This PR goes some way to improving the situation, given that the HTML file is auto-generated from Swagger. The style of the website nav is affected by the styling injected by Swagger, but at least we have a navigation now. And the index page no longer has an empty bullet point.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/772)

### DIFF
--- a/content/docs/pipelines/reference/api.md
+++ b/content/docs/pipelines/reference/api.md
@@ -1,7 +1,0 @@
-+++
-title = " Pipelines API Reference"
-description = "Reference documentation for the Kubeflow Pipelines API."
-weight = 10
-+++
-
-See the [generated reference for Kubeflow Pipelines API v0.1.20](/docs/pipelines/reference/api/kubeflow-pipeline-api-spec.html).
--- a/content/docs/pipelines/reference/api/kubeflow-pipeline-api-spec.html
+++ b/content/docs/pipelines/reference/api/kubeflow-pipeline-api-spec.html
@@ -1,3 +1,8 @@
++++
+title = "Pipelines API Reference"
+description = "Reference documentation for the Kubeflow Pipelines API"
+weight = 10
++++
 <!DOCTYPE html>
 <html>
 <head>


sarahmaddox:
@IronPan This PR builds on the work done in https://github.com/kubeflow/website/pull/742

This PR adds front matter at the top of the generated HTML file. The front matter looks like this:

```
+++
title = "Pipelines API Reference"
description = "Reference documentation for the Kubeflow Pipelines API"
weight = 10
+++
```

Is it problematic to include the front matter in the script that auto-generates the ref doc from Swagger? The front matter makes Hugo recognise this page as part of the website, and thus includes it in the nav. If we leave out the front matter, we have the empty bullet point on the index page, as currently visible here:
https://www.kubeflow.org/docs/pipelines/reference/

By including the front matter, we remove the empty bullet point and also add the wrap-around nav of the website (albeit with weird styling).

sarahmaddox:
Preview:

* https://deploy-preview-772--competent-brattain-de2d6d.netlify.com/docs/pipelines/reference/
* https://deploy-preview-772--competent-brattain-de2d6d.netlify.com/docs/pipelines/reference/api/kubeflow-pipeline-api-spec/

IronPan:
This looks cool. 

I am wondering if it can point the html from another repository? We are thinking of strong the html in pipeline repo so we can update it automatically as part of our release process. and be able to point to there would make sure the doc is always up-to-date.

sarahmaddox:
We can link to docs anywhere, but not include them within the site navigation. The SDK docs, for example, are hosted on Read The Docs. If you're planning to host the API docs somewhere other than kubeflow.org, it'd be good to have all such docs in one place. Suggestion: Talk to @Ark-kun about putting the docs on Read The Docs.

In the meantime, can we approve this PR to improve the situation until the final decision is made?

IronPan:
sounds good. 
/lgtm
/approve

Issue #680, opened 2019-05-03
Styling of reference docs - PyTorchJob api reference: text colors, size


kbhawkey:
- Field names, descriptions are hard to read (choose a darker gray?)
- The table has gray on gray (gray text and gray row color)
- In general, pale blue color is hard to read
- PyTorchReplicaType (string alias), string is light gray in the method signature -- difficult to read
- Why is PyTorchReplicaType listed at the bottom of the page? Link up?
- Could you add a link to the api reference generation tool, `gen-crd-api-reference-docs`
- Could vary the size of the text in the table for the descriptions

sarahmaddox:
Issue https://github.com/kubeflow/website/issues/730 addresses the styling of code within a table.

For the overall table styling, I recommend we use Bootstrap table styling. This will get rid of the alternating grey and white rows, add a border, and add a background to the header row. Example:

```
<div class="table-responsive">
  <table class="table table-bordered">
    <thead class="thead-light">
      <tr>
        <th>My col header 1</th>
        <th>My col header 1</th>
      </tr>
    </thead>
    <tbody>
      <tr>
        <td>My table cell 1</td>
        <td>My table cell 1</td>
      </tr>
    </tbody>
  </table>
</div>
```

kbhawkey:
👍

kbhawkey:
Closing. Re-open as needed.

sarahmaddox:
The tables in the ref docs could still do with some style improvements:
https://www.kubeflow.org/docs/reference/tfjob/v1/common/

In particular, it'd help to add borders, highlight the header row, and remove the alternating gre/white rows. See earlier comment for styling suggestion.

sarahmaddox:
Thank you @kbhawkey for the fix. The generated ref docs look sooo much better! See [improved styling](https://www.kubeflow.org/docs/reference/tfjob/v1beta2/common/) versus [original styling](https://v0-5.kubeflow.org/docs/reference/tfjob/v1beta2/common/).

PR #776, opened 2019-06-06
[WIP] testing table border and background color

kbhawkey:
Testing border and background colors for tables.
Testing colors for code elements.
See issue #680.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/776)

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -36,8 +36,6 @@ $navbar-dark-active-color: $navbar-dark-color;
 .td-content {
   > table {
     border: 1px solid #cccccc;
-    padding: 0;
-    width: 100%;
 
     td {
       background-color: $table-bg;
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -12,6 +12,11 @@ $dark: #213d7a;
 $info: #adb5bd;
 $light: #dee2e6;
 
+// Table colors
+
+$table-bg: $secondary;
+$table-head-bg: $light;
+
 // Nav bar colors
 
 $white: #fff;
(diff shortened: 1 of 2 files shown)


kbhawkey:
Hello. Will the netlify preview be updated after the `needs-ok-to-test` label is removed?

sarahmaddox:
Hallo @kbhawkey I've added the `ok-to-test` label, but the Netlify preview should have been updated anyway. Are you making CSS changes? (The procedure is a little tricky in that case, as Netlify doesn't fully support extended versions of Hugo.)

kbhawkey:
> 
> 
> Hallo @kbhawkey I've added the `ok-to-test` label, but the Netlify preview should have been updated anyway. Are you making CSS changes? (The procedure is a little tricky in that case, as Netlify doesn't fully support extended versions of Hugo.)
> 
> These guidelines should help: https://testydocsy.netlify.com/docs/getting-started/#deployment-with-netlify

hello @sarahmaddox. OK. I have a local clone of this repo, but did not clone using `git clone --recurse-submodules --depth 1 https://github.com/my/example.git`. Also, I did not specifically clone the docsy repo. I do have a sass processor, an installation of hugo extended,as well as postcss-cli. The am able to view a local preview with `hugo server`.
If I made updates to `_variables_project.scss`, and generated a file in resources, `main.scss_*`, what needs to happen in order to build/run these changes on this netlify site? I have probably misunderstood the use of the theme. Thanks.

sarahmaddox:
I think it should be fine not to recursively clone the theme, provided your diff shows the only update is to the relevant SCSS file and resulting resources.

The problem is that Netlify doesn't yet support the extended Hugo theme. Try this as a workaround:

* Force add your changes in the resources folder: `git add -f resources/`
* Commit the resulting changes and push them back up to your branch.

You should then see the styling changes in the Netlify preview.

kbhawkey:
> 
> 
> I think it should be fine not to recursively clone the theme, provided your diff shows the only update is to the relevant SCSS file and resulting resources.
> 
> The problem is that Netlify doesn't yet support the extended Hugo theme. Try this as a workaround:
> 
>     * Force add your changes in the resources folder: `git add -f resources/`

Something's still amiss. Adding the directory only commits one resource file, not the entire directory. Only one resource file is generated.

sarahmaddox:
> Something's still amiss. Adding the directory only commits one resource file, not the entire directory. Only one resource file is generated.

That could still be OK. Often CSS changes only result in one resource file being regenerated. Do your changes show up in the preview?

sarahmaddox:
Another thing I've found sometimes helps: When building your site locally, try stopping your Hugo server, then running just `hugo -D` to build the site without running the server, then restarting the server with `hugo server -D`.

Issue #716, opened 2019-05-17
.gitattributes path "resource/" is not correctly parsed


michelle192837:
Hi! Getting a error in Prow: "error parsing pattern: directory patterns are not matched recursively, use path/** instead: <resources/>"

Parsing code: https://github.com/kubernetes/test-infra/blob/1b4923af3b1b6de999f7f5d8e946ddd398233f40/prow/gitattributes/gitattributes.go#L90

Think someone should just need to use 'resources/**' instead of 'resources/' in the .gitattributes file.

sarahmaddox:
Hallo @michelle192837 Just checking, as I don't know much about this area: Is Prow complaining about a parsing error that originates in the kubeflow/website repo?

michelle192837:
I believe that's the case! The error is reported by hook, specifically for https://github.com/kubeflow/website/pull/715

PR #777, opened 2019-06-06
Fixed path in gitattributes file

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/716

I've checked the [docs](https://git-scm.com/docs/gitattributes), which state:

> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash `path/` syntax is pointless in an attributes file; use `path/**` instead)

I've also tested the new setting by making a change to the site's CSS, which resulted in a change in the files under `/resources` - it seemed to work as expected. I'll keep an eye on the diff/merge behaviour for CSS-related changes for a while.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/777)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 # Ignore the resources files in diffs and merge conflicts
-resources/    linguist-generated=true
-resources/    -diff -merge
+resources/**    linguist-generated=true
+resources/**    -diff -merge


sarahmaddox:
/assign @joeliedtke 
FYI @michelle192837

PR #794, opened 2019-06-11
testing table element styling

kbhawkey:
Again, experimenting with table styling and production site generation.
Updated a single content file as a test.
Built locally with hugo 0.55 extended. Preview was successful.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/794)

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -19,6 +19,28 @@ $navbar-dark-color: rgba($white, 1);
 $navbar-dark-hover-color: rgba($white, 0.75);
 $navbar-dark-active-color: $navbar-dark-color;
 
+.table-responsive, .td-content > table {
+    border: 1px solid #cccccc;
+    padding: 0;
+
+    td {
+      background-color: $secondary;
+      text-align: left;
+      width: 50%;
+    }
+
+    thead, th {
+      background-color: #e2e5eb;
+    }
+}
+
+.td-content {
+  h3 code {
+    color: #397af3;
+  }
+}
+
+
 // Front page styling
 
 .section-head {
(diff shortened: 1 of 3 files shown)


sarahmaddox:
@kbhawkey I'm putting a hold on this PR, as the description implies that it's a test only. Please do correct me if I'm wrong about that!

/hold

kbhawkey:
> 
> 
> @kbhawkey I'm putting a hold on this PR, as the description implies that it's a test only. Please do correct me if I'm wrong about that!
> 
> /hold

Absolutely fine -- I should have marked this PR a WIP. I thought I would try further to sort out what is happening to the generated assets. I wonder if main.css is not pulling in the updated dependency. My test markdown change built correctly.

sarahmaddox:
I'm not sure which tables in particular you're re-styling. Is it the tables in the generated reference docs? I was thinking the best way to tackle those is at time of generation. So, something like writing a wrapper script that runs the (existing) generation scripts, then grabs the output HTML and changes all the styles to use the Bootstrap table styling that I mentioned in issue https://github.com/kubeflow/website/issues/680

Is that more or less how you're looking at things too?

kbhawkey:
> 
> 
> I'm not sure which tables in particular you're re-styling. Is it the tables in the generated reference docs? I was thinking the best way to tackle those is at time of generation. So, something like writing a wrapper script that runs the (existing) generation scripts, then grabs the output HTML and changes all the styles to use the Bootstrap table styling that I mentioned in issue #680
> 
> Is that more or less how you're looking at things too?

I was hoping to view and test a draft of the reference doc tables and at the same time figure out the css generation. I agree that the reference docs script could be wrapped or updated, instead of a more general approach. Thanks.

Issue #806, opened 2019-06-14
The "Create issue" feature should auto-include doc URL


sarahmaddox:
The "Create issue" feature should auto-include the relevant doc URL in the created issue, but I suspect it doesn't. See this issue, which I suspect was created when the user clicked "Create issue": https://github.com/kubeflow/website/issues/804

It's possible that Docsy doesn't inject the doc URL - if so, raise a Docsy issue. Alternatively, there may be a particular site config in Kubeflow that's preventing the successful passing of the doc URL.

Start investigating in this file: `layouts/partials/page-meta-links.html `

sarahmaddox:
The issue creation may also be happening via the feedback widget.

sarahmaddox:
I've raised an issue for Docsy:
https://github.com/google/docsy/issues/91

PR #807, opened 2019-06-16
Added doc URL to 'Create issue' option

sarahmaddox:
Added a site-specific version of the `layouts/partials/page-meta-links.html` file, so that I can add code to inject the page URL into any issue created when the user clicks the **Create issue** option on a doc page. In the file, line 9 is new and file 10 is adjusted to add the `body` of the issue:

```
{{$doc_url := printf "https://www.kubeflow.org%s" .URL }}
{{ $issuesURL := printf "%s/issues/new?title=%s&body=%s" $gh_repo (htmlEscape $.Title) (htmlEscape $doc_url) }}
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/807)

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -1,0 +1,15 @@
+{{ if .Path }}
+{{ $gh_repo := ($.Param "github_repo") }}
+{{ if $gh_repo }}
+<div class="td-page-meta ml-2 pb-1 pt-2 mb-0">
+{{ $editURL := printf "%s/edit/master/content/%s" $gh_repo .Path }}
+{{ if .Site.IsMultiLingual }}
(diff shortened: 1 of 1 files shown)


sarahmaddox:
Preview:
https://deploy-preview-807--competent-brattain-de2d6d.netlify.com/docs/started/getting-started/

Click **Create issue** on the above page (or any page) and check the URL in the body of the new issue.

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/806

Note that this is a "good enough" fix, in that it hard-codes the domain into the URL. We need a more flexible fix to the Docsy theme, which should cater for different site configurations as mentioned in https://github.com/google/docsy/issues/91

Also as mentioned in the above issue, we need a similar fix for the Feedback widget.

abhi-g:
Good to lear the interesting hugo snippet.

/lgtm
/approve

PR #861, opened 2019-07-04
existing cluster: existing_arrikto: update docs on options for installing on existing K8s clusters

yanniszark:
The `existing_arrikto` platform and config file is a curated way to deploy Kubeflow on existing clusters using no cloud-specific features.

This PR updates the `Kubeflow on Kubernetes` page with the current working options for deploying Kubeflow on an existing Kubernetes cluster.

cc @sarahmaddox @jlewi 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/861)

### DIFF
--- a/content/docs/started/getting-started-k8s.md
+++ b/content/docs/started/getting-started-k8s.md
@@ -79,15 +79,15 @@ Follow these steps to deploy Kubeflow:
     tar -xvf kfctl_<release tag>_<platform>.tar.gz
     ```
 
-   Alternatively, you can easily build the `kfctl` binary yourself:
+    Alternatively, you can build the `kfctl` binary yourself:
 
     ```
     git clone https://github.com/kubeflow/kubeflow.git
     cd kubeflow/bootstrap
     make build-kfctl-container
     ```    
-    
-2. Run the following commands to set up and deploy Kubeflow. The code below includes an optional command to add the binary `kfctl` to your path. If you don't add the binary to your path, you must use the full path to the `kfctl` binary each time you run it.
+
+1. Run the following commands to set up and deploy Kubeflow. The code below includes an optional command to add the binary `kfctl` to your path. If you don't add the binary to your path, you must use the full path to the `kfctl` binary each time you run it.
 
 ```bash
 # Add kfctl to PATH, to make the kfctl binary easier to use.
@@ -112,6 +112,7 @@ kfctl apply all -V
   The value of this variable cannot be greater than 25 characters. It must
   contain just the directory name, not the full path to the directory.
   The content of this directory is described in the next section.
+
 
 ### Accessing Kubeflow
 
--- a/content/docs/started/getting-started-k8s.md
+++ b/content/docs/started/getting-started-k8s.md
@@ -4,84 +4,156 @@ description = "Instructions for installing Kubeflow on your existing Kubernetes 
 weight = 4
 +++
 
-Follow these instructions if you want to install Kubeflow on an existing Kubernetes
-cluster.
+Follow these instructions if you want to install Kubeflow on an existing Kubernetes cluster.
 
-If you are using a Kubernetes distribution or Cloud Provider which has specific
-instructions for installing Kubeflow we recommend following those instructions.
-Those instructions do additional Cloud specific setup to create a really great 
-Kubeflow experience.
+If you are using a Kubernetes distribution or Cloud Provider which has specific instructions for installing Kubeflow we recommend following those instructions. Those instructions do additional Cloud specific setup to create a really great Kubeflow experience.
 
-You can use either [kind](https://github.com/kubernetes-sigs/kind#installation-and-usage) or
-[Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) to install a Kubernetes cluster and it
-is recommended to use [kind](https://github.com/kubernetes-sigs/kind#installation-and-usage)
-as it is more simple to install.
+For installing Kubeflow on an existing Kubernetes Cluster, you can use one of the following options:
 
-Before installing Kubeflow on the command line:
+* [Kubeflow for Existing Clusters by Arrikto](#Kubeflow-for-Existing-Clusters---by-Arrikto)
 
-  * Ensure you have installed the following tools:
-    
-     * [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-     * [ks] (https://github.com/ksonnet/ksonnet/releases/)
+## Kubeflow for Existing Clusters - by Arrikto
 
+This installation of Kubeflow is maintained by [Arrikto](https://www.arrikto.com/), is geared towards existing Kubernetes clusters and does not depend on any cloud-specific feature.
 
-## Deploy Kubeflow
+In this reference architecture, we use [Dex](https://github.com/dexidp/dex) and [Istio](https://istio.io/) for vendor-neutral authentication.
+
+![platform existing architecture](https://i.imgur.com/OlaN73j.png)
+
+### Prerequisites
+- Kubernetes Cluster with LoadBalancer support.
+
+If you don't have a Kubernetes Cluster, you can create a compliant Kubernetes Engine (GKE) on Google Cloud Platform cluster with the following script:
+
+<details>
+
+<summary>GKE Cluster Creation Script</summary>
+
+```bash
+#!/bin/bash
+
+set -e
+
+# This script uses the gcloud command.
+# For more info, visit: https://cloud.google.com/sdk/gcloud/reference/container/
+
+# Edit according to your preference
+GCP_USER="$(gcloud config list account --format "value(core.account)")"
+GCP_PROJECT="$(gcloud config list project --format "value(core.project)")"
+GCP_ZONE="us-west1-b"
+
+CLUSTER_VERSION="$(gcloud container get-server-config --format="value(validMasterVersions[0])")" 
+CLUSTER_NAME="kubeflow"
+
+############################
+# Create and setup cluster #
+############################
+
+gcloud container clusters create ${CLUSTER_NAME} \
+--project ${GCP_PROJECT} \
+--zone ${GCP_ZONE} \
+--cluster-version ${CLUSTER_VERSION} \
+--machine-type "n1-standard-8" --num-nodes "1" \
+--image-type "UBUNTU" \
+--disk-type "pd-ssd" --disk-size "50" \
+--no-enable-cloud-logging --no-enable-cloud-monitoring \
+--no-enable-ip-alias \
+--enable-autoupgrade --enable-autorepair
+
+echo "Getting credentials for newly created cluster..."
+gcloud container clusters get-credentials ${CLUSTER_NAME} --zone=${GCP_ZONE}
+
+echo "Setting up GKE RBAC..."
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=${GCP_USER}
+```
+
+</details>
+
+### Deploy Kubeflow
 
 Follow these steps to deploy Kubeflow:
 
-1. Download a `kfctl` release from the 
-  [Kubeflow releases page](https://github.com/kubeflow/kubeflow/releases/).
-
-1. Unpack the tar ball:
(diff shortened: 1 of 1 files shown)


ryandawsonuk:
Would be great to see this in... actually back in as it was merged before a bit prematurely and had to be reverted at that point.

sarahmaddox:
/assign

For the record: This PR is a revival of PR https://github.com/kubeflow/website/pull/833. We had to revert the latter as it was merged prematurely. Now that PR https://github.com/kubeflow/kubeflow/pull/3538 is merged, we can look to merging the current PR (https://github.com/kubeflow/website/pull/861).

Preview: https://deploy-preview-861--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

content/docs/started/getting-started-k8s.md
```diff
@@ -4,84 +4,156 @@ description = "Instructions for installing Kubeflow on your existing Kubernetes
...
-instructions for installing Kubeflow we recommend following those instructions.
-Those instructions do additional Cloud specific setup to create a really great 
-Kubeflow experience.
+If you are using a Kubernetes distribution or Cloud Provider which has specific instructions for installing Kubeflow we recommend following those instructions. Those instructions do additional Cloud specific setup to create a really great Kubeflow experience.
 
-You can use either [kind](https://github.com/kubernetes-sigs/kind#installation-and-usage) or
-[Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) to install a Kubernetes cluster and it
-is recommended to use [kind](https://github.com/kubernetes-sigs/kind#installation-and-usage)
-as it is more simple to install.
+For installing Kubeflow on an existing Kubernetes Cluster, you can use one of the following options:
```
sarahmaddox: This paragraph combined with the single entry in the list below implies that there's only one option for deploying Kubeflow to existing clusters, and that that option is maintained by Arrikto.

This content was not in the original PR https://github.com/kubeflow/website/pull/833/files that I reviewed a while ago, and I'm wondering if I've missed some conversations.

I'm concerned about saying that this particular solution is maintained by a third party rather than by the Kubeflow community.

It's also not clear why this particular solution is special and needs to be maintained by a particular group. This document should describe the recommended solution, without giving people cause to think that the solution is possibly a special case rather than the recommended one.

@jlewi and @abhi-g and @yanniszark do you have any input about the reason for adding the special mention of Arrikto into the docs?

content/docs/started/getting-started-k8s.md
```diff
@@ -4,84 +4,156 @@ description = "Instructions for installing Kubeflow on your existing Kubernetes
...
-    # The following command is optional, to make kfctl binary easier to use.
-    export PATH=$PATH:<path to kfctl in your kubeflow installation>
-
-    export KFAPP=<your choice of application directory name>
-    # Default uses IAP.
-    kfctl init ${KFAPP}
-    cd ${KFAPP}
-    kfctl generate all -V
-    kfctl apply all -V
+   Alternatively, you can easily build the `kfctl` binary yourself:
```
sarahmaddox: Remove the word "easily". It's is tech writing best practice to avoid words like "easily" and "simply", as these words can cause frustration and unhappiness for the reader when the solution turns out not to be simple.

content/docs/started/getting-started-k8s.md
```diff
@@ -4,84 +4,156 @@ description = "Instructions for installing Kubeflow on your existing Kubernetes
...
-     The value of this variable becomes the name of your deployment.
-     The value of this variable cannot be greater than 25 characters. It must
-     contain just the directory name, not the full path to the directory.
-     The content of this directory is described in the next section.
+    git clone https://github.com/kubeflow/kubeflow.git
+    cd kubeflow/bootstrap
+    make build-kfctl-container
+    ```    
+    
+2. Run the following commands to set up and deploy Kubeflow. The code below includes an optional command to add the binary `kfctl` to your path. If you don't add the binary to your path, you must use the full path to the `kfctl` binary each time you run it.
```
sarahmaddox: The list formatting is broken in the rendered view - see preview: https://deploy-preview-861--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

To fix this, you should indent the preceding paragraphs so that the renderer includes them in the previous bullet point.

In the Markdown for numbered lists, please use a sequence of `1.`s rather than `1.` and `2.` etc. The former is easier to maintain when we want to add/remove bullet points. The rendered result is the same in both cases. Example:

```
1. First item.
1. Second item.
1. Third item.
```
The above yields this list:
1. First item.
1. Second item.
1. Third item.

content/docs/started/getting-started-k8s.md
yanniszark: Indeed, the changes in the PR are a result of the conversations in https://github.com/kubeflow/kubeflow/pull/3538.

To sum up:

After discussing with @jlewi we agreed on the following:
Existing, as a name, is too general and there can be multiple ways for deploying on existing clusters.
Some users might want something different from what we currently do (eg to bring their own Istio).
In addition, we agreed that since there isn't 1 correct way to deploy on existing clusters, each solution should have an arbitrating party, to clearly communicate whose opinions are represented.

> This paragraph combined with the single entry in the list below implies that there's only one option for deploying Kubeflow to existing clusters, and that that option is maintained by Arrikto.

To the best of my knowledge, at this particular time, it is the only config file specifically tailored for deploying on an existing Kubernetes Cluster and also the only one that supports authentication.

> I'm concerned about saying that this particular solution is maintained by a third party rather than by the Kubeflow community.

[From our discussion with @jlewi](https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-506771292), I understand that the current platforms (GCP, AWS) are arbitrated by Google and Amazon respectively, not the community.
While the whole community contributes to them, ultimately there is an arbitrating group who decides the contents of the config file.

That is the case as well with `existing_arrikto`.
Since there are many possible solutions and we are following a particular one, this config represents our opinion and as @jlewi suggested, that should be clear in the name and documentation.

> It's also not clear why this particular solution is special and needs to be maintained by a particular group.

Because there is not 1 universal solution, this solution represents a reference architecture targeted at existing clusters and providing auth functionality that wasn't available for existing clusters yet.
Since there are many solutions, we have to make some assumptions. These assumptions and opinions are why this platform is arbitrated by Arrikto.
As @jlewi put it: 

> Pick a name e.g. "arrikto" that clearly conveys whose oppinions are represented and who is the arbitrator (and message it as such)

With that said, anyone from the community can contribute and there have already been contributions from outside Arrikto (eg https://github.com/kubeflow/kubeflow/pull/3607)

cc @jlewi

content/docs/started/getting-started-k8s.md
yanniszark: This makes a lot of sense, I will rephrase it asap.
Do we have a guide for writing best-practices?

This is great advice and we should definitely document it.
Maybe as a Pull Request template ?

content/docs/started/getting-started-k8s.md
yanniszark: Thanks for catching that and helping me solve it, I will fix it asap.

content/docs/started/getting-started-k8s.md
sarahmaddox: Thank you for the clarifications! This LGTM.

content/docs/started/getting-started-k8s.md
sarahmaddox: Thanks! We do have a style guide here:
https://www.kubeflow.org/docs/about/style-guide/

If you'd like to create a PR to add a section on words like "easily", "simply", and "just do something" that'd be great.

The above Kubeflow style guide suggests a [fuller style guide](https://developers.google.com/style/) which does have a section on [words like "easy"](https://developers.google.com/style/tone#some-things-to-avoid-where-possible). It's probably still worth adding the info to the Kubeflow page, to make it easier to find.

sarahmaddox:
Thanks!

/lgtm
/approve

fberanizo:
Not sure if it's right place to tell, but I had a problem with this instruction:

https://github.com/kubeflow/website/commit/3a307d0a7689ee0863b83aa3de0d4ad6922d99db#diff-fe297c0331c2049d61d5c1972e94abfaR102

If I noticed it right, `--config` isn't part of a release yet. I have 0.5.1 installed (latest release so far) and can't run this command as described.

Also, I had other issue related to Kubernetes 1.15 (https://github.com/kubeflow/kubeflow/issues/3544), and these docs don't say anything about current restrictions on kubernetes <=1.14.x

Thanks! :)

yanniszark:
@fberanizo it seems a kfctl release with --config is not out yet.
You can alternatively build `kfctl` yourself with `make build-kfctl-container`.
Let's open an issue for this.

PercyLau:
hi, after following the latest instructions about installing kubeflow upon existing K8s clusters, I am stuck with getting istio IP addresses:
INFO[0064] Trying to get istio-ingressgateway Service External IP  filename="existing_arrikto/existing.go:392"

I use kind 0.4.0 to install an HA k8s cluster with dockers. Please find my kind configuration file as the follows:
======================
# a cluster with 3 control-plane nodes and 3 workers
kind: Cluster
apiVersion: kind.sigs.k8s.io/v1alpha3
nodes:
- role: control-plane
- role: control-plane
- role: control-plane
- role: worker
- role: worker
- role: worker
======================
Thanks in advance.

ryandawsonuk:
@PercyLau The pre-requisites for this option say 'Kubernetes Cluster with LoadBalancer support'. Dex is being used to secure an external URL/IP and so the installation waits for that external LoadBalancer IP to become available. For a setup without LoadBalancer you're better to use a different option for now. Or you can use this option on a cluster that does support LoadBalancer - there's a GCP script included in the instructions. (The closest local setup I can think of would be running minikube with minikube tunnel but haven't tried it.)

PercyLau:
@ryandawsonuk Thank you for the comprehensive explanation. 

The current version of kind does not support load balancer. See https://github.com/kubernetes-sigs/kind/issues/411.  But kind seems create some fake load balancer because it prints the following bold messages on my desktop.

admin@server# kind create cluster --config kind-three-node-config.yaml --name kf                                                                                  Creating cluster "kf" ...
 ✓ Ensuring node image (kindest/node:v1.15.0) 🖼�
 ✓ Preparing nodes 📦📦📦📦📦📦📦
 ✓ **Configuring the external load balancer** ⚖️
 ✓ Creating kubeadm config 📜
 ✓ Starting control-plane 🕹️️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
 ✓ Joining more control-plane nodes 🎮
 ✓ Joining worker nodes 🚜
Cluster creation complete. You can now use the cluster with:

export KUBECONFIG="$(kind get kubeconfig-path --name="kf")"
kubectl cluster-info

BenTheElder:
👋
The external loadbalancer is not fake, it is _external_. This loadbalancer load balances the Kubernetes API servers when you create multiple control plane nodes.

It is not a Kubernetes "type loadbalancer" object because that would be a chicken and egg problem.

Type loadbalancer is generally only supported properly in the cloud.

Is there a reason nodePort or similar cannot be supported? cc @jlewi

yanniszark:
Hi @BenTheElder. 
NodePort support is something that will be available in followup PRs.
However, LoadBalancer support is not something that is cloud-only.
See [MetalLB](https://metallb.universe.tf/) and `minikube tunnel`.

BenTheElder:
Hi, MetalLB does work with kind, but I think only on linux. LoadBalancer is of course anywhere that implements it, but many clusters will not have it since Kubernetes does not officially ship one and it is not even in conformance. Generally it is most well supported in cloud...

IMO, it is a bit heavy for development purposes, but of course useful in production. NodePort support will be great :-)

https://mauilion.dev/posts/kind-metallb/

PercyLau:
Hi, @BenTheElder. The attached article is fascinating. Thx a lot.

PercyLau:
Hi, @BenTheElder with MetalLB, we can deploy kubeflow over a k8s cluster created by kind. However, another problem arises, i.e., I can ping the nodes in the k8s cluster but cannot curl any http nor https ports now. Neither does the simple echo server provided in [](https://mauilion.dev/posts/kind-metallb/). Very appreciated for your help in advance.

Issue #826, opened 2019-06-23
Add examples for serving a model with Kubeflow


sarahmaddox:
Request from the KFServing team: Add more examples and docs for serving models with Kubeflow. 

Comment from sarahmaddox: We do have a couple of tutorials in the docs that include serving, and some serving examples on GitHub. At a minimum, we can create a page that links all the serving-related info from one place.

isaacmg:
I can work on adding a couple of serving examples with Seldon.

sarahmaddox:
/assign @isaacmg 

Thanks @isaacmg that's great!

sarahmaddox:
We now have many serving examples described in or linked from these pages:
* https://www.kubeflow.org/docs/components/serving/seldon/
* https://www.kubeflow.org/docs/components/serving/kfserving/
* https://www.kubeflow.org/docs/components/serving/tfserving_new/

/close

PR #895, opened 2019-07-11
Updated non-working URL

isaacmg:
Related to #826 though really more of an existing edit at this point.

While reading through the PyTorch Serving example I discovered the old URL to the Seldon Docs was not working. This appears to be their updated python wrapper documentation which has links to using Seldon-Core with Python.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/895)


sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/900

Thanks! I've added an issue to track this fix during the doc sprint.

Issue #867, opened 2019-07-08
Remove ksonnet page from "Components of Kubeflow"


richardsliu:
This page is no longer necessary: https://www.kubeflow.org/docs/components/ksonnet/

We should consider adding a page about Kustomize.

scottilee:
@sarahmaddox I can work on removing the ksonnet page and adding a kustomize page that at the least explains what it is and how to install it.

richardsliu:
@scottilee Please feel free to reference this page: https://github.com/kubeflow/manifests/blob/master/README.md

scottilee:
@richardsliu Will do, will have a pull request later today.

sarahmaddox:
/assign @scottilee 
Great! Thanks @scottilee and @richardsliu

scottilee:
I made a pull request here: https://github.com/kubeflow/website/pull/913

@sarahmaddox or @richardsliu could you please review and give me some feedback?

richardsliu:
This is done.

PR #913, opened 2019-07-12
#867: Remove ksonnet page and add kustomize page

scottilee:
* Removed the ksonnet page (technically, I `git mv`ed it to kustomize.md page)
* Added content about what kustomize is and how to install it

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/913)


content/docs/components/kustomize.md
```diff
@@ -0,0 +1,84 @@
...
+      export PATH=$PATH:${HOME}/bin/kustomize
+      ```
+
+## More about kustomize
+
+kustomize lets you customize raw, template-free YAML files for multiple
+purposes, leaving the original YAML untouched and usable as is.
+
+kustomize directories can be built using `kustomize build <kustomization_directory>`
+and applied using `kustomize build <kustomization_directory> | kubectl apply -f -`.
```
kbhawkey: - Do you want to mention the role of kustomization.yaml in more detail, or how Kubeflow has developed manifests to configure and deploy Kubeflow components?
- Is kustomize directly manipulated by the Kubeflow user, or are the commands embedded in scripts/templates?
- The directories contain manifest files (kustomization.yaml); the command takes the top level directory in the configuration file hierarchy. Does the user directly edit the manifest files to change the way Kubeflow is deployed?
- Do you want to provide a link to the Kubeflow manifest examples repo?

richardsliu: +1, please include a link to https://github.com/kubeflow/manifests/blob/master/README.md

scottilee: I tried adding some of this detail but I don't have any hands on experience so it was difficult to add all the information. I'd appreciate any help!

richardsliu: I think it is ok to leave out some of these details. The kfctl tool acts as a wrapper around the kustomize build commands, so we don't need to explain how kustomization.yaml is organized.

The user only needs to change manifest files if further customization is needed, in which case this page should help: https://www.kubeflow.org/docs/gke/customizing-gke/

scottilee: Thanks @richardsliu I added that link.

PR #942, opened 2019-07-17
Update oauth setup instructions and delete modeldb instructions

jlewi:
* With universal redirect URI you need to first create the OAuth credential
  so you know the client id

* Then you need to go back and set the universal IAP redirect URI

* Delete the ModelDB instructions because they are using ksonnet and they
  are outdated.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/942)


jlewi:
Need to verify it renders correctly

jlewi:
/assign @lluunn 

Formatting looks good 
https://deploy-preview-942--competent-brattain-de2d6d.netlify.com/docs/gke/deploy/oauth-setup/

PR #944, opened 2019-07-17
AWS_REGION to REGION

chie8842:
I found that we use `REGION` env parameter for aws_region at the document, but one command use `AWS_REGION` which isn't declared.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/944)


sarahmaddox:
/assign
/ok-to-test
/lgtm
/approve

Thanks @chie8842

PR #957, opened 2019-07-21
Add namespace credentials setup steps

### LINKED ISSUE #946: [GCP] Describe creating service accounts and using pod defaults
0.6 is designed to allow users to consume Kubeflow in their own namespace.

* When they create a new namespace; they will need to install GCP credentials into that namespace. This means running a bunch of gcloud and kubectl commands to create a service account, download the key and install it in the cluster.
* They will also want to create a PodDefault object in order to automatically attach the credentials to certain pods.

The relevant commands are documented [here](https://docs.google.com/document/d/1zT1yhDgmS59_uMWJ76hr7FCyXTDQ1Ju1MVADnlCfNEQ/edit#heading=h.dchzjmu0p6dp)

Related to #679 troubleshooting guide for credentials

scottilee:
Fix #946 I am adding steps to setup credentials for namespaces consuming Kubeflow.

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/957)

### DIFF
--- a/content/docs/gke/authentication.md
+++ b/content/docs/gke/authentication.md
@@ -16,7 +16,7 @@ The credentials to this service account can be accessed within the cluster as a
 The secret will have basic access to a limited set of GCP services by default, but more roles can be granted through the
 [GCP IAM console](https://console.cloud.google.com/iam-admin/).
 
-### Namespaces
+### Distributing Secrets
 
 Starting from Kubeflow version 0.6, you can consume Kubeflow in individual namespaces. The `kubeflow` namespace is
 intended for running Kubeflow system components while individual jobs and model deployments are expected to run in
@@ -67,7 +67,12 @@ kubectl create secret generic user-gcp-sa -n $NAMESPACE \ --from-file=user-gcp-s
 
 ##### PodDefault Object
 
-Create a pod default in a file called `add-gcp-secret.yaml` and apply it using: `kubectl apply -f add-gcp-secret.yaml -n $NAMESPACE`
+The PodDefault object is a way to centrally manage configurations that should be added to all pods.
+
+The PodDefault will match all pods with the specified selector and modify the pods to inject the volumes,
+secrets, and environment variables listed in the pod manifest.
+
+Create a pod default in a file called `add-gcp-secret.yaml` and apply it using: `kubectl apply -f add-gcp-secret.yaml -n $NAMESPACE`.
 ```
 apiVersion: "kubeflow.org/v1alpha1"
 kind: PodDefault
--- a/content/docs/gke/authentication.md
+++ b/content/docs/gke/authentication.md
@@ -183,6 +183,79 @@ By default, `Roles` and `RoleBindings` apply only to resources in a specific nam
 You can find more information in the 
 [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole).
 
+### Namespaces
+
+Starting from version 0.6, you can consume Kubeflow in individual namespaces. In order to do this you will need to
+install GCP credentials into the new namespace and create a PodDefault object to attach the credentials to certain
+pods.
+
+##### Credentials
+
+You can add credentials to the new namespace by either copying them from an existing Kubeflow namespace or by
+creating a new service account.
+
+To copy credentials from one namespace to another namespace use the following CLI commands (**Note**: there is an
+[issue](https://github.com/kubeflow/kubeflow/issues/3640) filed to automate this):
+
+```
+NAMESPACE=<new kubeflow namespace>
+SOURCE=kubeflow
+NAME=user-gcp-sa
+SECRET=$(kubectl -n ${SOURCE} get secrets ${NAME} -o jsonpath="{.data.${NAME}\.json}" | base64 -d)
+kubectl create -n ${NAMESPACE} secret generic ${NAME} --from-literal="${NAME}.json=${SECRET}"
+```
+
+To create a new service account instead of copying credentials, use the following steps:
+
+1. Create a service account with the desired roles
+```
+export PROJECT_ID=<GCP project id>
+export NAMESPACE=<new kubeflow namespace>
+export SA_NAME=<service account name>
+export GCPROLES=roles/editor
+gcloud --project=${PROJECT_ID} iam service-accounts create $SA_NAME
+gcloud projects add-iam-policy-binding $PROJECT_ID \
+    --member serviceAccount:$SA_NAME@$PROJECT_ID.iam.gserviceaccount.com \
+    --role $GCPROLES
+```
+
+2. Download the JSON service account key, set `KEYPATH` to the correct path, and create the key 
+```
+export KEYPATH=some/path/${SA_NAME}.gcp.json
+gcloud --project=${PROJECT_ID} iam service-accounts keys create ${KEYPATH} \
+   --iam-account $SA_NAME@$PROJECT_ID.iam.gserviceaccount.com
+```
+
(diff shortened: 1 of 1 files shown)


content/docs/gke/authentication.md
```diff
@@ -183,6 +183,79 @@ By default, `Roles` and `RoleBindings` apply only to resources in a specific nam
 You can find more information in the 
 [Kubernetes docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole).
 
+### Namespaces
```
jlewi: I think this belongs at the top of the page
https://deploy-preview-957--competent-brattain-de2d6d.netlify.com/docs/gke/authentication/

The first section is "in-cluster-authentication" and I think this text belongs there because it involves credentials from within a namespace.

The existing opening paragraph

> When you set up Kubeflow for GCP, it will automatically provision three service accounts with different privileges. In particular, the ${KFAPP}-user service account is meant to grant your user services access to GCP. The credentials to this service account can be accessed within the cluster as a Kubernetes secret.

Should probably be revised/replaced/merge with the paragraph you have here "starting from version 0.6".

content/docs/gke/authentication.md
```diff
@@ -183,6 +183,79 @@ By default, `Roles` and `RoleBindings` apply only to resources in a specific nam
...
+
+Starting from version 0.6, you can consume Kubeflow in individual namespaces. In order to do this you will need to
```
jlewi: HOw about adding a bit more detail explaining the namespace model.

i.e. explain that namespace "kubeflow" is intended for running kubeflow system components and users are expected to run individual jobs and model deployments in their own namespace.

scottilee: Done.

content/docs/gke/authentication.md
scottilee: I moved it to the top but I'm not clear as to why it would be revised with that opening paragraph since it seems like the information stands on its own?

content/docs/gke/authentication.md
```diff
@@ -7,15 +7,89 @@ weight = 4
...
+```
+
+3. Upload the JSON service account key to cluster as a secret
+```
+kubectl create secret generic user-gcp-sa -n $NAMESPACE \ --from-file=user-gcp-sa.json=${KEYPATH}
+```
+
+##### PodDefault Object
+
+Create a pod default in a file called `add-gcp-secret.yaml` and apply it using: `kubectl apply -f add-gcp-secret.yaml -n $NAMESPACE`
```
DanSanche: Can you explain the PodDefault object a bit here? Why it's needed/what it does

content/docs/gke/authentication.md
```diff
@@ -7,15 +7,89 @@ weight = 4
...
 [provision three service accounts](https://www.kubeflow.org/docs/gke/deploy/deploy-cli/#gcp-service-accounts) with different
-privileges. In particular, the `${KFAPP}-user` service account is meant to grant your user services access to GCP. 
+privileges. In particular, the `${KFAPP}-user` service account is meant to grant your user services access to GCP.
 The credentials to this service account can be accessed within the cluster as a
 [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/).
 
 The secret will have basic access to a limited set of GCP services by default, but more roles can be granted through the
 [GCP IAM console](https://console.cloud.google.com/iam-admin/).
 
+### Namespaces
```
DanSanche: I don't know if Namespaces makes sense a a section title. I'd probably call it "Creating Secrets", and then add the information from the previous paragraph about how some secrets are added automatically, but you have to manually create them in new namespaces

content/docs/gke/authentication.md
```diff
@@ -7,15 +7,89 @@ weight = 4
...
+You can add credentials to the new namespace by either copying them from an existing Kubeflow namespace or by
+creating a new service account.
+
+To copy credentials from one namespace to another namespace use the following CLI commands (**Note**: there is an
+[issue](https://github.com/kubeflow/kubeflow/issues/3640) filed to automate these commands):
+
+```
+NAMESPACE=<new kubeflow namespace>
+SOURCE=kubeflow
+NAME=user-gcp-sa
```
DanSanche: Is this the right name? It was my understanding that the secret name is now ${KFAPP}-user

jlewi: @DanSanche the K8s secret should still be user-gcp-sa
https://github.com/kubeflow/kubeflow/blob/1c07de4d4825f381062c7d7091398b837113b7f1/bootstrap/cmd/bootstrap/app/saKey.go#L152

jlewi: See below and
https://github.com/kubeflow/kubeflow/blob/1c07de4d4825f381062c7d7091398b837113b7f1/bootstrap/cmd/bootstrap/app/saKey.go#L152

This should be correct

content/docs/gke/authentication.md
jlewi: Good idea:

I think the key ideas are

1. PodDefault will match all pods matched by the selector
1. PodDefault will modify the pod to inject the volume, secrets and environment variable listed in the PodDefault

This is a way to centrally manage config that should be added to all pods.

content/docs/gke/authentication.md
scottilee: How about "Distributing Secrets"?

content/docs/gke/authentication.md
scottilee: Done.

jlewi:
/lgtm
/approve

/hold 

@DanSanche This looks pretty good to me

content/docs/gke/authentication.md
```diff
@@ -7,15 +7,94 @@ weight = 4
...
 The credentials to this service account can be accessed within the cluster as a
 [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/).
 
 The secret will have basic access to a limited set of GCP services by default, but more roles can be granted through the
 [GCP IAM console](https://console.cloud.google.com/iam-admin/).
 
+### Distributing Secrets
+
+Starting from Kubeflow version 0.6, you can consume Kubeflow in individual namespaces. The `kubeflow` namespace is
+intended for running Kubeflow system components while individual jobs and model deployments are expected to run in
```
DanSanche: Overall LGTM too, but I still think the two paragraph about secrets can be combined. Here's what I suggest, but feel free to ignore

> ## In-cluster authentication

> ### Service Accounts and Secrets
> 
> When you [set up Kubeflow for GCP](/docs/gke/deploy), it will automatically
> [provision three service accounts](https://www.kubeflow.org/docs/gke/deploy/deploy-cli/#gcp-service-accounts) with different privileges in the `kubeflow` namespace. In particular, the `${KFAPP}-user` service account is meant to grant your user services access to GCP.
> The credentials to this service account can be accessed within the cluster as a
> [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/).
> 
> The secret will have basic access to a limited set of GCP services by default, but more roles can be granted through the
> [GCP IAM console](https://console.cloud.google.com/iam-admin/).
> 
> Starting from Kubeflow version 0.6, you can also consume Kubeflow from custom namespaces (i.e. namespaces other than `kubeflow`), but secrets will not be automatically provisioned in these namespaces. In order to authenticate from a custom namespace you will need to install GCP credentials into the new namespace and create
> a PodDefault object to attach the credentials to your pods.
> ...

>

scottilee: I added most of the info and I think it's at a good in-between.

DanSanche: looks good to me 👍

Issue #889, opened 2019-07-10
Refactor the Kubeflow getting-started section


sarahmaddox:
Restructure the entries into groups:
* Cloud
* Local
* On prem

DanSanche:
I opened a [PR addressing this](https://github.com/kubeflow/website/pull/888), but we couldn't get consensus on how to organize this section, so I'm closing it and hoping we can use this bug to discuss how to improve the Getting Started experience.

I think creating a clear path for new users through the Getting Started guide should be a top priority. As it stands, the number of docs in the getting started docs is overwhelming and hard to navigate, and it seems like there are more docs coming as every platform wants to be represented in this section.

 I see two options to improve this:

1. **use folders to split up the guides into buckets**, so new users can focus only on the use-case that matters to them. This was what I tried to implement, but it's difficult to categorize all existing docs in a way that is clear and makes all stakeholders happy.
2. **create a single getting started guide** based on a standard open platform like minikube that we can direct all new users through. This option would be ideal if we could provide a great experience, but I haven't had much luck getting Kubeflow set up locally, so I don't know how feasible this is. 

Let me know your thoughts

kbhawkey:
I agree with both points. I am new to Kubeflow and am familiar with k8s, minikube, etc. I have minikube and Kubeflow installed locally, but the generate all script refers to ksonnet. Will see if I can get this patched up.

carmine:
I'm curious if things fall along the lines of "Installers", "Appliances" and "Managed", or something like that.  For instance, with kubernetes, I think you'd characterize GKE, AKS, EKS as managed kubernetes offerings, and things like kubeadm or charmed kubernetes as installers / distributions.

Agree that a better organization is strongly desired.

I think it starts with either "Where do you want to deploy Kubeflow" or "How do you want to consume Kubeflow". For the latter Question, things break down like this:

- From a cloud provider
- From an installer
  - In a virtual machine
  - In a desktop
  - In a cloud
- From an appliance (ie pre-baked VM)

I don't love it. To stick with previous effort, ie how to get started:

- Single node solution
  - appliance
  - vm
  - desktop
- Multi node solution
  - public cloud
  - kubernetes

Finally, as another option:

- New to kubeflow? Quickstart guides ...
- Ready for kubeflow in production? Installer guides .. (leads to managed kubeflow, all-in-one, etc)

carmine:
[...]

PR #977, opened 2019-07-26
Fixing doc for KF installation on existing clusters

johnugeorge:
The default KF installation on existing clusters should not be opinionated.  Since we follow composable architecture, users can add and install any component to their installation later.  Auth is an optional component.  

This PR updates the config to kfctl_k8s_istio.yaml(See https://github.com/kubeflow/kubeflow/pull/3663) and moved the auth specific section to a separate `Kubeflow Authentication` page.  This installs just KF components.  This also avoids confusion  regarding the KF installation requirements on existing kubernetes clusters. 
Related issue: https://github.com/kubeflow/kubeflow/issues/3750 , https://github.com/kubeflow/kubeflow/issues/3737

/assign @yanniszark 
/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/977)


jlewi:
Preview for Kubeflow on Kubernetes
https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

Preview for authentication guide
https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/other-guides/authentication/

content/docs/other-guides/authentication.md
```diff
@@ -0,0 +1,514 @@
++++
+title = "Kubeflow Authentication with Istio and Dex"
+description = "Instructions for installing Kubeflow with authentication support"
+weight = 4
++++
+
+This installation of Kubeflow is maintained by [Arrikto](https://www.arrikto.com/), is geared towards existing Kubernetes clusters and does not depend on any cloud-specific feature.
```
jlewi: Maybe drop by "maintained by [Arrikto]"
These instructions are clearly about Istio and Dex. So I'm hoping there's a general consensus around what this should be doing.

content/docs/other-guides/authentication.md
```diff
@@ -0,0 +1,514 @@
...
+
+In this reference architecture, we use [Dex](https://github.com/dexidp/dex) and [Istio](https://istio.io/) for vendor-neutral authentication.
+
+![platform existing architecture](https://i.imgur.com/OlaN73j.png)
+
+### Prerequisites
+- Kubernetes Cluster with LoadBalancer support.
```
jlewi: @yanniszark @johnugeorge Can we link to K8s docs describing this?

content/docs/other-guides/authentication.md
```diff
@@ -0,0 +1,514 @@
...
+
+If you don't have a Kubernetes Cluster, you can create a compliant Kubernetes Engine (GKE) on Google Cloud Platform cluster with the following script:
```
jlewi: Can we delete this? We already have GCP specific instructions that help people create a GKE cluster.

Can we refer people to
The K8s docs if they need help with creating a K8s cluster?

content/docs/other-guides/authentication.md
```diff
@@ -0,0 +1,514 @@
...
+
+Follow these steps to deploy Kubeflow:
+
+1. Download a `kfctl` release from the [Kubeflow releases page](https://github.com/kubeflow/kubeflow/releases/) and unpack it:
+
+    ```
+    tar -xvf kfctl_<release tag>_<platform>.tar.gz
+    ```
+
+    Alternatively, you can build the `kfctl` binary yourself:
```
jlewi: Should we drop the build kfctl yourself?
Building it yourself should be an advanced use case.
We should also have a single page with those instructions (or a markdown doc) so that we don't have multiple sets of instructions to update if it changes.

jlewi:
Thanks @johnugeorge 

@yanniszark WDYT?

johnugeorge:
Updated with review comments.  Please take a look. 
@jlewi @yanniszark 

Preview for Kubeflow on Kubernetes
https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

Preview for authentication guide
https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/other-guides/authentication/

content/docs/other-guides/authentication.md
johnugeorge: done

content/docs/other-guides/authentication.md
johnugeorge: done

content/docs/other-guides/authentication.md
johnugeorge: done

content/docs/other-guides/authentication.md
johnugeorge: done

johnugeorge: done

vkoukis:
Johnu, Jeremy, thanks for taking the time to propose potential fixes to the installing on existing K8s page.

@johnugeorge :
> The default KF installation on existing clusters should not be opinionated. Since we follow composable architecture, users can add and install any component to their installation later. Auth is an optional component.

Actually, we have already established with Jeremy, in multiple discussions, that installing on an existing cluster is opinionated by its very nature.
See kubeflow/kubeflow#3538 for all of the context, and more specifically [this comment](https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-506365781)
and Jeremy's [initial response](https://kubeflow.slack.com/archives/CF94DURGF/p1559836215017900?thread_ts=1559700461.000900&cid=CF94DURGF) on the Kubeflow Slack:

@jlewi :
> Instead of using a new flag introduce a new platform. The point of a platform is too have an oppionated way of installing Kubeflow with a clear set of owners who are the deciders of what those opinions should be.

It was with this in mind that we agreed with @jlewi to create a generic config file. This PR seems to be removing the `existing_arrikto.yaml` config file as an option from the "deploying on existing kubernetes" page altogether, and only leaves a single, limited configuration, `kfctl_k8s_istio.yaml`, as the only option, although we had explicitly established that there can't be a single way of doing things in such a broad area, see [comment by @jlewi](https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-506771292):
> From a user's point of view, I don't think a user would expect any of these values to correspond to generic, consensus approved configs.

and

> The challenge of a generic reference architecture is who gets to decide what assumptions are the right one?

The purpose of this page was never to have a one-size-fits-all config file.
We started with `existing_arrikto.yaml` as the only choice, because there was no other way to deploy Kubeflow easily on an existing, on-prem K8s cluster.

I understand that the current approach of using a LoadBalancer instance causes friction, and it makes sense to add `kfctl_k8s_istio.yaml` as a simpler config file that does not have any auth / multi-user support at all.

But this should not in any way be the *only* proposed solution.
Instead, we should expose multiple options, and make sure to explain exactly what they require as a starting point and what the end state of the cluster will be after following them.

@johnugeorge :
>This also avoids confusion regarding the KF installation requirements on existing kubernetes clusters.
> Related issue: kubeflow/kubeflow#3750 , kubeflow/kubeflow#3737

On the specific issue of the LoadBalancer causing friction now, which seems to be the only significant one so far, we will explain that MetalLB is a fast and easy way to solve this. Also see [comment by @halayudha](https://github.com/kubeflow/kubeflow/issues/3737#issuecomment-514673163):
> hi, emma, as suggested by yanniszark, just deploy metallb, deploy its configmap, the installation will be good...i got mine working now

Since @yanniszark is on PTO this week, I would be happy to make this addition as part of this or a separate PR. Also, to remove any potential confusion, I can highlight the specific requirement of each option explicitly, so it is crystal-clear for users to see exactly what each config file needs to work.

@johnugeorge :
> and moved the auth specific section to a separate Kubeflow Authentication page.

@yanniszark 's section referring to the `existing_arrikto.yaml` config file is *not* an auth-specific section. It is a tested, end-to-end deployment guide, using a full configuration file. Currently, it focuses on enabling auth and multi-user support regardless of platform being used. So, I propose we keep it as is, as an option for deploying KF on existing clusters.

By making authentication available outside GCP, and enabling multi-user auth, we make the base install much more powerful, and bring it en par with what Kubeflow supports on GCP. In the process we also discover and fix things that may be GCP-specific, and/or avoid inserting new GCP-specific code in areas where we can be generic. This helps keep Kubeflow vendor-neutral. For example [in kfctl](https://github.com/kubeflow/kubeflow/pull/3693#pullrequestreview-265010439) and [in multi-user support](https://github.com/kubeflow/kubeflow/pull/3503#pullrequestreview-267202104).

@jlewi :
> Maybe drop by "maintained by [Arrikto]"
These instructions are clearly about Istio and Dex. So I'm hoping there's a general consensus around what this should be doing.

Jeremy, we included this because we had agreed on this explicitly, as a way to 
make explicit that this is an opinionated configuration file. From [your comment](https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-506771292):
> Pick a name e.g. "arrikto" that clearly conveys whose oppinions are represented and who is the arbitrator (and message it as such)

But the interesting thing is, that even as it stands now, it's not obvious that there is a general consensus around how to deploy Istio+Dex. We have already agreed, and we are working towards this direction to make Istio+Dex a separate component, to be installed alongside KF components. But to make redirects work, we had to make the opinionated decision of using a LoadBalancer instance, so we had an explicit URL to redirect to. We are exploring making `existing_arrikto.yaml` work without a LoadBalancer instance altogether, so it's simpler. But this will need *even* more opinionated decisions on how to perform redirections.

So, `existing_arrikto.yaml` is *not* about just installing the Istio+Dex component. It's about the specific way to configure it, so it works out of the box. With this in mind, I would propose we keep the config file in its current location, as one more option.

We are OK with removing this specific reference, `maintained by Arrikto`, if we make explicit exactly what each option/config file needs as input, and what it provides as the end state of the cluster.

Finally, as also discussed in https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-506365781 the `existing_arrikto.yaml` config file represents a different point-of-view compared to `kfctl_k8s_istio.yaml`: it aims to be a full, batteries-included deployment, and aims to enable full auth and multi-user support. To do this, we had to make opinionated decisions on how to expose the service (LoadBalancer), auth (Istio+Dex, which we will move into a distinct component as already agreed), SSL (currently self-signed HTTPS certificates, we will soon add support for cert-manager).

Also note, we already have a pending issue for improving current `existing_arrikto.yaml`, kubeflow/kubeflow#3739, and @yanniszark is already working on a design doc for a new iteration of kfctl, he will be submitting it early next week, when he returns from PTO.
Part of this design doc will be how to make the process more composable and extend kfctl to work with plugins: https://github.com/kubeflow/kubeflow/issues/3709

I am looking forward to ensuring most of the code between distinct config files is shared, and I understand moving towards plugins would help, but config files should remain separate, as distinct options in the Getting Started Page, as we have already established that there can be no single way of deploying on an existing cluster.

So, to summarize, let's *not* remove the existing_arrikto.yaml config file, it is designed explicitly to serve as an installation guide for existing clusters, and it seems to have served its purpose well so far. Of course, we will be watching all incoming issues and make fixes/improvements to streamline its applicability as much as possible.

Can we have this PR *add* `kfctl_k8s_istio.yaml` as a new option/section, explaining what it requires and what it provides, in the Getting Started page?

@johnugeorge, @jlewi, WDYT? Looking forward to your comments.

johnugeorge:
@vkoukis 
Thanks for the detailed reply.  

> Finally, as also discussed in kubeflow/kubeflow#3538 (comment) the existing_arrikto.yaml config file represents a different point-of-view compared to kfctl_k8s_istio.yaml: it aims to be a full, batteries-included deployment, and aims to enable full auth and multi-user support.

I differ in following aspects.
1. Kubeflow should never be a platform that mandate users to follow a specific way to enable any feature(auth or multi user support).   It is designed to be  a highly composable platform where users can select and install whatever components they wish to based on their env. So  dex should not be the only way to do auth, it should be seen as one of the ways to do it.  
2. Existing clusters can on-prem or on-cloud like GCP, Azure, AWS.   I see, auth installation to be more of a platform specific feature because each cloud provider have different auth implementations.  It cannot be generalized. eg: GKE on prem provides OIDC support  by itself. Similar with AKS.   For on-prem, it is a different story as requirements and restrictions are different.  There is more flexibility for auth in on-prem environments as API server config is accessible. This is not possible in case of the cloud deployments.
3. Debugging and config maintainability becomes easy when we stick just to KF components and their direct dependencies.  Not all users need auth or multi user support.  See  https://github.com/kubeflow/kubeflow/issues/3750  who wants to try just Katib component.  Individual feature owners will find it difficult to figure out the root cause in case of errors.  See https://github.com/kubeflow/katib/issues/696

>  To do this, we had to make opinionated decisions on how to expose the service (LoadBalancer), auth (Istio+Dex, which we will move into a distinct component as already agreed), SSL (currently self-signed HTTPS certificates, we will soon add support for cert-manager).

 It is true that `existing_arrikto` config provides a different point of view but I feel that direction that you are moving, fits more into a category of platform KF installation (similar to GCP,AWS etc).   LoadBalancer and SSL are not necessary for general KF deployments. This should be optional components. Users should be able to install it once and when they need it.

jlewi:
Thanks @vkoukis 

Perhaps we should discuss in a meeting; maybe tomorrow's community meeting?

As far as I can tell this PR is not getting rid of existing_arrikto.yaml it is expanding the docs to make our install process more composable

1) The first page
     https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

   * Describes how to deploy Kubeflow with ISTIO but not exposing it behind an LB or with Dex

2)  The 2nd page
     https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/other-guides/authentication/

    * Describes how to deploy Kubeflow with ISTIO + LB + Dex
    * This is still using existing_arrikto.yaml

Is the only question how to lay out this information?

On the [Kubeflow on Kubernetes](https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/) should we just list the different options

I think we should aim to provide instructions/support the following use cases

Create Cluster | Install Istio | Install Dex | Install KF Apps
-- | -- | -- | -- 
No | Yes | Yes | Yes 
No | Yes | No | Yes
No | No | Yes | Yes
No | No | No | Yes

Do we generally agree that we should aim to support the above cases?

I think if we want to support all of those cases well then we need to move in the direction of making our deployment process as composable as possible. So ISTIO and Dex are ideally just "applications" or "plugins" that users can enable/disable depending on their use case.

I think we should to move to a world where  ISTIO and Dex are just applications listed in KfDef.Spec.Applications and users just include them or don't if they don't want to install them.

This is consistent with how users would list which KF applications to install or not install.

There is some complexity because as @yanniszark discovered some of the Dex configuration isn't known until after some resources are created.

Nonetheless do we broadly agree on the following

1. We'd like to "support" all of the use cases listed in the table above
1. We'd like to move towards a composable spec based on KfDef.Spec.Applications and KfDef.Spec.Plugins

vkoukis:
@jlewi :
> As far as I can tell this PR is not getting rid of existing_arrikto.yaml it is expanding the docs to make our install process more composable
> 
>  1. The first page
>        https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/
>        
>        * Describes how to deploy Kubeflow with ISTIO but not exposing it behind an LB or with Dex
> 
> 2. The 2nd page
>        https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/other-guides/authentication/
>        
>       * Describes how to deploy Kubeflow with ISTIO + LB + Dex
>       * This is still using existing_arrikto.yaml
> 
> Is the only question how to lay out this information?
> 
> On the [Kubeflow on Kubernetes](https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/) should we just list the different options
> 

Yes, list all options, equivalently. Config `kfctl_k8s_istio.yaml` is one of these options. We have agreed, and @yanniszark is putting in effort towards this, to make everything into composable components.
But *still*, there will be pre-determined config files, each detailing an opinionated way to deploy said components.

Some config files will be more on the vanilla side, e.g., will only deploy (some of) KF's components, others will be richer, e.g., will deploy Istio, and/or authservice + Dex behind a LoadBalancer.

The present discussion is about the different config files / options that will be available to users.
These options are all at the same level, so they should all be presented in a flat list.
This is the current format of the Getting Started page, and I argue we should leave it as is, and add `kfctl_k8s_istio.yaml` or any other similar config files / opinions.

Exactly because we have established there can be no single way of doing things, we should not treat a single config file differently, by placing it in the first page separately, or by moving other config files to different pages.

> I think we should aim to provide instructions/support the following use cases
> Create Cluster 	Install Istio 	Install Dex 	Install KF Apps
> No 	Yes 	Yes 	Yes
> No 	Yes 	No 	Yes
> No 	No 	Yes 	Yes
> No 	No 	No 	Yes
> 
> Do we generally agree that we should aim to support the above cases?
> 

Yes. I am not sure we should be defining every single variable in a matrix, because we may end up continuously having to define new columns. For example, what if a contributor wants to offer a solution that authenticates users with a different type of software / component / KfDef.spec.plugin, and use it in a proposed config. We would have to define a new "Install X" column, which could be cumbersome. Maybe just having a bullet list of requirements for each config file, and then a bullet list of expected actions ["Create Cluster", "Install Istio", "Install OIDC auth service + Dex for authentication", etc] would be a more flexible way of describing each opinionated config.

> I think if we want to support all of those cases well then we need to move in the direction of making our deployment process as composable as possible. So ISTIO and Dex are ideally just "applications" or "plugins" that users can enable/disable depending on their use case.
> 
> I think we should to move to a world where ISTIO and Dex are just applications listed in KfDef.Spec.Applications and users just include them or don't if they don't want to install them.
> 
> This is consistent with how users would list which KF applications to install or not install.
>

@yanniszark's recent work with kfctl has been pushing towards exactly this direction, and this is towards where he will be driving the design doc as well.
But the present discussion is not about making everything into a component, we have all agreed on this. This discussion is about proposed, pre-determined, opinionated, composable specs, which currently are config files for kfctl.
I think these config files should all appear in a flat list, in a single page, and we should make sure to explain exactly what each config file expects, and what it provides.
 
> There is some complexity because as @yanniszark discovered some of the Dex configuration isn't known until after some resources are created.
> 
> Nonetheless do we broadly agree on the following
> 
>  1. We'd like to "support" all of the use cases listed in the table above
> 
>  2. We'd like to move towards a composable spec based on KfDef.Spec.Applications and KfDef.Spec.Plugins

I think we are in broad agreement on this, yes.
This discussion is about exposing all config files in a flat way.

sarahmaddox:
Hallo all

Thanks for all the input - this is a very interesting discussion. It looks like we need to adjust our doc information architecture to account for multiple supported ways of deploying Kubeflow on K8s (which includes on prem and non-cloud-specific installations).

If you think it's useful, I'd like to put together a **doc plan** for discussion. 

Basically, I'm thinking along these lines for the structure of the docs:

* **Kubeflow on Kubernetes** (New doc section within the "getting started" section)

  * Page: **Overview**
    Describe purpose of this section (i.e. guides to installing Kf on existing Kf) plus a list of the supported configs with a short description of each installation option, plus a link to the details page. We currently have two installation options (simple, and multi-user auth) but there will be more in future.

     I think a list will work better than a table, because it's hard to predict what the columns will be in the table. But if someone has a good idea of those columns, or if you think we can start simple and adjust as we go along, then we can use a table. Similar to how [Kubernetes does it](https://kubernetes.io/docs/setup/).

  * Page: **Installation option 1**
    Describe the features of this installation option (simple, single user) and give detailed installation instructions.

  * Page: **Installation option 2**
  Describe the features of this installation option (multi-user and auth - this is the Arrikto solution) and give detailed installation instructions.

  * Future pages for more installation options.

  * Maybe move the existing **multi-node cluster** guide here too?

Some more points for discussion (probably best in a design doc, if you agree I should create one):

* We should move the **Kubeflow on Kubernetes** section/page to the top of the getting-started section. It's currently at the bottom.
* We need to think of a better name for **Kubeflow on Kubernetes**.
* Naming the individual installation options is going to be hard. Maybe just a straight numerical naming convention as I've done above.

What do you think - shall I put together a doc plan for further discussion?

vkoukis:
@sarahmaddox 
> Hallo all
> 
> Thanks for all the input - this is a very interesting discussion. It looks like we need to adjust our doc information architecture to account for multiple supported ways of deploying Kubeflow on K8s (which includes on prem and non-cloud-specific installations).
>
> If you think it's useful, I'd like to put together a **doc plan** for discussion.
>

Sarah, thank you  for putting together this proposal. I think it's in the right direction, acknowledges that there will be multiple supported ways of deploying Kubeflow, and provides details for each one of them independently. I understand these installation options will be distinct kfctl config files.

Comments follow inline.

> Basically, I'm thinking along these lines for the structure of the docs:
> 
>     * **Kubeflow on Kubernetes** (New doc section within the "getting started" section)
>       

Would it make sense to rename it to "Kubeflow on existing Kubernetes"? Kubeflow always runs on Kubernetes, so I think the title should point out what is different about this page compared to other pages in the docs, e.g., "Kubeflow on [generic] existing Kubernetes" vs. "[on Kubernetes] on GCP", "[on Kubernetes] on AWS/Azure", etc.

>       * Page: **Overview**
>         Describe purpose of this section (i.e. guides to installing Kf on existing Kf) plus a list of the supported configs with a short description of each installation option, plus a link to the details page. We currently have two installation options (simple, and multi-user auth) but there will be more in future.
>         I think a list will work better than a table, because it's hard to predict what the columns will be in the table. But if someone has a good idea of those columns, or if you think we can start simple and adjust as we go along, then we can use a table. Similar to how [Kubernetes does it](https://kubernetes.io/docs/setup/).

I would propose we use a list of all proposed kfctl config files.

>       * Page: **Installation option 1**
>         Describe the features of this installation option (simple, single user) and give detailed installation instructions.
>       * Page: **Installation option 2**
>         Describe the features of this installation option (multi-user and auth - this is the Arrikto solution) and give detailed installation instructions.
>       * Future pages for more installation options.
>       * Maybe move the existing **multi-node cluster** guide here too?
> 

Given that installation options are essentially kfctl config files, I think the existing "multi-node cluster" guide is not a good fit for this list. It seems to be more of a preparation step to deploy NFS-based shared storage  before choosing an installation option and installing Kubeflow. I think this becomes apparent at the place where it actually links back to the "Kubeflow on Kubernetes" page:

> Now that you have prepared your on-prem, multi-node Kubernetes cluster to manage volumes using an NFS server, you can install Kubeflow using the Kubeflow on Kubernetes guide.


So it would be strange to list this as an option, then have it link back to the list of options, creating a circular dependency. Maybe we can list it independently, at the bottom of the first page, as a related, optional preparation step?

> Some more points for discussion (probably best in a design doc, if you agree I should create one):
> 

Yes, please do.
I would be happy to participate in implementing this, and sorting out all the details. Please include @yanniszark and me in the loop.

>     * We should move the **Kubeflow on Kubernetes** section/page to the top of the getting-started section. It's currently at the bottom.
> 
>     * We need to think of a better name for **Kubeflow on Kubernetes**.
> 

Please see above.

>     * Naming the individual installation options is going to be hard. Maybe just a straight numerical naming convention as I've done above.
> 

Since the individual installation options will be opinionated kfctl config files, shall we use their actual names in the repository to refer to them? e.g., `kfctl_k8s_istio.yaml`. This would also create a 1-1 correspondence between what the user mentions in their command lines when they report issues ["I am running `kfctl --config ...` and bump into this"] , and what exists in the docs.

> 
> What do you think - shall I put together a doc plan for further discussion?

Please go ahead. Thank you again for your proposal.

johnugeorge:
Thanks @sarahmaddox for your proposal

@vkoukis 
In general, I fail to understand why dex cannot be treated as an composable component like any other component. Similar to istio component. This would avoid much confusion and effort in maintaining/testing multiple config files.  

However if you still want to maintain a separate config file, I don't see any problem in having flat list of options in the main page given,

1. Each config file has an owner and user knows whom to approach for any errors. It is expected that config owner is responsible for all issues related to KF installation and deployment(including individual KF components). Eg:  https://github.com/kubeflow/katib/issues/696

2 Users can clearly differentiate community supported vs Ecosystem configs( like in the link shared - [kubernetes  setup](https://kubernetes.io/docs/setup/)  Standard testing env would be a community supported config which is a vanilla KF deployment.

vkoukis:
@johnugeorge 
> However if you still want to maintain a separate config file, I don't see any problem in having flat list of options in the main page given,
> 
>  1. Each config file has an owner and user knows whom to approach for any errors. It is expected that config owner is responsible for all issues related to KF installation and deployment(including individual KF components). Eg:  [kubeflow/katib#696](https://github.com/kubeflow/katib/issues/696) 
>  2. Users can clearly differentiate community supported vs Ecosystem configs( like in the link shared - [kubernetes  setup](https://kubernetes.io/docs/setup/)  Standard testing env would be a community supported config which is a vanilla KF deployment.

We have been supporting deployments based on the `existing_arrikto` kfctl config, and will continue to do so. We are actively monitoring issues, but please feel free to involve @yanniszark , myself in issues that may be due to the specific config file used.

I agree with being able to differentiate the different config files by listing a different maintainer.

Regarding kubeflow/katib#696, I will try to reproduce and report back. Given that the internal katib-db service itself is inaccessible, it may be a different problem and unrelated to the config file.

If you agree with moving with a flat list of options, what would the next step be?
Could you modify this PR to add `kfctl_k8s_istio.yaml` and list both config files in the existing page?

In parallel, I am now working on providing the simplest possible instructions for deploying MetalLB, this should solve most problems with people trying the `existing_arrikto.yaml` config.

johnugeorge:
@vkoukis 

One thing to add is, it is not technically correct to say that kfctl_k8s_istio cannot support auth or multi user.  Once auth is added to manifests(See https://github.com/kubeflow/manifests/pull/195) ,  users can install them if needed based on their env. The aim in maintaining kfctl_k8s_istio is to make sure that there is a vendor neutral config which is verified by the community. This config  is not restrictive to users in any way.  If we follow the composable architecture, it should support all features.  In fact, extra sections should be added to the same page to describe the steps that are needed to enable each optional feature(from Kustomize manifests) like auth, storage integration for on-prem etc.  Related: https://github.com/kubeflow/kubeflow/issues/3774 , https://github.com/kubeflow/kubeflow/issues/3728#issue-471688389 

I feel that the right direction for the long term community health is to enforce composability design to all features.  From your earlier comments, I still couldn't figure out why you don't take this route of maintaining single config as Dex is an open platform? 

With respect to PR changes, there should be clear distinction between  community maintained and vendor maintained config. we could maintain a table(similar to earlier Kubernetes link) with two columns like `Community`(KF community supported) and Ecosystem('Vendor supported').  Under each header, we could point to the  corresponding config. 

What do you think?

sarahmaddox:
@vkoukis @johnugeorge @yanniszark @pdmack @jlewi 
I've created a doc which I hope combines all the comments that we have so far, to create a draft doc design. Comments welcome:
https://docs.google.com/document/d/1HSkYmh1ALDB9XhCffK70jjzREnXhjdbM0X8pd6arXnA/edit?usp=sharing

If you'd like to propose a completely different doc design, feel free to add a new section called "Proposed doc design (draft # 2)" or whatever. :)

I'll share this doc via kubeflow-discuss too.

johnugeorge:
@sarahmaddox Thanks for creating the doc. It has given much more clarity now.  The proposal looks ok to me.

jlewi:
@sarahmaddox @johnugeorge @vkoukis 

Thanks @sarahmaddox for putting the doc plan together; I don't have comment access so I'll comment here.

* I don't think "Kubeflow on Existing Kubernetes" is the right terminology

  * This will potentially confuse users who want to reuse an existing cluster but want provider specific instructions (e.g. instructions specific to reusing a GKE/Cisco/EKS/AKS/etc....) cluster

* What if we had two separate sections with names along the following

  * Community Maintained Solutions

       * Vendor neutral solutions governed by community consensus would go here

  * Provider Maintained Solutions

        * Any vendor/provider who wanted to provide an oppionated solution would go here


I'm not particularly fond of the names so I'm open to suggestions.

I think what we're seeing is that when it comes to deploying Kubeflow with vendor neutral, OSS tools  (e.g. Dex and ISTIO) the docs could potentially go in either.

I think as a community we want to have some community/maintained vendor/provider neutral solutions.  I think the default should be to put them in the community maintained bucket and to name them accordingly (i.e. vendor/provider neutral).

vkoukis:
> @sarahmaddox @johnugeorge @vkoukis
> 
> Thanks @sarahmaddox for putting the doc plan together; I don't have comment access so I'll comment here.
> 
>     * I don't think "Kubeflow on Existing Kubernetes" is the right terminology
>       
>       * This will potentially confuse users who want to reuse an existing cluster but want provider specific instructions (e.g. instructions specific to reusing a GKE/Cisco/EKS/AKS/etc....) cluster
>

This is already solved by the explicit reference to cloud-specific instructions in the page, and the cloud-specific setup they perform. I think this removes any potential for confusion. Here is the relevant segment:
> If you are using a Kubernetes distribution or Cloud Provider which has specific instructions for installing Kubeflow we recommend following those instructions. Those instructions do additional Cloud specific setup to create a really great Kubeflow experience.

>     * What if we had two separate sections with names along the following
>       
>       * Community Maintained Solutions
>         
>         * Vendor neutral solutions governed by community consensus would go here
>       * Provider Maintained Solutions
>         ```
>         * Any vendor/provider who wanted to provide an oppionated solution would go here
>         ```
> 

I'm not sure what problem having two separate sections tries to solve. We only have two options in the flat list now, and the structure proposed by Sarah clearly mentions who maintains each one. I propose we leave it as is.

> 
> I'm not particularly fond of the names so I'm open to suggestions.
> 
> I think what we're seeing is that when it comes to deploying Kubeflow with vendor neutral, OSS tools (e.g. Dex and ISTIO) the docs could potentially go in either.
> 
> I think as a community we want to have some community/maintained vendor/provider neutral solutions. I think the default should be to put them in the community maintained bucket and to name them accordingly (i.e. vendor/provider neutral).

johnugeorge:
I agree with Jeremy. As a community, the right step is to make a clear distinction between community maintained(vendor neutral) vs vendor maintained(opinionated) configs.  There will be more community maintained configs as well as vendor maintained configs in the future.  

I have added proposal doc #2.  Please have a look. There is more clarity than the earlier doc.

sarahmaddox:
Thanks @jlewi for the suggestions! (BTW I've given you access to the doc - not sure why it didn't give you access originally, as it's in the Kubeflow public shared drive.)

> What if we had two separate sections with names along the following
> Community Maintained Solutions
> Provider Maintained Solutions

I'd prefer not to have two separate sections. Splitting the info makes it harder for people to find and parse. 

> I think what we're seeing is that when it comes to deploying Kubeflow with vendor neutral, OSS tools (e.g. Dex and ISTIO) the docs could potentially go in either.
> I think as a community we want to have some community/maintained vendor/provider neutral solutions. I think the default should be to put them in the community maintained bucket and to name them accordingly (i.e. vendor/provider neutral).

It'd be great to have most of the solutions classified as community-maintained. If at all possible, a solution should be classified that way. But from the past discussions mentioned in this PR, it seems like we need the ability to offer provider-maintained solutions as well. Also, as we grow as a product and a community, is it likely (maybe even desirable) that we'll be able to offer more and more provider-maintained deployment solutions?

johnugeorge:
@sarahmaddox can you have a look at proposal design doc #2? It is an enhanced version of yours with more clarity IMO

sarahmaddox:
Thanks @johnugeorge! I'll take a look in the doc.

vkoukis:
@johnugeorge @sarahmaddox 

Johnu, thanks for updating the PR.
I see the following issues with the current state of the PR:

I understand Sarah's suggestion was to have a single page per installation option / config *under* a "Kubeflow on Existing Kubernetes Cluster" section in the docs.

So, the following should hold:
* All configs should be equivalent, each having their own page.
* All config pages should be named equivalently / at the same level.

The current approach keeps the name `getting-started-k8s.md` but instead of having this as the overview page, it converts it to a "Kubeflow Deployment with hkfctl_k8s_istio", which is config-specific. It keeps it as a link on the Getting Started page, and creates a new "Overview of Kubeflow Deployment on Existing Kubernetes Cluster" page, right next to it. This is very confusing.

I think we should just make the existing `getting-started-k8s.md` page *into* the overview page, and just name it "Kubeflow Deployment on Existing Kubernetes Cluster", for simplicity.

Then:
* Create a new page per config file and name it appropriately, i.e., something similar to `kfctl-k8s-istio.md` for the new config you are now documenting"
* Move the existing description of `existing_arrikto.yaml` into a `kfctl-existing-arrikto.md` page, so each config has its own, equivalent page
* Make sure the overview page *links* to all individual config pages, currently there are no links. I understand the phrase "Follow instructions: ..." should be a link to a config's individual page, in the overview page.

This ensures that no config has a direct link from the Getting Started page, they are all equivalently linked in the overview, and are easy to find.

The current PR does not even link to the Arrikto config, it seems to be totally unreachable from any page. Please ensure all configs have pages  that are named simarly/equivalently, and are all linked to properly.

Finally, I understand you insist in not following Sarah's suggestion here:
> I'd prefer not to have two separate sections. Splitting the info makes it harder for people to find and parse.

I still think it would be best to just maintain a flat list in the overview page, especially given Sarah's experience in technical documentation. And again, here is how K8s itself does it, a single flat table:
https://kubernetes.io/docs/setup/

Would it make sense for @sarahmaddox to do the initial PR, based on the current contents of her shared doc?

Once this PR is in, I will follow up with a PR improving the individual page for the config file maintained by us, and rename it based on the input from @sarahmaddox 's shared doc.

johnugeorge:
@vkoukis  The changes are not complete yet.  As it is difficult to verify links  from markdown page,  I am pushing changes one by one and verifying it in rendered website. I will post when I am ready with changes

vkoukis:
> @vkoukis The changes are not complete yet. As it is difficult to verify links from markdown page, I am pushing changes one by one and verifying it in rendered website. I will post when I am ready with changes

@johnugeorge , thank you for the update. I understood your forced update to the PR branch was the final proposed state of the docs. Please let us know when the branch is ready for review.

johnugeorge:
I kept the structure similar to proposal doc 2. 
Preview is ready at https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/

jlewi:
When I look at [Kubernetes](https://kubernetes.io/docs/setup/) Guide the hierarchy is

* Getting Started

  * Production Environment
    
    * Installing Kubernetes with Deployment Tools

      * Using kops

      * etc

    * Turnkey Cloud Solutions
        
       * AWS EC2

       * Google Cloud

      *  etc

    * On-Premises VMs
   
      * Cloud stack

So it doesn't seem like Kubernetes has a flat list in the overview page.

I don't know that we 3 need levels; but maybe we should have two?

Getting started

  * Deploying with open tools
     
     * ISTIO & Dex
      
  * Turnkey Solutions

     * AWS

     * Arrikto

I don't know exactly what to call the two nodes. But in the first section "deploying with open tools" is where I would expect to find vendor neutral solutions governed by the community. I wouldn't expect to see vendor/company names in the first section. 

@sarahmaddox I agree that having an extra level of navigation isn't great for users; but it is it worth it in order to try to have cleaner separation between vendor and community?

sarahmaddox:
Direct link to preview:
* https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/getting-started-k8s/

content/docs/k8s/_index.md
```diff
@@ -0,0 +1,4 @@
++++
```
sarahmaddox: Let's keep these docs in the "getting-started" section, since they're specifically about installing Kubeflow. (I'm hoping we won't get a set of docs per solution, but just one setup guide per solution. Let me know if that's not a realistic hope.)

To move the docs into the getting-started section:

* Move this doc from `content/docs/k8s/_index.md` to `content/docs/started/k8s/_index.md
* Give this doc a weight that puts it near the top of the "getting started" section. To do that, insert a new line between lines 3 and 4, with the following content: `weight = 1`. (We may need to juggle the weights of the other pages if this doesn't work as expected. Basically, the lowest weight goes at the top of the tree within a given section.)
* Move `content/docs/started/getting-started-k8s.md ` to `content/docs/started/k8s/overview.md`.
* Move the other 2 pages to `content/docs/started/k8s/kfctl-existing-arrikto.md` and `content/docs/started/k8s/kfctl-k8s-istio.md`.

content/docs/started/getting-started-k8s.md
```diff
@@ -1,524 +1,30 @@
...
-* [Kubeflow for Existing Clusters by Arrikto](#Kubeflow-for-Existing-Clusters---by-Arrikto)
+### Community maintained
 
-## Kubeflow for Existing Clusters - by Arrikto
+This section includes vendor-neutral solutions governed by community consensus. Below deployment configs are maintained and supported by the community.
 
-This installation of Kubeflow is maintained by [Arrikto](https://www.arrikto.com/), is geared towards existing Kubernetes clusters and does not depend on any cloud-specific feature.
 
-In this reference architecture, we use [Dex](https://github.com/dexidp/dex) and [Istio](https://istio.io/) for vendor-neutral authentication.
+|  Deployment config|      Description      |
```
sarahmaddox: For code readability and better table rendering, use HTML instead of Markdown for tables.

Example:
```
<div class="table-responsive">
  <table class="table table-bordered">
    <thead class="thead-light">
      <tr>
        <th>My col header 1</th>
        <th>My col header 2</th>
      </tr>
    </thead>
    <tbody>
      <tr>
        <td>My content row 1 col 1</td>
        <td>My content row 1 col 2</td>
      </tr>
      <tr>
        <td>My content row 2 col 1</td>
        <td>My content row 2 col 2</td>
      </tr>
    </tbody>
  </table>
</div>
```

See [this page](https://www.kubeflow.org/docs/about/style-guide/#use-simple-present-tense) for an example of how the tables render on the page.

content/docs/started/getting-started-k8s.md
```diff
@@ -1,524 +1,30 @@
...
+|----------|:-------------:|
```
sarahmaddox: Same comment about HTML for tables.

content/docs/k8s/deploy/kfctl-existing-arrikto.md
```diff
@@ -0,0 +1,462 @@
++++
+title = "Kubeflow Deployment with kfctl_existing_arrikto"
+description = "Instructions for installing Kubeflow with kfctl_existing_arrikto config"
+weight = 4
++++
+
+This deployment uses [Dex](https://github.com/dexidp/dex) and [Istio](https://istio.io/) for vendor-neutral authentication.
+
+**Maintainer and supporter: Arrikto**
```
sarahmaddox: To make this line stand out a bit more, use the following markup:

```
{{% alert title="Solution supported by provider" color="info" %}}
This deployment solution is provided and supported by <a href="https://www.arrikto.com/">Arrikto</a>.
{{% /alert %}}
```

content/docs/k8s/deploy/kfctl-k8s-istio.md
```diff
@@ -0,0 +1,92 @@
++++
+title = "Kubeflow Deployment with kfctl_k8s_istio"
+description = "Instructions for installing Kubeflow on your existing Kubernetes cluster using kfctl_k8s_istio config"
+weight = 4
++++
+
+This config creates a vanilla deployment of Kubeflow with all its core components without any external dependencies. The deployment can be customized based on your environment needs.
+
+**Maintainer and supporter: Kubeflow community**
```
sarahmaddox: To make this line stand out a bit more, use the following markup:

```
{{% alert title="Solution supported by community" color="info" %}}
This deployment solution is provided and supported by the <a href="https://github.com/kubeflow">Kubeflow</a> community.
{{% /alert %}}
```

content/docs/started/getting-started-k8s.md
```diff
@@ -1,524 +1,30 @@
...
-- Kubernetes Cluster with LoadBalancer support.
+### Vendor maintained
 
-If you don't have a Kubernetes Cluster, you can create a compliant Kubernetes Engine (GKE) on Google Cloud Platform cluster with the following script:
+This section includes the deployment solutions that are supported by specific vendors/providers.
 
-<details>
+| Deployment config     | Description                                                                                                                  | Maintainer/Supporter |
+|-----------------------|------------------------------------------------------------------------------------------------------------------------------|:----------:|
+| existing_arrikto.yaml | This deployment uses Dex and Istio for vendor-neutral authentication. <br />Follow instructions: [Kubeflow Deployment with existing_arrikto](/docs/k8s/deploy/kfctl-existing-arrikto/) |   Arrikto  |
```
sarahmaddox: Hyperlink Arrikto (and you'll need to use HTML as it's within the table):

```
<a href="https://www.arrikto.com/">Arrikto</a>
```

sarahmaddox:
Thanks @jlewi for your comment!

> When I look at [Kubernetes](https://kubernetes.io/docs/setup/) Guide the hierarchy is
> 
> * Getting Started
>   
>   * Production Environment
>     
>     * Installing Kubernetes with Deployment Tools
>       
>       * Using kops
>       * etc
>     * Turnkey Cloud Solutions
>       
>       * AWS EC2
>       * Google Cloud
>       * etc
>     * On-Premises VMs
>       
>       * Cloud stack
> 
> So it doesn't seem like Kubernetes has a flat list in the overview page.
> 
> I don't know that we 3 need levels; but maybe we should have two?
> 
> Getting started
> 
> * Deploying with open tools
>   
>   * ISTIO & Dex
> * Turnkey Solutions
>   
>   * AWS
>   * Arrikto
> 
> I don't know exactly what to call the two nodes. But in the first section "deploying with open tools" is where I would expect to find vendor neutral solutions governed by the community. I wouldn't expect to see vendor/company names in the first section.
> 
> @sarahmaddox I agree that having an extra level of navigation isn't great for users; but it is it worth it in order to try to have cleaner separation between vendor and community?

Having seen the comments by you and @johnugeorge in the Google doc, I'm happy with the plan outlined in draft # 2 in that doc - that is, to have two separate sections in the doc: one for community-provided solutions, one for provider solutions.

But now, reading your comment, I'm wondering if you're suggesting something entirely different.

My thinking was to keep the cloud-provider-specific guides separate from the on-prem/local guides.
So, we'd have:

* Kubeflow on generic existing cluster (this is the section currently under discussion)

  * Community-maintained
  * Provider-maintained (Arrikto currently)

* Kubeflow on AWS
* Kubeflow on GCP
* Kubeflow on Azure
* And so on - other clouds.

Are you thinking rather that we should have something like this:

* Kubeflow with open tools

  * Minikube
  * ...
  * Kubeflow on existing K8s with Istio and Dex

* Turnkey solutions

  * Minikf
  * Kubeflow on existing K8s with Arrikto
  * AWS
  * GCP
  * Azure
  * Etc

The [Kubernetes page](https://kubernetes.io/docs/setup/) implies to me that the community-maintained solutions are for **learning** and the provider-maintained solutions are for the production environment. It's hard to find the community-maintained solutions in the production section. :)

sarahmaddox:
@jlewi Following on from my previous comment: A summary might be this: Currently, we're restructuring [the generic K8s getting-started page (becoming a section](https://www.kubeflow.org/docs/started/getting-started-k8s/). I'm wondering if your proposal is to restructure the [overall Kf getting started page and section](https://www.kubeflow.org/docs/started/getting-started/).

If that's so, how about we let this current PR go through, then tackle the bigger restructure as tracked by issue https://github.com/kubeflow/website/issues/889?

Comments from @johnugeorge and @vkoukis and all are welcome too!

jlewi:
@sarahmaddox I think you raise a good point to punt on larger refactoring to the future. I think your proposal and the current structure in this PR is good. 

There is an open issue kubeflow/kubeflow#3739 tracking future improvements to the existing_arrikto.yaml config. I added some of the questions that came up here in that PR.

The one suggestion I have for the current doc is currently the titles in the left hand navigation bar are 

"Kubeflow deployment with kfctl_existing_arrikto"
"Kubeflow deployment with kfctl_k8s_istio"

Should we change those to be more meaningful to users in terms of what they actually do?

sarahmaddox:
Hallo @jlewi 

Thanks! Answering this question from you...
> The one suggestion I have for the current doc is currently the titles in the left hand navigation bar are
> 
> "Kubeflow deployment with kfctl_existing_arrikto"
> "Kubeflow deployment with kfctl_k8s_istio"
> 
> Should we change those to be more meaningful to users in terms of what they actually do?

We thought about having a meaningful title, but it's really hard to put differentiating, meaningful info into the short text offered in a title. Instead, we thought it'd be best to provide a one-to-one mapping between the YAML files and the title, and have the description in the overview doc and at the top of each detail doc.

johnugeorge:
@sarahmaddox Addressed your comments. Please take a look

If we are planning for a major restructuring, we can merge this as you suggested and tackle the rest in separate PR.

content/docs/started/k8s/overview.md
```diff
@@ -0,0 +1,55 @@
++++
+title = "Kubeflow on Existing Kubernetes Clusters"
```
sarahmaddox: Change the title so that it doesn't repeat the section name. Suggestion:

"Deploying Kubeflow on Existing Clusters"

sarahmaddox:
Latest preview links:

* https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/k8s/overview/
* https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/k8s/overview/
* https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/k8s/kfctl-k8s-istio/
* https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/k8s/kfctl-existing-arrikto/

sarahmaddox:
This LGTM in terms of doc structure. I'll add a hold so that we can wait for approval from other reviewers before merging the PR.

/hold
/lgtm

@vkoukis Are you happy with the new structure of this section of the getting-started docs? It has two sections on the overview page, one for community-supported and one for provider-supported solutions:
https://deploy-preview-977--competent-brattain-de2d6d.netlify.com/docs/started/k8s/overview/

content/docs/started/k8s/overview.md
johnugeorge: Updated. Please have a look

sarahmaddox:
Thanks! LGTM, and awaiting comments from the last reviewers.

/lgtm

vkoukis:
@sarahmaddox
> @vkoukis Are you happy with the new structure of this section of the getting-started docs? It has two sections on the overview page, one for community-supported and one for provider-supported solutions:

The two currently available config files definitely represent different opinions within the community, but we have already established and @jlewi has been very vocal about deployment on existing Kubernetes clusters being opinionated by its very nature.

Given that the config file Arrikto proposes does not depend on vendor-specific code or tools, that this config file was developed via a rather contentious process inside the community itself, and that Arrikto has been driving support for existing clusters within the community also ironing out quite a few issues in the process, I think the distinction between "community maintained" and "vendor maintained" config files introduces a rather artificial divide.

Despite that, I wouldn't want to block this PR. Let's move forward.

We will follow up with a PR to clarify issues reported by users for for kfctl_existing_arrikto, and also propose a more descriptive/informative title.

/lgtm

jlewi:
@vkoukis If you'd like this to be treated as a vendor neutral solution and not be separated from the other configs; then can we just find a name that is vendor neutral? (We can use kubeflow/kubeflow#3739 to track).

My objection to the original name "kfctl_existing.yaml" was that it was just too generic since there would be more than one config file for existing K8s clusters. 

I think all Kubeflow Providers (not just Cloud Vendors) should be on equally footing. So In my view long term all Provider specific solutions should be separated out from vendor neutral OSS ones. That's why I think long-term we might want to adjust the doc structure to delineate between the two
https://github.com/kubeflow/website/pull/977#issuecomment-517886460

johnugeorge:
+1 to @jlewi's comment

jlewi:
Everyone has LGTM'd this PR.  So I'm going to approve it and remove the hold.

@vkoukis I'm don't want to ignore your concerns so lets iterate in follow on PRs. I'd like to get this merged because I've been pointing a number of users at these docs with questions about minikube and other issues related to installing Kubeflow on existing cluster.

We can use kubeflow/kubeflow#3739 to track how to put the Arrikto config on equal footing.

PR #1009, opened 2019-08-01
Add anchor links to markdown headings.

Bobgy:
Not meant to be perfect, but I think this is very helpful in sharing links to specific sections.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1009)

### DIFF
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -9,6 +9,8 @@
         var a = document.createElement('a');
         // set visibility: hidden, not display: none to avoid layout change
         a.style.visibility = 'hidden';
+        // [a11y] hide this from screen readers, etc..
+        a.setAttribute('aria-hidden', 'true');
         a.innerHTML = ' #'
         a.href = '#'+heading.id;
         heading.insertAdjacentElement('beforeend', a);
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -1,14 +1,23 @@
 <script>
-  // Add anchor links to headings in markdown.
-  const article = document.getElementsByTagName('main')[0];
-  const headings = article.querySelectorAll('h1, h2, h3, h4, h5, h6');
-  headings.forEach(function(heading){
-    if (heading.id){
-      var a = document.createElement('a');
-      a.innerHTML = heading.innerHTML;
-      a.href = '#'+heading.id;
-      heading.innerHTML = '';
-      heading.appendChild(a);
-    }
-  });
+  // avoid polluting global namespace
+  (function() {
+    // Append anchor links to headings in markdown.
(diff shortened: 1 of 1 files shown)


Bobgy:
Note, credits to https://stackoverflow.com/a/45461609

sarahmaddox:
Preview:
https://deploy-preview-1009--competent-brattain-de2d6d.netlify.com/docs/about/style-guide/

Compare with prod version:
https://www.kubeflow.org/docs/about/style-guide/

sarahmaddox:
Thanks for this idea @Bobgy!

I think it needs a little more work to improve the UX. 

* Comparing the [version with linked headings](https://deploy-preview-1009--competent-brattain-de2d6d.netlify.com/docs/about/style-guide/) with the [current version](https://www.kubeflow.org/docs/about/style-guide/): The styling of the headings when linked is not as aesthetically appealing. I think we'd need to adjust the styles so that the headings retain the black colour, to give a more restful reading experience.

* The on-hover experience is also not as restful as when there are no links on the headings. I've seen some sites adopt the pattern of showing a little link icon at the end of a heading on hover, and people can click that icon if they want a link. I think this may be a better UX - what do you think?

It's also true that people can grab the heading links from the table of contents on the right of the page. This may not be immediately obvious, but perhaps people soon learn? 

If you do design a more restful reading experience that provides the link on the headings, it may be worth contributing that to the [Docsy theme](https://github.com/google/docsy), as well as adding it to the Kubeflow website. (Our website uses Docsy.)

sarahmaddox:
A question: Does the script expose any risk of script injection (XSS)? For example, if someone puts a script into the heading text?

Bobgy:
@sarahmaddox Thanks for the feedback. I think the UX changes are relatively easy. I can do that too soon.

For usability, my reading experience of every other doc site is using the UX you mentioned. Also, it matches user attention too: when user wants to get link to a title, they get it right next to it. For existing sidebar, I didn't find it even after 2 weeks.

For XSS, I don't think it exposes any new risks. If someone can inject script to a heading, it would already work. My script keeps the same html content, so it will still be the same.

Bobgy:
I made required UX changes. Didn't find a url icon, so I just used # instead. What do you think?

sarahmaddox:
Nice! This looks really good.

For the icon, how about the [link icon from material design](https://material.io/resources/icons/?icon=insert_link&style=baseline).

In case it's useful, here's more about [using the material icons](https://google.github.io/material-design-icons/).

sarahmaddox:
Thanks, this is a thing of beauty. :)

Would you like to contribute it to the [Docsy theme](https://github.com/google/docsy) too? I've raised issue https://github.com/google/docsy/issues/118

/lgtm
/approve

PR #1031, opened 2019-08-06
fix mini problem in fairing doc

jinchihe:
There is no var named `path-to-cloned-fairing-repo`, and in fact, that's no need since git clone to current path, Just remove `path-to-cloned-fairing-rep` to avoid confusing user. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1031)


jinchihe:
/assign @sarahmaddox  
Thanks.

Issue #1070, opened 2019-08-13
pin config links to v0.6-branch instead of master


yanniszark:
Links to kfctl config files for v0.6 should point to v0.6-branch instead of master.
eg. 
`https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_aws.yaml` 
->
`https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_aws.yaml`

/assign @yanniszark

PR #1071, opened 2019-08-13
pin config links for v0.6 to v0.6-branch

yanniszark:
Fixes https://github.com/kubeflow/website/issues/1070

/hold
Until the existing_arrikto rename is cherry picked in v0.6-branch: https://github.com/kubeflow/kubeflow/pull/3884

/assign @sarahmaddox 
/cc @abhi-g @lluunn 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1071)

### DIFF
--- a/content/docs/aws/deploy/install-kubeflow.md
+++ b/content/docs/aws/deploy/install-kubeflow.md
@@ -40,10 +40,10 @@ In order to deploy Kubeflow on your existing Amazon EKS cluster, you need to pro
 
     # Download config file
     export CONFIG="/tmp/kfctl_aws.yaml"
-    wget https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_aws.yaml -O ${CONFIG}
+    wget https://raw.githubusercontent.com/kubeflow/kubeflow/{{% kf-latest-version %}}/bootstrap/config/kfctl_aws.yaml -O ${CONFIG}
     ```
 
-    * `kfctl_aws.yaml` is one of setup manifests, please check [kfctl_aws_cognito.yaml](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_aws_cognito.yaml) for the template to enable authentication.
+    * `kfctl_aws.yaml` is one of setup manifests, please check [kfctl_aws_cognito.yaml](https://github.com/kubeflow/kubeflow/blob/v0.6-branch/bootstrap/config/kfctl_aws_cognito.yaml) for the template to enable authentication.
 
 1. Customize your config file. Retrieve the Amazon EKS cluster name, AWS Region, and IAM role name for your worker nodes.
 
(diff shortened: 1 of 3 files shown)


content/docs/aws/deploy/existing-cluster.md
```diff
@@ -20,7 +20,7 @@ If you would like to deploy Kubeflow on existing Amazon EKS cluster, the only di
 
     # Download config files
     export CONFIG="/tmp/kfctl_aws.yaml"
-    wget https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_aws.yaml -O ${CONFIG}
+    wget https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_aws.yaml -O ${CONFIG}
```
abhi-g: instead of hardcoding the versions in the url, we may be able to use hugo shortcodes that Sarah has setup for exactly this reason..

abhi-g:
Thanks for updating this Yannis. Please see my comment about using a shortcode variable. That way with the branching of the site, it will auto update to the correct branch.

content/docs/aws/deploy/existing-cluster.md
yanniszark: That would be very helpful!
@sarahmaddox can you provide some detail on how to use them?

sarahmaddox: Hmm, I think we may need an additional shortcode. The "latest-version" shortcode that we use for the docs specifies 3 levels of version. It's currently set to `v0.6.1`:
https://github.com/kubeflow/website/blob/master/layouts/shortcodes/kf-latest-version.html

But I notice that in this PR you need only `v0.6` - is that right?

Here's the syntax to use a shortcode:

```
{{% kf-latest-version %}}
```

And some [examples of its use](https://github.com/kubeflow/website/search?q=kf-latest-version&unscoped_q=kf-latest-version).

yanniszark: Actually, I think we should use latest-version.
Having the full version allows for immutability of the config files, while v0.6-branch may change.
cc @jlewi

yanniszark:
Since we are pinning to a specific tag, there is no need to wait for kubeflow/kubeflow#3884
That change will be picked up in v0.6.2.
@sarahmaddox @cspavlou I am reverting the filename back to `kfctl_existing_arrikto.v0.6.yaml` because that's the name in v0.6.1.
When v0.6.2 is out, we should make it `kfctl_existing_arrikto.yaml` again.

sarahmaddox:
Preview:
* https://deploy-preview-1071--competent-brattain-de2d6d.netlify.com/docs/aws/deploy/install-kubeflow/
* https://deploy-preview-1071--competent-brattain-de2d6d.netlify.com/docs/started/k8s/kfctl-existing-arrikto/
* https://deploy-preview-1071--competent-brattain-de2d6d.netlify.com/docs/started/k8s/kfctl-k8s-istio/

sarahmaddox:
Thanks!

/hold cancel
/lgtm
/approve

PR #1090, opened 2019-08-21
Change sdk package path to latest

hongye-sun:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1090)

### DIFF
--- a/content/docs/pipelines/sdk/install-sdk.md
+++ b/content/docs/pipelines/sdk/install-sdk.md
@@ -73,7 +73,7 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/{{% pipelines-sdk-version %}}/kfp.tar.gz --upgrade
+pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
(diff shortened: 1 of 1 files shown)


content/docs/pipelines/sdk/install-sdk.md
```diff
@@ -73,7 +73,7 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/{{% pipelines-sdk-version %}}/kfp.tar.gz --upgrade
+pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
```
sarahmaddox: Instead of linking to `latest`, I think we should offer a known good version of the SDK, if possible. How often does the SDK version change?

We have this file, which contains the latest SDL version number as a "variable":
https://github.com/kubeflow/website/blob/master/layouts/shortcodes/pipelines-sdk-version.html

So, there are two options:
* Either you should update the version number in the above file, and revert the change on line 76 of this file.
* Or you should keep the change in this file, and remove the above file altogether.

sarahmaddox:
This PR is superceded by https://github.com/kubeflow/website/pull/1095

Thanks @hongye-sun and @Bobgy!

/close

PR #1106, opened 2019-08-28
Update the v0.6.2 gcp instructions.

jlewi:
With 0.6.2 we want to start using the --config flag.

We decided to pin to specific commits of the KFDef files. This
might be overkill because on the tip of the branch the config files
are named based on the version.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1106)


jlewi:
@gabrielwen lets not block on these changes.

kunmingg:
/hold
Can you point to config after https://github.com/kubeflow/kubeflow/pull/4031 merge?

jlewi:
/hold cancel
Pull in changes for kubeflow/kubeflow#4031

PR #1169, opened 2019-09-24
Update oauth-setup.md

thesandlord:
Make it clearer what the Client ID is

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1169)

### DIFF
--- a/content/docs/gke/deploy/oauth-setup.md
+++ b/content/docs/gke/deploy/oauth-setup.md
@@ -72,7 +72,7 @@ address to verify the user's identity.
     https://iap.googleapis.com/v1/oauth/clientIds/<CLIENT_ID>:handleRedirect
     ```
     * `<CLIENT_ID>` is the OAuth client ID, something like 
-      `XXX.apps.googleusercontent.com`. Do not include the angle brackets around
+      `XXX.apps.googleusercontent.com`. In this example, it would be `711890160434-uhj4e162q3r11n8nmtnjoasg66op5mvi.apps.googleusercontent.com` Do not include the angle brackets around
       the client ID.
     * Note that the URI is not dependent on the Kubeflow deployment or endpoint. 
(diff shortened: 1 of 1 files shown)


content/docs/gke/deploy/oauth-setup.md
```diff
@@ -72,8 +72,9 @@ address to verify the user's identity.
     https://iap.googleapis.com/v1/oauth/clientIds/<CLIENT_ID>:handleRedirect
     ```
     * `<CLIENT_ID>` is the OAuth client ID, something like 
-      `XXX.apps.googleusercontent.com`. Do not include the angle brackets around
-      the client ID.
+      `XXX.apps.googleusercontent.com`. In this example, it would be
```
sarahmaddox: It's unclear where the client ID in your example comes from, as it doesn't seem to match the screenshots. See the preview of the updated page here:
https://deploy-preview-1169--competent-brattain-de2d6d.netlify.com/docs/gke/deploy/oauth-setup/

Also, if someone changes a screenshot they'll probably forget to change the text. 

Instead of giving a full example client ID, how about explaining where the client ID comes from:

```
`<CLIENT_ID>` is the OAuth client ID that you copied from the dialog box in an earlier step. The client ID looks something like this:
      `XXX.apps.googleusercontent.com`, where `XXX` represents a long string of characters. Do not include the angle brackets around the client ID.
```

thesandlord: Good point on the screenshot changing! I like the explanation of where the Client ID comes from, lgtm

sarahmaddox: Great! Can you update the PR, or would you like me to do that in a subsequent PR?

sarahmaddox:
Thanks!

/lgtm
/approve

Issue #793, opened 2019-06-11
Upgrade site to latest version of Hugo


kbhawkey:
Update the hugo version in netlify.toml to v0.53 or later.
Seems as if Netlify now supports Hugo 0.53 extended on Ubuntu xenial 16.04.
https://gohugo.io/hosting-and-deployment/hosting-on-netlify/#configure-hugo-version-in-netlify

scottilee:
@sarahmaddox how often should Hugo be updated? As of today, the most recent version is `0.56.3`. Is it as simple as updating the two `HUGO_VERSION` lines in https://github.com/kubeflow/website/blob/master/netlify.toml?

If it should be updated frequently maybe we can get some automation, in the form of a bot, that continuously checks for Hugo releases and sends a pull request?

kbhawkey:
hi! I raised this issue to upgrade the production site to >= hugo v0.53 *extended*. I have not evaluated the latest version of hugo ext. with netlify. Would this version upgrade simplify the CSS generation (no git check in of generated files to source)?

sarahmaddox:
Hallo both! I think we need to update our site to the latest Netlify, to pick up support for Hugo extended.

These updates (Netlify, Hugo, and also Docsy) are on my radar, but we have some content updates that take higher priority at the moment. Any update of the site infrastructure needs careful testing of the whole site.

sarahmaddox:
I've added https://github.com/kubeflow/website/issues/1139 for the Netlify upgrade.

PR #1191, opened 2019-10-02
Upgrade Hugo

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/793

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1191)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,8 +3,8 @@
   command = "hugo"
 
 [context.deploy-preview.environment]
-  HUGO_VERSION = "0.55"
+  HUGO_VERSION = "0.55.6"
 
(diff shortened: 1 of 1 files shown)


sarahmaddox:
I've examined the Hugo release notes for versions 0.51 to 0.55. Nothing shouts out that there may be a problem with upgrading.

The [Docsy config](https://github.com/google/docsy/blob/master/netlify.toml) currently specifies Hugo 0.55.6, and the Docsy docs recommend 0.53 or later.

I've tested the change locally and in preview:
https://deploy-preview-1191--competent-brattain-de2d6d.netlify.com/

Issue #1221, opened 2019-10-09
update 'End-to-end Kubeflow on GCP' tutorial as necessary


amygdala:
https://www.kubeflow.org/docs/gke/gcp-e2e/
Looks like we'll probably need to update this -- as per Dan, it only works on KF 0.5, but 0.5 deployment is now unsupported via the C2D app.  Creating placeholder to investigate.
(The docs do say to use 0.6 -- it's not clear to me whether the example breaks with 0.6.2...?)

amygdala:
kind/bug

amygdala:
Just some notes on things that need to be addressed:

- existing web ui service spec uses ClusterIP and won't provide external IP
  - related: could it make sense to add routing so that they can go through IAP endpoint?
- this line showing edits in web ui deployment.yaml: `- image: gcr.io/<your-project>/<your-deployment-name>` should be `- image: gcr.io/<your-project>/<your-deployment-name>-web-ui`
- let's consider creating a timestamp-based subdir under the bucket for the training results so that they can run the example multiple times
- the Dockerfile for the web ui image needs to pin to some earlier version of TF -- currently it is breaking with the unpinned build.

(cc @sarahmaddox @DanSanche )

amygdala:
Even using the prebuilt web ui image I get the following error -- do you see this, Dan?
![image](https://user-images.githubusercontent.com/115093/66611477-91b78280-eb73-11e9-8854-8d376b8282a2.png)

DanSanche:
hmm no I didnt see that when I ran through the codelab! It seems like theres an issue with the service. I can run through the doc again if that would help

amygdala:
I wonder if it could be IAP-related (which the tutorial indicates to use but the codelab does not). I tried with a fresh (IAP-enabled) cluster and the prebuilt webui image and got the same result again.  Hmm. I'll try with a basic auth endpoint next.

amygdala:
I think I figured it out -- the tutorial sets the tf-serving service name via
`kustomize edit add configmap mnist-map-serving   --from-literal=name=mnist-gcs-dist`  — which sets the service name to `mnist-gcs-dist`, but the flask app assumes the service name will be `mnist-service`— which is what the codelab uses.  
So I don't see how the tutorial variant actually worked previously. 😬

amygdala:
put in this PR to pin the TF version used for the web UI image: https://github.com/kubeflow/examples/pull/658  (Tests are breaking for reasons unrelated to my change).

[...]

PR #1240, opened 2019-10-15
intro tutorial updates

amygdala:
tracking issue: https://github.com/kubeflow/website/issues/1221 .
Accompanying update to the example code: https://github.com/kubeflow/examples/pull/658 (pin the TF version in the web-ui dockerfile) .

My text editor trims EOL whitespace, so there are a bunch of diffs for those.  The changes of substance:

 - changed the tf-serving service from `mnist-gcs-dist` to `mnist-service` so that the web UI would work out of the box. (If there is a good reason for using the other name, we can have them edit the service in the web ui, but they'll see that scary error first)
- when they set up their bucket paths, add a timestamp-based subdir (so that they can run the example more than once)
- since the web-ui service is using `ClusterIP`, and not generating an external IP, changed the instructions to have them port-forward. (I left some of the old stuff in but commented out -- feel free to ask me to just delete)
- fixed some instructions on how to edit the gcr image in the web ui `deployment.yaml `file.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1240)

### DIFF
--- a/content/docs/gke/gcp-e2e.md
+++ b/content/docs/gke/gcp-e2e.md
@@ -44,17 +44,17 @@ The MNIST dataset contains a large number of images of hand-written digits in
 the range 0 to 9, as well as the labels identifying the digit in each image.
 
 After training, the model can classify incoming images into 10 categories
-(0 to 9) based on what it's learned about handwritten images. In other words, 
-you send an image to the model, and the model does its best to identify the 
+(0 to 9) based on what it's learned about handwritten images. In other words,
+you send an image to the model, and the model does its best to identify the
 digit shown in the image.
-<img src="/docs/images/gcp-e2e-ui-prediction.png" 
+<img src="/docs/images/gcp-e2e-ui-prediction.png"
     alt="Prediction UI"
     class="mt-3 mb-3 p-3 border border-info rounded">
 
 In the above screenshot, the image shows a hand-written **7**. This image was
-the input to the model. The table below the image shows a bar graph for each 
-classification label from 0 to 9, as output by the model. Each bar 
-represents the probability that the image matches the respective label. 
+the input to the model. The table below the image shows a bar graph for each
+classification label from 0 to 9, as output by the model. Each bar
+represents the probability that the image matches the respective label.
 Judging by this screenshot, the model seems pretty confident that this image
 is a 7.
 
@@ -77,7 +77,7 @@ Here's an overview of what you accomplish by following this guide:
 
   * Saving the trained model to [Cloud Storage][cloud-storage].
   * Using [TensorFlow Serving][tf-serving] to serve the model.
-  * Running a simple web app to send a prediction request to the model and 
+  * Running a simple web app to send a prediction request to the model and
     display the result.
 
 Let's get started!
@@ -90,7 +90,7 @@ To simplify this tutorial, you can use a set of prepared files that include
 a TensorFlow application for training your model, a web UI to send prediction
 requests and display the results, and the [Docker][docker] files to build
 runnable containers for the training and prediction applications.
-The project files are in the [Kubeflow examples 
+The project files are in the [Kubeflow examples
 repository](https://github.com/kubeflow/examples/tree/master/mnist)
 on GitHub.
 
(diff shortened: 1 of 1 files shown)


content/docs/gke/gcp-e2e.md
```diff
@@ -532,19 +532,20 @@ Next, upload the container image to Container Registry so that you can run it on
     kustomize edit add configmap mnist-map-training   --from-literal=learningRate=0.01
     ```
 
-1. Configure parameters and save the model to Cloud Storage:
+1. Configure parameters for where the training results and exported model will be saved in Cloud Storage.  We'll use a timestamp-based subdirectory so that if the tutorial is run more than once, the training will start fresh each time.
 
     ```
-    kustomize edit add configmap mnist-map-training   --from-literal=modelDir=gs://${BUCKET_NAME}/
-    kustomize edit add configmap mnist-map-training   --from-literal=exportDir=gs://${BUCKET_NAME}/export
+    export BUCKET_PATH=${BUCKET_NAME}/$(date +%s)
```
DanSanche: It might make more sense to use VERSION_TAG here instead of grabbing a new timestamp, so the folder name matches the container tag that generated the output

amygdala: That's a good idea.

content/docs/gke/gcp-e2e.md
```diff
@@ -532,19 +532,20 @@ Next, upload the container image to Container Registry so that you can run it on
     kustomize edit add configmap mnist-map-training   --from-literal=learningRate=0.01
     ```
 
-1. Configure parameters and save the model to Cloud Storage:
+1. Configure parameters for where the training results and exported model will be saved in Cloud Storage.  We'll use a timestamp-based subdirectory so that if the tutorial is run more than once, the training will start fresh each time.
```
sarahmaddox: Style hint: Avoid the use of "we" in tech docs, as it's not clear who the "we" refers to. Use "you" instead. Also avoid the use of future tense, as some non-English languages don't have future tense in the same way as English, and readers get confused and wonder when exactly this thing will happen. (Full details in the [style guide](https://www.kubeflow.org/docs/about/style-guide/#address-the-audience-directly).

content/docs/gke/gcp-e2e.md
```diff
@@ -736,26 +737,34 @@ The example comes with a simple web front end that can be used with your model.
...
 ### Access the web UI in your browser
 
-Follow these steps to access the web UI in your web browser. It may take a few 
+The web-ui service added is of type `ClusterIP`, meaning it can’t be accessed from outside the cluster. In order to load the web UI in your web browser, you have to set up a direct connection to the cluster:
+
+```sh
+kubectl port-forward svc/web-ui 8080:80
+```
+
+Then, visit `http://localhost:8080/` to access the web UI in your browser. (If you're working from the [Cloud Shell](https://cloud.google.com/shell/), you can instead click the ['web preview'](https://cloud.google.com/shell/docs/using-web-preview) button and then select "Preview on Port 8080" to connect.)
```
sarahmaddox: s/Then, visit/Visit

content/docs/gke/gcp-e2e.md
```diff
@@ -736,26 +737,34 @@ The example comes with a simple web front end that can be used with your model.
...
+
+<!-- Follow these steps to access the web UI in your web browser. It may take a few
```
sarahmaddox: As you suggested ;) I'd recommend removing this content rather than commenting it out. The commented text is visible in the source preview, since it's HTML, and it's not necessary here. We can retrieve it from version control if we need it.

QQ: I know we had problems with port-forwarding in the distant past. Does it work reliably now?

content/docs/gke/gcp-e2e.md
amygdala: done, though if it doesn't read well to you feel free to make additional suggestions :)

amygdala:
PTAL again.

content/docs/gke/gcp-e2e.md
amygdala: good question -- it works in this case b/c we're connecting directly to the 'web-ui' service, not going through the KF dashboard.  AFAIK there are still issues with port-forwarding to dashboard, especially w/ IAP endpoint.

sarahmaddox: Thanks!

sarahmaddox:
Thanks @amygdala This looks great.
/lgtm
/approve

Issue #1246, opened 2019-10-16
Kubeflow Deployment with kfctl_k8s_istio doesn't contain updated way to install Kubeflow on existing cluster


navado:
https://www.kubeflow.org/docs/started/k8s/kfctl-k8s-istio/

Te documentation is irrelevant for kfctl v0.6.2. 
`kfctl init` reports that it's deprecated and `kfctl apply` should be used.
However `kfctl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_k8s_istio.0.6.2.yaml` doesn't work with an error:
```
ubuntu@ip-172-31-45-150:~/kf-experiment$ kfctl apply -f https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_k8s_istio.0.6.2.yaml
ERRO[0000] couldn't create directory  Error mkdir : no such file or directory  filename="coordinator/coordinator.go:432"
Error: failed to build kfApp from URI https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_k8s_istio.0.6.2.yaml:  (kubeflow.error): Code 400 with message: Error creating KfApp from config file:  (kubeflow.error): Code 400 with message: Error creating KfApp from config file: mkdir : no such file or directory
Usage:
  kfctl apply [all(=default)|k8s|platform] [flags]

Flags:
  -f, --file string   Static config file to use. Can be either a local path or a URL.
                      For example:
                      --config=https://raw.githubusercontent.com/kubeflow/kubeflow/master/bootstrap/config/kfctl_platform_existing.yaml
                      --config=kfctl_platform_gcp.yaml
  -h, --help          help for apply
  -V, --verbose       verbose output default is false

failed to build kfApp from URI https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6-branch/bootstrap/config/kfctl_k8s_istio.0.6.2.yaml:  (kubeflow.error): Code 400 with message: Error creating KfApp from config file:  (kubeflow.error): Code 400 with message: Error creating KfApp from config file: mkdir : no such file or directory

```

The error message is messy too, but it's probably separate issue.

if downloading the YAML file from the link above and using it locally the following error generated:
```
ubuntu@ip-172-31-45-150:~/kf-experiment$ kfctl apply -V -f $(readlink -f ../kfctl_k8s_istio.0.6.2.yaml)
[...]

PR #1250, opened 2019-10-17
Pin kfctl release binary to v0.6.2

krishnadurai:
This is done because the config also points to the same version.

Relates to issue #1246

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1250)


sarahmaddox:
Makes sense, thanks @krishnadurai 
/lgtm
/approve
/unassign @johnugeorge

PR #1277, opened 2019-10-25
[Pipelines] Greatly simplified the "Create Reusable Components" tutorial

Ark-kun:
The recently added artifact passing feature makes data passing and component development much simpler.
Updating the tutorials.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1277)

### DIFF
--- a/content/docs/pipelines/reference/component-spec.md
+++ b/content/docs/pipelines/reference/component-spec.md
@@ -54,9 +54,10 @@ This section describes the
 * `description`: Description of the component.
 * `metadata`: Standard object's metadata:
 
-    * `annotations`: A string key-value map used to add information about the component. Current limitation: the key cannot contain more that one slash ("/").
-        The annotations get translated to Kubernetes annotations when the component task is executed. See more information in the
+    * `annotations`: A string key-value map used to add information about the component.
+        Currently, the annotations get translated to Kubernetes annotations when the component task is executed on Kubernetes. Current limitation: the key cannot contain more that one slash ("/"). See more information in the
         [Kubernetes user guide](http://kubernetes.io/docs/user-guide/annotations).
+    * `labels`: Deprecated. Use `annotations`.
 
 ### Interface
 
@@ -116,7 +117,7 @@ use the component inside a pipeline.
 
 ### Consuming input by value
 
-The `{inputValue: <input name>}` placeholder is replaced by the value of the input argument:
+The `{inputValue: <Input name>}` placeholder is replaced by the value of the input argument:
 
 * In `component.yaml`:
   
@@ -139,7 +140,7 @@ The `{inputValue: <input name>}` placeholder is replaced by the value of the inp
 
 ### Consuming input by file
 
-The `{inputPath: <Input name>}` placeholder is replaced by the (auto-generated) local file path where the system has put the argument data passed for the "<input name>" input.
+The `{inputPath: <Input name>}` placeholder is replaced by the (auto-generated) local file path where the system has put the argument data passed for the "Input name" input.
 
 * In `component.yaml`:
 
--- a/content/docs/pipelines/reference/component-spec.md
+++ b/content/docs/pipelines/reference/component-spec.md
@@ -23,7 +23,7 @@ name: xgboost4j - Train classifier
 description: Trains a boosted tree ensemble classifier using xgboost4j
 
 inputs:
-- {name: Training data URI}
+- {name: Training data}
 - {name: Rounds, type: Integer, default: '30', help: Number of training rounds}
 
 outputs:
@@ -34,7 +34,7 @@ implementation:
     image: gcr.io/ml-pipeline/xgboost-classifier-train@sha256:b3a64d57
     command: [
       /ml/train.py,
-      --train-uri, {inputValue: Training data URI},
+      --train-set, {inputPath: Training data},
       --rounds,    {inputValue: Rounds},
       --out-model, {outputPath: Trained model},
     ]
@@ -54,16 +54,9 @@ This section describes the
 * `description`: Description of the component.
 * `metadata`: Standard object's metadata:
 
-    * `annotations`: An unstructured key value map
-        stored with a resource. External tools can set this property to store and
-        retrieve arbitrary metadata. Annotations are not queryable and should be
-        preserved when modifying objects. See more information in the
+    * `annotations`: A string key-value map used to add information about the component. Current limitation: the key cannot contain more that one slash ("/").
+        The annotations get translated to Kubernetes annotations when the component task is executed. See more information in the
         [Kubernetes user guide](http://kubernetes.io/docs/user-guide/annotations).
-
-    * `labels`: A map of string keys and values that can be used to
-        organize and categorize (scope and select) objects. May match selectors
-        of replication controllers and services. See more information in the
-        [Kubernetes user guide](http://kubernetes.io/docs/user-guide/labels).
 
 ### Interface
 
@@ -100,11 +93,11 @@ This section describes the
         * `command`: Entrypoint array. The Docker image's
             ENTRYPOINT is used if this is not provided. Each item is either a
             string or a placeholder. The most common placeholders are
-            `{inputValue: Input name}` and `{outputPath: Output name}`.
+            `{inputValue: Input name}`, `{inputPath: Input name}` and `{outputPath: Output name}`.
         * `args`: Arguments to the entrypoint. The Docker
             image's CMD is used if this is not provided. Each item is either a
             string or a placeholder. The most common placeholders are
-            `{inputValue: Input name}` and `{outputPath: Output name}`.
+            `{inputValue: Input name}`, `{inputPath: Input name}` and `{outputPath: Output name}`.
         * `env`: Map of environment variables to set in the container.
         * `fileOutputs`: Legacy property that is only needed in
             cases where the container always stores the output data in some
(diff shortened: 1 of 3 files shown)


sarahmaddox:
Preview:
* https://deploy-preview-1277--competent-brattain-de2d6d.netlify.com/docs/pipelines/reference/component-spec/
* https://deploy-preview-1277--competent-brattain-de2d6d.netlify.com/docs/pipelines/sdk/best-practices/
* https://deploy-preview-1277--competent-brattain-de2d6d.netlify.com/docs/pipelines/sdk/component-development/

content/docs/pipelines/sdk/component-development.md
```diff
@@ -1,151 +1,67 @@
...
-  Then **read the data** from that URI.  
-* Outputs: Upload the data to the URI provided as a command-line argument. Then 
-  write that URI to a local file, using a path provided as a command-line
-  argument.
+## Components and data passing
 
-### More about input data
+The concept of component is very similar to the concept of a function.
+Every component can have inputs and outputs (which must be declared in the component specification).
+The component code takes the data passed to it's inputs and produces the data for its outputs.
```
amygdala: it's --> its

content/docs/pipelines/sdk/component-development.md
```diff
@@ -386,20 +283,18 @@ Final version of `component.yaml`:
...
 implementation:
   container:
     image: gcr.io/my-org/my-image@sha256:a172..752f
     command: [
       python3, /pipelines/component/src/program.py,
-      --input1-path,       {inputValue: Input 1 URI},
-      --param1,            {inputValue: Parameter 1},
-      --output1-path,      {inputValue: Output 1 URI template},
-      --output1-path-file, {outputPath: Output 1 URI},
+      --input1-path,  {inputValue: Input 1},
```
amygdala: did you mean this to be of type `inputPath`?

content/docs/pipelines/sdk/component-development.md
```diff
@@ -416,6 +311,10 @@ import kfp
 dummy_op = kfp.components.load_component_from_file(os.path.join(component_root, 'component.yaml')) 
 # dummy_op = kfp.components.load_component_from_url('http://....../component.yaml')
 
+# Load two more components for importing and exporting the data:
+download_from_gcs_op = kfp.components.load_component_from_url('http://....../component.yaml')
+upload_to_gcs_op = kfp.components.load_component_from_url('http://....../component.yaml')
```
amygdala: was the intent to use these in the example pipeline?

content/docs/pipelines/sdk/component-development.md
```diff
@@ -1,151 +1,67 @@
...
-```
-
-Why should the program output the URI it has just received as an input argument?
-The reason is that the URIs specified in the pipeline are usually not the real
-URIs, but rather URI templates containing UIDs. The system resolves the URIs at
-runtime when the containerized program starts. Only the containerized program 
-sees the fully-resolved URI.
-
-Below is an example of such a URI:
+In some scenarios, the goal of the component is to create some data object in an external service (E.g. a BigQuery table).
```
sarahmaddox: s/E.g./for example,

Here and below.

Reason: Many people confuse the abbreviations "e.g." and "i.e.". Also, Latin abbreviations are hard for some people who have English as second language.

content/docs/pipelines/sdk/component-development.md
```diff
@@ -1,151 +1,67 @@
...
+In thais case the component program should do that and then output some identifier of the produced data location (e.g. BigQuery table name) instead of the data itself.
```
sarahmaddox: s/thais/this

content/docs/pipelines/sdk/component-development.md
```diff
@@ -1,151 +1,67 @@
...
+This scenario should be limited to cases where the data must be put into external system instead of keeping it inside the Pipelines system.
+The class of components where this behavior is common is exporters (e.g. "upload data to GCS").
+Note that the Pipelines system cannot provide consistency and reproducibility guarantees for the data outside of its control.
```
sarahmaddox: s/Pipelines/Kubeflow Pipelines

Or just "pipelines"

Here and below.

Reason: We should use either the brand name, capitalized, or a generic name not capitalized. This provides clarity in the docs.

content/docs/pipelines/sdk/component-development.md
```diff
@@ -1,151 +1,67 @@
...
-outputs to local files, then add a storage-specific wrapper that downloads and 
-uploads the data from/to a specific storage solution such as
-[Cloud Storage](https://cloud.google.com/storage/docs/) or 
-[Amazon S3](https://aws.amazon.com/s3/).
-For example, create a wrapper script that uses the [`gsutil
-cp`](https://cloud.google.com/storage/docs/gsutil/commands/cp) command to
-download the input data before running the main program and to upload the output
-data after the program finishes.
+There are two main ways a command-line program usually consumes data:
+* Small pieces of data are usually passed as command-line arguments: `program.py --param1 100`.
```
sarahmaddox: List rendering is broken here. To fix, add an empty line before the list.

See preview:https://deploy-preview-1277--competent-brattain-de2d6d.netlify.com/docs/pipelines/sdk/component-development/

content/docs/pipelines/sdk/component-development.md
```diff
@@ -268,7 +171,7 @@ chmod +x build_image.sh
 
 ## Writing your component definition file
 
-You need a component specification in YAML format that describes the
+To create a component from you containerized program you need to write component specification in YAML format that describes the
```
sarahmaddox: s/from you/from your

sarahmaddox: s/component specification/a component specification

content/docs/pipelines/sdk/component-development.md
```diff
@@ -314,12 +216,16 @@ three placeholders available:
 *   `{inputValue: Some input name}`   
     This placeholder is replaced by the **value** of the argument to the
     specified input. This is useful for small pieces of input data.
+*   `{inputPath: Some input name}`   
+    This placeholder is replaced by the auto-generated **local path** where the
+    system will put the input data passed to the component during the pipeline run.
+    This placeholder instructs the system to write the input argument data to a file and pass the path to that data file to the component program (insttead of passing the whole data).
```
sarahmaddox: s/insttead/instead

Also, instead of "passing the whole data", how about: `passing all the data as a placholder value`

content/docs/pipelines/reference/component-spec.md
```diff
@@ -54,17 +54,10 @@ This section describes the
...
 
-    * `annotations`: An unstructured key value map
-        stored with a resource. External tools can set this property to store and
-        retrieve arbitrary metadata. Annotations are not queryable and should be
-        preserved when modifying objects. See more information in the
+    * `annotations`: A string key-value map used to add information about the component. Current limitation: the key cannot contain more that one slash ("/").
+        The annotations get translated to Kubernetes annotations when the component task is executed. See more information in the
         [Kubernetes user guide](http://kubernetes.io/docs/user-guide/annotations).
 
-    * `labels`: A map of string keys and values that can be used to
```
sarahmaddox: Are labels no longer supported? I see they're still in the spec. As a reader, it's a bit confusing to see the `metadata` list item with only one sub-item (`annotations`). If for some reason we don't want to describe labels, we should either just mention them briefly as a second list item, or change the `annotations` entry from a list item to a continuation of the `metadata` list item.

content/docs/pipelines/reference/component-spec.md
```diff
@@ -121,51 +114,75 @@ use the component inside a pipeline.
...
   ```
 
-### Outputs
+### Consuming input by file
 
-Output paths are filled in by the pipeline system. The `outputPath` placeholder
-is replaced by a path. (The path can point to a mounted output volume, for
-example.) The parent directories of the path may or may not not exist. Your
-program must handle both cases without error.
+The `{inputPath: <Input name>}` placeholder is replaced by the (auto-generated) local file path where the system has put the argument data passed for the "<input name>" input.
```
sarahmaddox: s/"<input name>"/"&lt;input name&gt;"

Otherwise the page attempts to render the input field -- see preview: https://deploy-preview-1277--competent-brattain-de2d6d.netlify.com/docs/pipelines/reference/component-spec/#consuming-input-by-value

content/docs/pipelines/reference/component-spec.md
Ark-kun: `labels` are still supported, but I want to de-emphasize them and possibly deprecate in future.
When I was adding annotations and labels I did not consider them to be Kubernetes-specific, but now I see them as such (especially labels). The component spec should be portable, so Kubernetes-specific attributes are note desirable.
If we forget about the Kubernetes-related properties of those two attributes, then the become essentially the same.
So the idea is to at least deprecate the `labels` as too Kubernetes-specific in favor of `annotations`.
I can list `labels` as "deprecated".

>As a reader, it's a bit confusing to see the metadata list item with only one sub-item

The outline follows the structure of the component specification file:

```yaml
metadata:
    annotations:
        k1: v1
        k2: v2
    labels:
        k1: v1
        k2: v2
```

sarahmaddox:
Thanks @Ark-kun This LGTM. I'll approve and apply a hold in case @amygdala has any further comments.

/lgtm
/approve
/hold

content/docs/pipelines/sdk/component-development.md
```diff
@@ -329,47 +236,38 @@ command-line flags which are usually quite short.
...
+- {name: Output 1}
 implementation:
   container:
     image: gcr.io/my-org/my-image@sha256:a172..752f
     command: [
       python3, /pipelines/component/src/program.py,
+
       --input1-path,
-      {inputValue: Input 1 URI}, # Refers to the "Input 1 URI" input
+      {inputValue: Input 1}, # Refers to the "Input 1" input
```
amygdala: `inputPath` ?

amygdala:
/hold cancel
(Not sure that will work, Sarah might have to do..)
LGTM, though I think there is still one place where inputValue should be inputPath ? -- see my comment above.

content/docs/pipelines/sdk/component-development.md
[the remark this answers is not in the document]
Ark-kun: Done

content/docs/pipelines/sdk/component-development.md
Ark-kun: Great catch. The first one was also wrong.

content/docs/pipelines/sdk/component-development.md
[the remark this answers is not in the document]
Ark-kun: Done

content/docs/pipelines/sdk/component-development.md
Ark-kun: Done

content/docs/pipelines/sdk/component-development.md
Ark-kun: Done

content/docs/pipelines/reference/component-spec.md
Ark-kun: Fixed. Thanks for showing me the preview link.

content/docs/pipelines/reference/component-spec.md
Ark-kun: I've added `labels` back with a deprecation notice.

content/docs/pipelines/sdk/component-development.md
Ark-kun: I've reworded this phrase.

Ark-kun: But it might still be not great....

content/docs/pipelines/sdk/component-development.md
Ark-kun: Fixed

content/docs/pipelines/reference/component-spec.md
therc: Kubernetes adopters use labels extensively. They are not the same as annotations. If you merge them into annotations, you

a) make things confusing for users
b) will break Kubernetes tools that people like us have built, e.g. monitoring and accounting

Ark-kun:
> LGTM, though I think there is still one place where inputValue should be inputPath ? -- see my comment above.

Sorry, I missed that one. It's now merged in https://github.com/kubeflow/website/pull/1286

content/docs/pipelines/reference/component-spec.md
Ark-kun: > Kubernetes adopters use labels extensively. They are not the same as annotations.

I know about Kubernetes labels and annotations.
The annotations in Kubeflow Pipelines component specifications are not the same as Kubernetes annotations though. They were never intended to be the same as Kubernetes' annotations/labels. They were copied to Kubernetes pods at one point, but it was a mistake. Especially for labels (since they are very Kubernetes-specific). All platform-specific options should be specified on component instances, not on components themselves.

>will break Kubernetes tools that people like us have built, e.g. monitoring and accounting

Can you please tell me more about cases where you use Kubeflow Pipelines Components with Kubernetes tools.

content/docs/pipelines/sdk/component-development.md
Ark-kun: Fixed in  #1286

content/docs/pipelines/sdk/component-development.md
Ark-kun: Fixed

content/docs/pipelines/reference/component-spec.md
therc: We do resource accounting and troubleshooting based on Kubernetes labels. We use our own pipeline tools right now, but we are starting to test and will eventually adopt KF Pipelines. We need to find a way to propagate labels to the pods that run in the cluster. If you don't want to keep both, perhaps deprecate annotations instead? Labels are a more powerful construct, since that's what Kubernetes selectors allow you to use. Prometheus can tag data with either annotations or labels from pods, but e.g. Stackdriver Monitoring, DataDog and SignalFx only look at labels.

So right now the plan is for people to call `add_pod_label()` manually after instantiating a component? That means the burden is now shifted to all the callers, correct?

PR #1306, opened 2019-11-02
Update kfp-latest-version.html

IronPan:
/assign @Bobgy

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1306)


Bobgy:
@IronPan note that, the file shouldn't have an new line at the end.

Ark-kun:
> @IronPan note that, the file shouldn't have an new line at the end.

Maybe we can later lift this limitation?

Bobgy:
@Ark-kun it's more a hugo (the framework we used for doc website) limitation. This html file is replaced inline whenever there's a directive like `{{% kfp-latest-version %}}`, [link](https://github.com/kubeflow/website/blob/master/content/docs/pipelines/standalone-deployment-gcp.md#deploying-kubeflow-pipelines-standalone-to-an-existing-cluster)

Therefore, if we add a newline, there will be a newline after replace.

To be honest, this really sounds a little abusing hugo's templating feature. That's the existing solution in this repo. There may be a workaround.

@IronPan to edit the file so it doesn't have a new-line. You can use command line:
`echo -n "0.1.31" > kfp-latest-version.html`
it won't add the new line.

@sarahmaddox do you have any suggestions?

sarahmaddox:
Agreed that the shortcode file must not have a newline, due to the way we're using shortcodes. We could swap to params instead, but they're clumsier and we've settled on the convention of shortcodes.

@Bobgy Yes, please do send me a PR adding yourself to the OWNERS file for the website repo.

Issue #1313, opened 2019-11-04
Update profile docs to v1beta1 and 0.7


jlewi:
https://www.kubeflow.org/docs/other-guides/multi-user-overview/#manual-profile-creation

Its still referencing v1alpha1 but this should now be v1beta1.

/assign @kunmingg

Issue #1283, opened 2019-10-25
[GCP] Document IAM and multi-user


jlewi:
To share a cluster on GCP people need to be granted appropriate IAM roles.

We should document this.

To access the UI viap IAP they need http accessor role

```
gcloud projects add-iam-policy-binding ${PROJECT} --member=user:${EMAIL} --role=roles/iap.httpsResourceAccessor
```

To be able to do `gcloud get credentials` and see logs in stackdriver they will need viewer access on the project

```
gcloud projects add-iam-policy-binding ${PROJECT} --member=user:${EMAIL} --role=roles/viewer
```

jlewi:
@sarahmaddox yup I think that's a good place for the docs at least for now.
https://www.kubeflow.org/docs/other-guides/multi-user-overview/#pre-requisite-grant-user-minimal-kubernetes-cluster-view-access

sarahmaddox:
@jlewi I think PR https://github.com/kubeflow/website/pull/1318 addressed this issue. Can we close this issue now or is there still work to do?

jlewi:
PR #1318 updated the instructions about workload identity; this covers how people run workloads that authenticate to GCP.
https://www.kubeflow.org/docs/gke/authentication/

The instructions mentioned in the initial comment cover sharing access to the cluster. We have some instructions here
https://www.kubeflow.org/docs/other-guides/multi-user-overview/#pre-requisite-grant-user-minimal-kubernetes-cluster-view-access

but are missing the gcloud commands in the initial comment.

So I think before we close this issue we just need to add the gcloud commands to

https://www.kubeflow.org/docs/other-guides/multi-user-overview/#pre-requisite-grant-user-minimal-kubernetes-cluster-view-access

sarahmaddox:
/assign

Started work on this.

PR #1318, opened 2019-11-05
update profile doc API; add instrucitons for GCP access fine grain control

kunmingg:
fix https://github.com/kubeflow/website/issues/1313

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1318)

### DIFF
--- a/content/docs/gke/authentication.md
+++ b/content/docs/gke/authentication.md
@@ -53,15 +53,18 @@ gcloud --project=${PROJECT} iam service-accounts get-iam-policy ${KFNAME}-user@$
 
 When a pod uses KSA default-editor, it can access GCP APIs with the role granted to the GSA.
 
-**fine grain access control**:
-To control what GCP resources are accessible from a specific profile/namespace, you can explicitly specify which GCP service account to use when creating the profile.
+**Provisioning Custom Google Service Accounts In Namespaces**:
+Instead of using default GCP service account for every profile namespace, if you want to use custom GCP service account per namespace to control what GCP resources are accessible from a specific profile namespace, 
+you can explicitly specify which GCP service account to use when creating the profile.
 
-1. for user `user1@email.com`, if you don't already have a GCP service account you want to use, create a new one like: `user1-gcp@<project-id>.iam.gserviceaccount.com`: 
+Prerequisite: you need to have permission to edit GCP project IAM policy and create profile CR in kubeflow cluster.
+
+1. if you don't already have a GCP service account you want to use, create a new one like: `user1-gcp@<project-id>.iam.gserviceaccount.com`: 
 ```
 gcloud iam service-accounts create user1-gcp@<project-id>.iam.gserviceaccount.com
 ```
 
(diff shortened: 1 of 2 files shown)


sarahmaddox:
Also addresses issue https://github.com/kubeflow/website/issues/1283

content/docs/other-guides/multi-user-overview.md
```diff
@@ -151,6 +151,32 @@ The following resources are created as part of the profile creation:
 Namespaces, Profiles and Namespaces are sometimes used interchangably in the
 documentation.
 
+### Cloud API Authenticate with Profile
+
+**GCP**: User profile on GCP will be integrated with [GKE Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity).
+User pods will be authenticated automatically when talk to GCP APIs.
```
jlewi: "User pods will be authenticated automatically when talk to GCP APIs." -> 
"With workload identity, GCP credentials can be assigned to a pod by binding a Google Service Account to the Kubernetes Service Account used by the pod."

content/docs/other-guides/multi-user-overview.md
```diff
@@ -151,6 +151,32 @@ The following resources are created as part of the profile creation:
...
+
+By default all profiles in a cluster will share a public GCP service account identity.
```
jlewi: Could we clarify what is meant by default here?
I assume this is only true if user deployed with kfctl_gcp_iap.yaml or if their KFDef contains the GCPPluginSpec.

So we could say something like that e..g

"If you deployed Kubeflow following the GCP instructions then the profiler controller will automatically bind the "default-editor" service account to the GCP service account".

jlewi:
@kunmingg think we can another revision done today?

sarahmaddox:
Preview: https://deploy-preview-1318--competent-brattain-de2d6d.netlify.com/docs/gke/authentication/#starting-in-0-7-gke-workload-identity

/assign @jlewi 

@kunmingg thanks for these updates! @jlewi I'll leave the review to you, as you're doing such a great job of it already. :)

content/docs/gke/authentication.md
```diff
@@ -52,6 +53,52 @@ gcloud --project=${PROJECT} iam service-accounts get-iam-policy ${KFNAME}-user@$
...
+    name: user1@email.com   # replace with the email of the user
+  plugins:
+  - kind: WorkloadIdentity
+    spec:
+      gcpServiceAccount: user1-gcp@project-id.iam.gserviceaccount.com
+```
+
+**Note:**
+Profile controller currently doesn't perform any access control check to see whether the user creating the profile should be able to use the GCP service account. 
+So any user who can create a profile can get access to any service account for which the admin controller has owner permissions on. We will improve this in subsequent releases and link to an issue.
```
jlewi: Either delete "link to an issue" or replace it with an actual link to an issue

kunmingg: sorry, updated.

PR #1319, opened 2019-11-05
Document compatible Kubernetes versions

### LINKED ISSUE #1281: Document the compatible/incompatible/tested versions of various dependencies
Create a page with a compatibility matrix showing compatible/incompatible/tested versions of various dependencies, including:
* K8s, 
* ISTIO, 
* knative, 
* KFServing, 
* and maybe others....

Distinguish between:

* What we test/support
* What we know doesn't work
* What we haven't tested (maybe this can be specified implicitly by not listing it).

For example, for K8s:
* We know that Kubeflow doesn't work with Kubernetes 1.16.
* We haven't tested Kubeflow on Kubernetes 1.15 so we don't really know.
* We test Kubeflow on GKE with 1.14.

Note that we also include the K8s versions on this page:
https://www.kubeflow.org/docs/started/k8s/overview/#minimum-system-requirements

We should use shortcodes to contain the various version numbers, to free us up to use it in many places if necessary. (A shortcode also makes searching easy.) The shortcode should be similar to the one we use for K8s min version: https://github.com/kubeflow/website/blob/master/layouts/shortcodes/kubernetes-min-version.html

Location for doc - a page in this section:
https://www.kubeflow.org/docs/other-guides/

mameshini:
Document compatible Kubernetes versions 
fixes #1281

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1319)

### DIFF
--- a/content/docs/started/k8s/overview.md
+++ b/content/docs/started/k8s/overview.md
@@ -26,7 +26,7 @@ The Kubernetes cluster must meet the following minimum requirements:
     {{% kubernetes-min-version %}}. 
     * Kubeflow **does not work** on Kubernetes 
       {{% kubernetes-incompatible-versions %}}.
-    * Older versions of Kubernetes may not be compatible with the latest Kubernetes versions.  The following matrix
+    * Older versions of Kubernetes may not be compatible with the latest Kubeflow versions.  The following matrix
     provides information about compatibility between Kubeflow and Kubernetes versions.
 
     
--- a/content/docs/started/k8s/overview.md
+++ b/content/docs/started/k8s/overview.md
@@ -13,6 +13,11 @@ existing Kubernetes cluster is from one of those, consider following those instr
 
 The Kubernetes cluster must meet the following minimum requirements:
 
+  * Your cluster must include at least one worker node with a minimum of:
+    * 4 CPU
+    * 50 GB storage
+    * 12 GB memory
+
   * The recommended Kubernetes version is {{% kubernetes-tested-version %}}.
     Kubeflow has been validated and tested on Kubernetes
     {{% kubernetes-tested-version %}}.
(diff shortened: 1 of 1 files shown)


sarahmaddox:
Preview:
https://deploy-preview-1319--competent-brattain-de2d6d.netlify.com/docs/started/k8s/overview/

Thanks @mameshini - please would you add a comment when this PR is ready for review.

One thing that struck me: Do you think it's worth moving the version info and table into a separate page, in the [reference](https://www.kubeflow.org/docs/reference/) section of the docs? A couple of reasons:

* First, and most importantly, putting the info on a separate page makes it easier to search for (SEO) and to find by navigating through the doc menu. People won't necessarily look for this info in the page about installing on existing K8s clusters.
* Second, the table is likely to get quite large.

mameshini:
@sarahmaddox This PR is ready for review.
I updated it based on feedback I received from Jeremy.

We are providing this information to users who need to deploy on existing clusters, so it's certainly helpful to have the compatibility matrix in this section.  We can also copy it to the Reference section as well.  I will keep research about broader compatibility matrix, for example what versions of Istio, dex, etc are compatible with what Kubeflow versions.  But at this time I don't think I can accurately document it.

content/docs/started/k8s/overview.md
```diff
@@ -13,18 +13,81 @@ existing Kubernetes cluster is from one of those, consider following those instr
...
+        <th>Kubeflow 0.6</th>
+        <th>Kubeflow 0.7</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>1.11</td>
+        <td>compatible</td>
+        <td>compatible</td>
+        <td>incompatible</td>
```
sarahmaddox: It's a little hard to pick out the incompatible entries. For now, let's make them bold:
```
<td><b>incompatible</b></td>
```

It would be nice to have a more visual grid. For example a tick (check mark) for compatible, and a cross for incompatible. That's likely to be some work. We'd likely have to explore Bootstrap and make a shortcode (which we can then contribute back to the Docsy theme.) If you agree with the interim solution of making "incompatible" entries bold, then I'll raise an issue asking if anyone wants to prettify this grid (and others to come).

sarahmaddox:
Thanks @mameshini - just one comment then it'll be LGTM.

content/docs/started/k8s/overview.md
mameshini: I highlighted compatible versions to make the table easier to parse/make it easier find compatible versions.  For checkbox/cross symbols - it may look nicer although in that case we also need to add a description that check means compatible and cross means incompatible.

sarahmaddox:
Thanks!

/lgtm
/approve

PR #1335, opened 2019-11-11
Add doc fixit hosts to OWNERS for WtD AU 2019

alecglassford:
This change adds @alecglassford  and @rionam to the website OWNERS file as approvers and reviewers, so that we can review + approve changes contributed during the [Write the Docs AU 2019 doc fixit](https://www.writethedocs.org/conf/australia/2019/doc-fixit-kubeflow/).

This is intended to be a temporary change, and I'll send a followup change removing the two of us after the event(~2019-11-15).

@sarahmaddox: The [contributors guide section on OWNERS](https://www.kubeflow.org/docs/about/contributing/#owners-a-name-owners-1-a) says, "All users are expected to be assignable. In GitHub terms, this means they are either collaborators of the repo, or members of the organization to which the repo belongs," so I think you will indeed need to add us as collaborators on the kubeflow/website repo. (I'm guessing that this is easier/more proper than adding as members of the Kubeflow org.) I found [an example of the k8s bot getting mad](https://github.com/kubeflow/website/pull/777#issuecomment-499310239) when I tried to /lgtm which supports this. (It looks like /approve maybe doesn't require collaborator status though?)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1335)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -25,6 +25,8 @@ approvers:
   - vicaire
   - joeliedtke
   - animeshsingh
+  - alecglassford # Temporary for WtD AU 2019 fixit
+  - rionam # Temporary for WtD AU 2019 fixit
 reviewers:
   - abhi-g
   - aronchick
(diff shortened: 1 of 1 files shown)


sarahmaddox:
I've created PR https://github.com/kubeflow/internal-acls/pull/187 to add alecglassford and rionam to the Kubeflow org

PR #1408, opened 2019-11-20
Add makefile.

irvifa:
This will make build and test locally easier.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1408)

Maintenance plan:

Update `HUGO_VERSION` in Makefile and build the Dockerimage again if we want to use different version of hugo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+# Credit to Julien Guyomard (https://github.com/jguyomard). This Dockerfile
+# is essentially based on his Dockerfile at
+# https://github.com/jguyomard/docker-hugo/blob/master/Dockerfile. The only significant
+# change is that the Hugo version is now an overridable argument rather than a fixed
+# environment variable.
+
+FROM alpine:latest
+
+MAINTAINER Luc Perkins <lperkins@linuxfoundation.org>
+
+RUN apk add --no-cache \
+    curl \
+    git \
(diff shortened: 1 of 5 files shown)


sarahmaddox:
@irvifa Please would you explain the reason for this PR in more detail. 

If you're proposing a new way of building the website locally (that is, different from what's in the README) please would you include some documentation in the PR. Prerequisites, and how to use the MAKE file.

It will also be good to include a maintenance plan - what needs to change when the website is updated to different versions.

Note that the process needs testing before we can merge this PR.

irvifa:
@sarahmaddox can you explain how the test works for the makefile? Thanks!

sarahmaddox:
@irvifa We'll need to manually test the process after you've documented it.

irvifa:
@sarahmaddox okay i’ll update the readme and how we maintain it. Thanks for the advice..

irvifa:
@sarahmaddox I added the how to's in the `README.md` and plan of its maintenance.

abhi-g:
I don't think we need to complicate the website preview any more than it needs to be. The current prices is fairly straightforward. Either run hugo locally to create a preview, or create a PR and the current repo setup automatically creates a preview for you.

If you'd still like your Docker setup, I suggest you commit to your own GitHub repository.

Issue #1268, opened 2019-10-23
Document the Configurations and Extra Resources fields in the Notebooks setup screen


sarahmaddox:
Document the **Configurations** and **Extra Resources** fields in the Notebooks setup screen. (We currently mention **Extra Resources** when discussing GPUs, but we should document other available options too.)

Doc to update:
https://www.kubeflow.org/docs/notebooks/setup/#create-a-jupyter-notebook-server-and-add-a-notebook

Screenshot:
![image](https://user-images.githubusercontent.com/6138251/67442074-385c4400-f5b4-11e9-9594-0015dde35dd0.png)

sarahmaddox:
Guide to updating the docs:
https://github.com/kubeflow/website/blob/master/README.md

shanidchamal:
CPU, RAM and GPU are the main resources for cluster. for gpu, nvidia is common and amd also i think. the plugin set for extending Kubernetes are listed [here](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/). based on that, if somebody uses those plugins, then they may add those resources as extra. Also, [enabling gpu and tpu](https://www.kubeflow.org/docs/pipelines/sdk/gcp/enable-gpu-and-tpu/) suggests ways for adding gpu and nodeSelector for selecting a distinct type of gpu for particular notebook. nodeSelector is important for on-prem users as well but this feature not implemented yet. same the case for adding tpu in gke. For configurations.. are they low-level things users should be aware of ?

sarahmaddox:
@lluunn Maybe you can help? We need to tell users (a) which options are supported, and (b) the format supported by the UI fields. If you have any useful info/links that you can add to this issue, then I can get started with a PR.

/assign

lluunn:
cc @kimwnasptd

lluunn:
For extra resource, the [value](https://github.com/kubeflow/kubeflow/blob/master/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py#L402) should be a json string value for container.resouces.limit: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container

For example: `{"memory": "128Mi", "cpu": "500m"}`

lluunn:
Configuration is a list of labels: https://github.com/kubeflow/kubeflow/blob/master/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py#L365

[...]

PR #1419, opened 2019-11-27
Added guide to additional configurations and resources for notebook servers

sarahmaddox:
Documents the **Configurations** and **Extra Resources** fields in the Notebooks setup screen

Fixes https://github.com/kubeflow/website/issues/1268

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1419)

### DIFF
--- a/content/docs/notebooks/setup.md
+++ b/content/docs/notebooks/setup.md
@@ -81,6 +81,7 @@ When Kubeflow is running, you can access the Kubeflow user interface (UI). If th
   be the same as the namespace that you selected in a previous step. This 
   ensures that the new notebook server is in a namespace that you can access.
 
+<a id="docker-image"></a>
 1. Select a Docker **image** for the baseline deployment of your notebook 
   server. You can choose from a range of *standard* images or specify a 
   *custom* image:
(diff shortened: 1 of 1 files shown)


sarahmaddox:
/assign @lluunn 

@lluunn Please would you review this PR to check I've understood the tech correctly, and that we're giving users the right amount of information so that they can fill in the fields on the UI if they need to.

sarahmaddox:
This PR is ready for review. Applying a hold to allow for multiple reviewers. CCing @alecglassford for tech writer review.

/hold
/cc @alecglassford

content/docs/notebooks/setup.md
```diff
@@ -126,25 +126,65 @@ When Kubeflow is running, you can access the Kubeflow user interface (UI). If th
...
+
+1. *(Optional)* Change the setting for **enable shared memory**. The default is
+  that shared memory is enabled. Some libraries like PyTorch use shared memory
+  for multiprocessing. Currently there is no implementation in Kubernetes to 
+  activate shared memory. As a workaround, Kubeflow creates ab empty directory 
+  at `/dev/shm`. 
+
+1. *(Optional)* Specify one or more **extra resources** as a JSON string. The
+   JSON string must specify the value for one or more of the
+   `spec.containers[].resources.limits` options described in the [Kubernetes
```
alecglassford: The linked page seems to describe three options for the type of resource: `cpu`, `memory`, and `hugepages-<size>`. But the example you have added below seems to reserve GPU resources (`nvidia.com/gpu`) which doesn't match any of these (as far as I can tell). So I'm a bit confused as to what I can use in this section.

Also, let's say I do want to allocate extra cpu or memory in the **Extra Resources** section.  It seems like I already allocated these resources in steps 8 and 9 of this procedure, so what is the point or difference of doing that here?

content/docs/notebooks/setup.md
```diff
@@ -126,25 +126,65 @@ When Kubeflow is running, you can access the Kubeflow user interface (UI). If th
...
+  Kubeflow matches the labels in the **configurations** field against
+  the properties specified in the PodDefault manifest. Kubeflow then injects 
+  these configurations into all the notebook Pods on this notebook server.
+  
+    For example, enter the label `addgcpsecret` in the **configurations** field
+    to match to a PodDefault manifest containing the following configuration:
+
+    ```
+    matchLabels:
+    addgcpsecret: "true"
```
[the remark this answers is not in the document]
sarahmaddox: Thanks @alecglassford! Good pickup on the indentation - I made a copy/paste error. I'll fix this in my next commit.

+1 to us needing more info about the PodDefault manifest. When I was putting this PR together, I also created issue https://github.com/kubeflow/website/issues/1423 saying we need to document the use of PodDefault. In this current PR, I'm hoping to give expert users enough of a pointer about what goes into the UI field. @lluunn I'd be glad of your input here.

content/docs/notebooks/setup.md
```diff
@@ -126,25 +126,65 @@ When Kubeflow is running, you can access the Kubeflow user interface (UI). If th
...
+    ```
+
+    For indepth information on PodDefault usage, see the [admission-webhook
+    README](https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/README.md).
+
+1. *(Optional)* Change the setting for **enable shared memory**. The default is
+  that shared memory is enabled. Some libraries like PyTorch use shared memory
+  for multiprocessing. Currently there is no implementation in Kubernetes to 
+  activate shared memory. As a workaround, Kubeflow creates ab empty directory
```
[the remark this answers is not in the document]
sarahmaddox: Thanks!

content/docs/notebooks/setup.md
sarahmaddox: @lluunn Do you have an answer to Alec's question about how the JSON string `{"nvidia.com/gpu": 2}` fits into the `spec.containers[].resources.limits options`?

lluunn: This should be for GPU as cpu/memory can be set in previous steps as pointed out.
We can link to this page: https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/
There is an example usage of `nvidia.com/gpu: 1`

content/docs/notebooks/setup.md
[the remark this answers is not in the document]
sarahmaddox: I'll go with what we have for now, as the config option is an advanced one. We'll add a link to the new doc when issue #1423 is addressed.

content/docs/notebooks/setup.md
sarahmaddox: Thanks both! I've clarified the section.

sarahmaddox:
Thanks @lluunn PTAL

PR #1420, opened 2019-11-28
Update upgrade doc for upgrading from v0.6.2

janeman98:
Currently, kfctl of v0.6.2 does not support -f flag, following exiting documentation in https://www.kubeflow.org/docs/upgrading/upgrade/ 

Step 2 : 
...
kfctl delete -V -f ${CONFIG_FILE}

will throw an error. 
The documentation should be modified to download the kfctl of v0.7.0 first.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1420)

### DIFF
--- a/content/docs/upgrading/upgrade.md
+++ b/content/docs/upgrading/upgrade.md
@@ -13,7 +13,7 @@ Nonetheless, here are some instructions for updating your deployment:
 1. Check your Kubeflow configuration directory (`${KF_DIR}`) into source control
   as a backup.
 
-1. Delete your existing Kubeflow cluster:
+1. Delete your existing Kubeflow cluster (if you are upgrading from v0.6.2, apply step# 3 and 4 below first to get the latest verison of kfctl (Kubeflow v0.7.0) that support -f flag):
 
   ```
   export CONFIG_FILE=<the path to your Kubeflow config file>
--- a/content/docs/upgrading/upgrade.md
+++ b/content/docs/upgrading/upgrade.md
@@ -13,16 +13,13 @@ Nonetheless, here are some instructions for updating your deployment:
 1. Check your Kubeflow configuration directory (`${KF_DIR}`) into source control
   as a backup.
 
-1. Delete your existing Kubeflow cluster (if you are upgrading from v0.6.2, apply step# 3 and 4 below first to get the latest verison of kfctl (Kubeflow v0.7.0) that support -f flag):
+1. Delete your existing Kubeflow cluster:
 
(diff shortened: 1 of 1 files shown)


jinchihe:
Hey @janeman98 why not just remove `-f` option in the step 2? if user uses old kfctl client, should also can remove the kubeflow cluster. Thanks.

janeman98:
@jinchihe Thanks for the review!

From upgrading from v0.6.2: remove -f option in the step 2 should work
However, for upgrading from v0.7.0 to future version (say v1.0) : remove -f option in the step 2 will NOT work (since -f is required -- see below)

(v.0.7.0 kfctl)
kfctl delete -V
Error: Must pass in -f configFile
Usage:
  kfctl delete [flags]

Flags:
      --delete_storage   Set if you want to delete app's storage cluster used for mlpipeline.
  -f, --file string      The local config file of KfDef.
  -h, --help             help for delete
  -V, --verbose          verbose output default is false

Must pass in -f configFile

janeman98:
@jinchihe 
Since v1.0 release is not there yet, can we remove -f option in the step 2 for now? When we have v1.0, we can document how to upgrade from v0.6.2 to v1.0 and v0.7.0 to v1.0.

What do you think?

jinchihe:
> @jinchihe
> Since v1.0 release is not there yet, I believe we can remove -f option in the step 2 for now. When we have v1.0, we can document how to upgrade from v0.6.2 to v1.0 and v0.7.0 to v1.0.
> 
> What do you think?

I think the upgrading feature will be enhanced and the docs will be rewritten if 0.7 -> 1.0. Since this is hard to say this is upgrading... personally, we just need to consider old -> 0.7 case. Agree?

janeman98:
> > @jinchihe
> > Since v1.0 release is not there yet, I believe we can remove -f option in the step 2 for now. When we have v1.0, we can document how to upgrade from v0.6.2 to v1.0 and v0.7.0 to v1.0.
> > What do you think?
> 
> I think the upgrading feature will be enhanced and the docs will be rewritten if 0.7 -> 1.0. Since this is hard to say this is upgrading... personally, we just need to consider old -> 0.7 case. Agree?

Agree. I will remove -f option for now.

janeman98:
@jinchihe
I have updated the file as per your review feedback (+ move description on CONFIG_FILE to step#5). Please review/approve and let me know if further modification is needed.

jinchihe:
/lgtm
/approve

Thanks @janeman98

PR #1462, opened 2019-12-13
existing-arrikto: Update config to v0.7.1

yanniszark:
/cc @sarahmaddox 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1462)

### DIFF
--- a/layouts/shortcodes/config-uri-existing-arrikto.html
+++ b/layouts/shortcodes/config-uri-existing-arrikto.html
@@ -1,1 +1,1 @@
-https://raw.githubusercontent.com/kubeflow/manifests/dc04ff600cee722d93cf80d413aa73ddd8387f1f/kfdef/kfctl_existing_arrikto.0.7.0.yaml
+https://raw.githubusercontent.com/kubeflow/manifests/b37bad9eded2c47c54ce1150eb9e6edbfb47ceda/kfdef/kfctl_existing_arrikto.0.7.1.yaml


layouts/shortcodes/config-uri-existing-arrikto.html
```diff
@@ -1 +1 @@
-https://raw.githubusercontent.com/kubeflow/manifests/dc04ff600cee722d93cf80d413aa73ddd8387f1f/kfdef/kfctl_existing_arrikto.0.7.0.yaml
\ No newline at end of file
+https://raw.githubusercontent.com/kubeflow/manifests/b37bad9eded2c47c54ce1150eb9e6edbfb47ceda/kfdef/kfctl_existing_arrikto.0.7.1.yaml
```
sarahmaddox: Remove the empty line at the end of the file. (We use the shortcode as a global variable, and if you add a line break, that line break is included wherever the variable is used.)

sarahmaddox:
This LGTM, but we haven't yet released v0.7.1 - do you want to hold off on this change until the release happens?
https://github.com/kubeflow/kubeflow/releases

layouts/shortcodes/config-uri-existing-arrikto.html
yanniszark: @sarahmaddox it seems this is the result of using vim, thanks for catching that.

yanniszark:
@sarahmaddox thanks for the review!
It should be ready now.

sarahmaddox:
Thanks @yanniszark! I have a question about the fact that we haven't released a binary with a matching version number - please see the discussion on https://github.com/kubeflow/website/pull/1466

yanniszark:
@sarahmaddox that's a good question, now that kfctl lives in a different repo, it has its own distinct versioning.
In this PR, I'm more concerned about the fixes contained regarding the authentication part.
It solves this issue which is hitting a lot of users: https://github.com/arrikto/oidc-authservice/issues/2

So, it would be nice if we could get it in asap while we discuss the versioning story.

sarahmaddox:
> @sarahmaddox that's a good question, now that kfctl lives in a different repo, it has its own distinct versioning.
> In this PR, I'm more concerned about the fixes contained regarding the authentication part.
> It solves this issue which is hitting a lot of users: [arrikto/oidc-authservice#2](https://github.com/arrikto/oidc-authservice/issues/2)
> 
> So, it would be nice if we could get it in asap while we discuss the versioning story.

Sounds good!

/lgtm
/approve

PR #1466, opened 2019-12-16
Update config file versions to 0.7.1

lluunn:
/cc @jlewi @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1466)


lluunn:
> Thanks! This file needs updating too:
> https://github.com/kubeflow/website/blob/master/layouts/shortcodes/kf-latest-version.html

updated, thanks

sarahmaddox:
LGTM, except that the release isn't yet available:
https://github.com/kubeflow/kubeflow/releases

layouts/shortcodes/kf-latest-version.html
```diff
@@ -1 +1 @@
-v0.7.0
\ No newline at end of file
+v0.7.1
```
jlewi: Should we revert this change since we haven't release a v0.7.1 binary? We are only updating the configs?

Or should we release a v0.7.1 binary?

sarahmaddox: If at all possible we should release a v0.7.1 binary. It will be super confusing to have the configs with a different version number.

lluunn: But binary is the same I think. What we change is the manifest and kfdef.

jlewi: I don't think we want to keep kfctl and the manifests in lock stop. Downloading a new binary just to pick up config changes also doesn't seem desirable.

@sarahmaddox I agree with you that folks will get confused. Is there anything we can do to minimize confusion?

sarahmaddox: @jlewi When we do release a new version of the binary, will it be v0.7.1 or v0.7.2?

If the next binary will be v0.7.1, then we need to decouple the names of the configs from the version number of the binary entirely - so, have a different numbering scheme for the configs. This begs the question of whether there's a "Kubeflow" version number as a whole.

If the next binary will be v0.7.2 (so, we skip a version) then I think we can make it work by adding notes everywhere that there's no need for a new binary for Kubeflow v0.7.1. We need release notes for Kubeflow as a whole, rather than per component, to explain this.

jlewi: @sarahmaddox what would a different versioning schema look like? X.Y.Z is semantic versioning. All Kubeflow applications in Kubeflow follow semantic versioning. 

The fact that some applications happen to be in "sync" is more an accident of history and indicating that their release cycles happen to be in sync.

sarahmaddox: My assumption was that a config file v0.7.1 relates to kfctl v0.7.1. I wasn't aware that a config file is defined by an application version as opposed to a kfctl version. I saw configs not as applications but as things linked to a specific Kubeflow/kfctl version.

We'll need to change the docs, which currently refer to "Kubeflow v0.x.x" as if that's a thing, whereas that doesn't seem right? We'll also need to re-examine the versioning policy for Kubeflow v1.0 in this light.

For now, it's not good to hold up the release of the configs, so I'm happy if you want to LGTM and approve this PR.

sarahmaddox:
@lluunn Please would you change the PR description from this:
"update version from 0.7.0 to 0.7.1"

To this:
"Update config file versions to 0.7.1"

lluunn:
title updated

PR #1484, opened 2019-12-23
Add missing `gcloud auth login` step to "Deploy using CLI" doc

Bobgy:
/assign @sarahmaddox 

@rmgogogo found this step is also required for following actions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1484)

### DIFF
--- a/content/docs/gke/deploy/deploy-cli.md
+++ b/content/docs/gke/deploy/deploy-cli.md
@@ -51,7 +51,7 @@ some handy environment variables:
     tar -xvf kfctl_{{% kf-latest-version %}}_<platform>.tar.gz
     ```
 
-1. Login. You only need to run this command once:
+1. Log in. You only need to run this command once:
 
     ```
     gcloud auth login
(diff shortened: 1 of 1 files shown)


content/docs/gke/deploy/deploy-cli.md
```diff
@@ -51,6 +51,12 @@ some handy environment variables:
     tar -xvf kfctl_{{% kf-latest-version %}}_<platform>.tar.gz
     ```
 
+1. Login. You only need to run this command once:
```
sarahmaddox: s/Login/Log in

When using "log in" as a verb like the above, it's two words. When using "login" as a noun or adjective, it's one word. For example:
* "I want to log in."
* "I can't find the login button."

The pattern is the same for other words/phrases. Like:

* "I want to sign up but the signup process is ridiculously long."
* "I did back up my data but the backup failed."

I wrote a blog post about it once: https://ffeathers.wordpress.com/2018/05/20/login-or-log-in-sign-up-or-signup-how-to-tell-when-to-use-one-word-when-two/

:)

Bobgy: TIL, thanks for the detailed explanation

sarahmaddox: :)

PR #1518, opened 2020-01-11
Update run.md

SinaChavoshi:
currently in the API recurring run is referred to as a Job, updating the documentation to reflect this.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1518)

### DIFF
--- a/content/docs/pipelines/overview/concepts/run.md
+++ b/content/docs/pipelines/overview/concepts/run.md
@@ -1,5 +1,5 @@
 +++
-title = "Run and Recurring Run (also known as a Job) "
+title = "Run and Recurring Run"
 description = "Conceptual overview of runs in Kubeflow Pipelines"
 weight = 50
 +++
@@ -11,7 +11,7 @@ details page on the Kubeflow Pipelines UI, where you can see the runtime graph,
 output artifacts, and logs for each step in the run.
 
 <a id=recurring-run> 
-A *recurring run* (also refered to as a job in the API) is a repeatable run of
+A *recurring run* , or job in the Kubeflow Pipelines [backend APIs](https://github.com/kubeflow/pipelines/tree/06e4dc660498ce10793d566ca50b8d0425b39981/backend/api/go_http_client/job_client), is a repeatable run of
 a pipeline. The configuration for a recurring run includes a copy of a pipeline
 with all parameter values specified and a 
 [run trigger](/docs/pipelines/concepts/run-trigger/).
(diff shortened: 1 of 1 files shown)


joeliedtke:
This change can be previewed at:
https://deploy-preview-1518--competent-brattain-de2d6d.netlify.com/docs/pipelines/overview/concepts/run/

content/docs/pipelines/overview/concepts/run.md
```diff
@@ -11,9 +11,9 @@ details page on the Kubeflow Pipelines UI, where you can see the runtime graph,
 output artifacts, and logs for each step in the run.
 
 <a id=recurring-run> 
-A *recurring run* is a repeatable run of a pipeline. The
-configuration for a recurring run includes a copy of a pipeline with all
-parameter values specified and a 
+A *recurring run* (also refered to as a job in the API) is a repeatable run of
```
joeliedtke: I would recommend clarifying which API this refers to and adding a link. Also, I think it may be possible to shorten the sentence a little bit. How about something like this:

A *recurring run*, or [*job*](/docs/pipelines/reference/api/kubeflow-pipeline-api-spec/#tag-JobService) in the Kubeflow Pipelines API, is a repeatable run of a pipeline.

content/docs/pipelines/overview/concepts/run.md
```diff
@@ -1,5 +1,5 @@
 +++
-title = "Run and Recurring Run"
+title = "Run and Recurring Run (also known as a Job) "
```
joeliedtke: It would have been nice if KFP used the same term in the API and UI. I would recommend focusing on the UI terms for the title:

Run and Recurring Run 

Do you think that there are a significant set of users who primarily work with the API instead of the UI? If so, perhaps we should consider something like the following:

Run, and Recurring Run or Job

SinaChavoshi: That is a good point, I have only heard of one instance were this created confusion as the api name is hidden deep in the code and it is generally invisible to the users. However that being said there is no documentation that clarifies recurring run  is actually the same thing as the job API.  I'm ok with keeping the Run and Recurring Run title.

content/docs/pipelines/overview/concepts/run.md
SinaChavoshi: Awesome! love your recommendation, changing accordingly.

SinaChavoshi:
Thank you for your feedback Joe, updated the changes accordingly.

PR #1519, opened 2020-01-11
Update Microk8s install to specify K8s version

ofaz:
Current installation instructions will install (at the time of writing) k8s v1.17 - which is listed on https://www.kubeflow.org/docs/started/k8s/overview/ as being an incompatible version. Specifying the installation channel ensures the user installs a version that Kubeflow is verified against. 

These instructions will need to be updated as Kubeflow supports newer versions of K8s.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1519)

### DIFF
--- a/content/docs/started/workstation/getting-started-linux.md
+++ b/content/docs/started/workstation/getting-started-linux.md
@@ -19,7 +19,7 @@ GCP, AWS and Azure.
 [snapd](https://snapcraft.io/docs/getting-started), which is pre-installed on Ubuntu
 and Ubuntu derivative operating systems.
 
-1. Install MicroK8s - `snap install microk8s --classic`
+1. Install MicroK8s - `snap install microk8s --classic --channel=1.15/stable`
 2. Follow the getting started guide for Kubeflow on an
 [existing Kubernetes cluster](/docs/started/k8s/overview/).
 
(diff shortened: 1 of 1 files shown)


content/docs/started/workstation/getting-started-linux.md
```diff
@@ -19,7 +19,7 @@ GCP, AWS and Azure.
 [snapd](https://snapcraft.io/docs/getting-started), which is pre-installed on Ubuntu
 and Ubuntu derivative operating systems.
 
-1. Install MicroK8s - `snap install microk8s --classic`
+1. Install MicroK8s - `snap install microk8s --classic --channel=1.15/stable`
```
sarahmaddox: According to the [Kubeflow doc](https://www.kubeflow.org/docs/started/k8s/overview/), we should install K8s 1.14. Does channel 1.15 install K8s 1.14? That's confusing!

ofaz: @sarahmaddox Apologies - yes you're correct if we're wanting to ensure compatibility with Kubeflow 0.4 as well. I was focussing on ensure that the latest release (at present 0.7) is supported. Shall I amend?

ofaz: Based on the statement from the docs of "The recommended Kubernetes version is 1.14. Kubeflow has been validated and tested on Kubernetes 1.14." I have updated to specify 1.14 when installing Microk8s. Hopefully should prevent any installation woes!

sarahmaddox:
Thanks!

/lgtm
/approve

Issue #1263, opened 2019-10-23
Remove the 'version' param from config.toml if unused


sarahmaddox:
The `version` param is set in our `config.toml` but I don't think it's used any more.
To verify, search the repo for 'params "version"'.

Try removing the param and test the website.

sarahmaddox:
Also need to search for `.Params.description`

sarahmaddox:
I've searched the site for `params "version"` and `.Params.version`. Neither of those text strings currently appears anywhere.
However, I plan to use the `version` parameter in the fix for issue https://github.com/kubeflow/website/issues/1126, so I'll update the parameter description in `config.toml` as part of the fix for that issue.

sarahmaddox:
Fixed by PR https://github.com/kubeflow/website/pull/1535

/close

Issue #1126, opened 2019-09-05
In archived docs, add note that this is an unmaintained snapshot


sarahmaddox:
In the archived snapshot of docs for each major version (e.g. https://v0-5.kubeflow.org/, https://v0-4.kubeflow.org/): Add a note at the top of every page to let readers know that they’re seeing a snapshot rather than the latest docs:

> Kubeflow vx.x documentation is no longer actively maintained. The version you are currently viewing is a static snapshot. For up-to-date documentation, see the [latest version](https://www.kubeflow.org/docs/).

Investigate usage of a variable to make the above link point to the specific page on which the user clicked the link (if that page exists in the archived doc set) rather than to the top-level `/docs` page.

For this pattern in another doc set, see a [Kubernetes example](https://v1-14.docs.kubernetes.io/docs/tasks/tools/install-kubectl/).

_Note:_ This is an improvement that we'd like to see when we move to **Kubeflow v1.0** but we can implement this at any time. We don't need to wait for v1.0.

Source: _Kubeflow v1.0 Docs and Processes Design_ ([bit.ly/2OX11vw](https://bit.ly/2OX11vw))

sarahmaddox:
We should also add a note to the release processes, telling the release team how to add the note to the version that they're archiving.

sarahmaddox:
Starting work on this now.

sarahmaddox:
Just one more archived doc set to go: **v0.2**. I'll tackle that tomorrow.

sarahmaddox:
PR in progress to update the release procedures for website versioning: https://github.com/kubeflow/kubeflow/pull/4665
I'll test the procedure when creating the v0.7 archive branch of the docs, then submit the PR for review.

sarahmaddox:
This is now done.
/close

PR #1535, opened 2020-01-16
Added logic to create a banner on archived doc sets

sarahmaddox:
This PR adds a banner to the docs if the `archived_version` parameter is set to `true` in `config.toml`. We'll set the parameter to `true` for the archive branches of the docs, such as https://v0-6.kubeflow.org/ and https://v0-5.kubeflow.org/.

The banner doesn't show up in the preview, because the value of the parameter is `false`, which is the value we need in the master docs site. Here's the [preview to show no harm done if parameter is false](https://deploy-preview-1535--competent-brattain-de2d6d.netlify.com/docs/). :)

Here's a screenshot of the banner:

banner-for-doc-archives

For comparison, here's the [equivalent banner on the K8s docs](https://v1-16.docs.kubernetes.io/).

Implementation notes:

* The file `layouts/docs/baseof.html` is a copy of the theme's file at `themes/docsy/layouts/docs/baseof.html` with one additional line at line 21 to add the `version-banner.html` partial.
* The file `layouts/partials/version-banner.html ` is new and contains the text and logic for the banner.

* At first I tried URL pattern-matching as a way of discovering if a branch is showing the latest version or an archive version of the docs. But there's no way to get the host part of the URL from Hugo. (`.Permalink` shows a relative URL always, except when on `localhost`.) So I swapped to a true/false parameter instead.
* We could use JavaScript instead of Hugo templates to do the pattern matching and to inject the page banner. This is a little complex, as we'd have to put the URL pattern into a file on the website (as opposed to a Hugo variable). I've raised issue https://github.com/kubeflow/website/issues/1540 to discuss this option. It may be that a simple flag is more straightforward for Docsy users anyway - this is relevant because I'd like to contribute this change to Docsy once we're happy with it.
* Having a true/false parameter does mean that we have to adjust the value of the parameter in the version branch after creating the branch. Pattern matching would allow us to bypass that step. I'll create a PR adding this step to the versioning procedure.

Addresses issue https://github.com/kubeflow/website/issues/1126, but we'll need further PRs to update the archived docs in all the branches for the already-archived docs.
Fixes issue https://github.com/kubeflow/website/issues/1263

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1535)


sarahmaddox:
This PR is now ready for review. Applying a hold to allow for multiple reviewers.
/hold

/assign @abhi-g

sarahmaddox:
/assign @ccarpentiere 

Assigning @ccarpentiere for tech writing review.

abhi-g:
/lgtm

Waiting to let other reviewers take a look as well

layouts/partials/version-banner.html
```diff
@@ -1,13 +1,14 @@
...
+{{ if .Site.Params.archived_version }}
   {{ $color := "primary" }}
+  {{ $latest := .Site.Params.url_latest_version }}
   <div class="pageinfo pageinfo-{{ $color }}">
     {{ with .Site.Params.version }}<p>Version {{ . | markdownify }} of the
       documentation is no longer actively maintained. The site that you are
       currently viewing is an archived snapshot. For up-to-date documentation,
-      see the <a href="https://kubeflow.org/docs/">latest version</a>.</p>
+      see the 
+      <a href="{{ $latest | safeURL }}">latest version</a>.</p>
```
kbhawkey: hello @sarahmaddox , I read through the updates. I like the single parameter setting to trigger the include of the banner on the page. Does this partial place a banner on the archived pages or is there a different issue?

sarahmaddox: Hallo @kbhawkey Thanks for the review and feedback! This partial places the banner, but this PR applies only to the master branch, which affects only the most recent version of the docs.

As a next step, I'll need to apply the same changes to all the archive branches (`v0.6-branch`, `v0.5-branch`, and so on). I'll need to update the parameter on those branches, so that the branches display the banner. Those will be separate PRs. 

In addition, I'd like to get this PR merged into master before we create the `v0.7-branch` from master, so that I can test the whole versioning procedure with a branch that already contains the partial.

kbhawkey: :+1:

sarahmaddox:
Thanks for the tech writing review @aarondietz234!

This PR is now ready for merging.
/hold cancel
/approve

PR #1552, opened 2020-01-23
Update install-sdk.md

SinaChavoshi:
updating the kfp sdk installation to use pip3 , also adding instructions on how to deal with --user flag

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1552)

### DIFF
--- a/content/docs/pipelines/sdk/install-sdk.md
+++ b/content/docs/pipelines/sdk/install-sdk.md
@@ -73,7 +73,17 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
+pip3 install kfp --upgrade
+```
+*Note that if you are not using a virutal envirometn such as conda in step 1, you may recieve a error during installation as follows: 
+```bash
+ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/lib/python3.5/dist-packages/kfp-0.2.0.dist-info'
+Consider using the `--user` option or check the permissions.
+```
+you can bypass this erro by using --user which installs the binaries under ~/.local/bin note that in Ubuntu this is not part of $PATH. You can correct that by adding the following to the end of your.bashrc 
+
+```bash
+export $PATH=$PATH:~/.local/bin
 ```
 
 After successful installation, the command `dsl-compile` should be available.
--- a/content/docs/pipelines/sdk/install-sdk.md
+++ b/content/docs/pipelines/sdk/install-sdk.md
@@ -73,14 +73,19 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip3 install kfp --upgrade
+python3 -m pip install kfp --upgrade
 ```
-*Note that if you are not using a virutal envirometn such as conda in step 1, you may recieve a error during installation as follows: 
+**Note:** If you are not using a virtual environment, such as `conda`, when installing the Kubeflow Pipelines SDK, you may receive the following error:
 ```bash
 ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/lib/python3.5/dist-packages/kfp-0.2.0.dist-info'
 Consider using the `--user` option or check the permissions.
 ```
-you can bypass this erro by using --user which installs the binaries under ~/.local/bin note that in Ubuntu this is not part of $PATH. You can correct that by adding the following to the end of your.bashrc 
+
+If you get this error, install `kfp` with the `--user` option:
+```bash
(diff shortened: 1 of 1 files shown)


joeliedtke:
This change can be previewed at: 

https://deploy-preview-1552--competent-brattain-de2d6d.netlify.com/docs/pipelines/sdk/install-sdk/

content/docs/pipelines/sdk/install-sdk.md
```diff
@@ -73,7 +73,17 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
+pip3 install kfp --upgrade
+```
+*Note that if you are not using a virutal envirometn such as conda in step 1, you may recieve a error during installation as follows:
```
joeliedtke: Suggested revisions - spelling, formatting, etc: 

**Note:** If you are not using a virtual environment, such as `conda`, when installing the Kubeflow Pipelines SDK, you may receive the following error:

content/docs/pipelines/sdk/install-sdk.md
```diff
@@ -73,7 +73,17 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
...
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
+pip3 install kfp --upgrade
+```
+*Note that if you are not using a virutal envirometn such as conda in step 1, you may recieve a error during installation as follows: 
+```bash
+ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/lib/python3.5/dist-packages/kfp-0.2.0.dist-info'
+Consider using the `--user` option or check the permissions.
+```
+you can bypass this erro by using --user which installs the binaries under ~/.local/bin note that in Ubuntu this is not part of $PATH. You can correct that by adding the following to the end of your.bashrc
```
joeliedtke: Suggested revisions - spelling, clarity, etc:

If you get this error, install `kfp` with the `--user` option:

```
pip3 install kfp --upgrade --user
```

This command installs the `dsl-compile` and `kfp` binaries under `~/.local/bin`, which is not part of the PATH in some Linux distributions, such as Ubuntu. You can add `~/.local/bin` to your PATH by appending the following to a new line at the end of your `.bashrc` file:
 


 which installs the binaries under ~/.local/bin note that in Ubuntu this is not part of $PATH. You can correct that by adding the following to the end of your.bashrc

content/docs/pipelines/sdk/install-sdk.md
```diff
@@ -73,7 +73,17 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
+pip3 install kfp --upgrade
```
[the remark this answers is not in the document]
SinaChavoshi: That is a really good point... 
>pip alias needs to correspond to the python alias
this is correct, however if we use pip when users do not have a conda / virtual env it will fail as pip will point to python2 by default. 

Good news is that there is a work around that will bypass both issues 

python3 -m pip install --upgrade kfp

This will always install where your python points to. I'm upgrading the instructions accordingly. Also could you verify and let me know if this works for your env as well.

content/docs/pipelines/sdk/install-sdk.md
SinaChavoshi: done

content/docs/pipelines/sdk/install-sdk.md
SinaChavoshi: done

content/docs/pipelines/sdk/install-sdk.md
```diff
@@ -73,7 +73,22 @@ up Python using [Miniconda](https://conda.io/miniconda.html):
 Run the following command to install the Kubeflow Pipelines SDK:
 
 ```bash
-pip install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade
+python3 -m pip install kfp --upgrade
```
joeliedtke: I double checked this and I was incorrect, conda appears to map python and python3 to the same install of Python. So, pip3 install should be safe. Would you be willing to change it back?

pip install kfp --upgrade

Sorry for the confusion on my part.

joeliedtke: Oops, meant pip3...

pip3 install kfp --upgrade

SinaChavoshi: no worries at all. This is quite tricky, across the team we have had many issues with conda , virtual env python3 and pip3 :). 
I agree, python3 -m pip ...  is the safest path, however the advantage is not significant comparing to pip3. I like the pip3 more as it will be more familiar for the users. 
I'm updating to pip3 install kfp --upgrade as per your comment.

Thank you for all the work  and support :) 

Cheers, 
Sina

Issue #1421, opened 2019-11-28
Improve the Kubeflow getting-started experience [analytics-based]


sarahmaddox:
Action item from our [November doc analytics report](https://docs.google.com/presentation/d/1j_Eg3b3FWIb_NhLEimAQQx1URZq1VnqkLAWHS_TXBOo/edit#slide=id.g78f928ec47_1_116):  improve ratings for the [getting-started section](https://www.kubeflow.org/docs/started/) as a whole.

Recommended action: Work on a single, recommended flow for getting started quickly, focusing on a local installation for people who want to try Kubeflow on their own machines before committing to a cloud or on-prem installation.

sarahmaddox:
Related issue: https://github.com/kubeflow/website/issues/1097

sarahmaddox:
General guidance from the UX research team - what users want from the docs in general:
from the UX team: customer feedback of what they want from us:
* View from space.
* Give an opinion of the best path.
* Dive into the details of all options.

sarahmaddox:
Some related ideas here: https://github.com/kubeflow/website/issues/889

sarahmaddox:
More related ideas here: https://github.com/kubeflow/website/issues/877

sarahmaddox:
Working on this now.

PR #1566, opened 2020-01-28
Clarified the Kubeflow installation guide and added a diagram

sarahmaddox:
UX research and Google analytics show that people have trouble getting started with Kubeflow. We want to provide the smoothest getting-started experience possible.

* Added a diagram of installation options.
* Moved the local installation options down below cloud/K8s and added a note that local is not the best experience for new starters.
* Separated the cloud vs existing K8s options into two sections rather than one.

Preview: https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/started/getting-started/

Fixes https://github.com/kubeflow/website/issues/1421

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1566)


sarahmaddox:
/assign @jlewi 
/assign @abhi-g 
/assign @pdmack 
/assign @jbottum 

This PR is now ready for review. I'll apply a hold to allow for multiple reviewers.
/hold

sarahmaddox:
/assign @joeliedtke 

Assigning @joeliedtke for tech writer review.

content/docs/started/getting-started.md
```diff
@@ -16,34 +22,61 @@ It's important that you have some knowledge of the following systems and tools:
...
+Follow the
+[guide to deploying Kubeflow on Kubernetes](/docs/started/k8s/overview/).
+
+### Installing Kubeflow on a desktop or server
+
+{{% alert title="<b>Not</b> recommended for getting started with Kubeflow" 
+  color="info" %}}
+The procedures for installing Kubeflow on a desktop or server can be complex and
+highly dependent on your environment. The current recommendation is to choose
+a cloud installation if you're just getting started with Kubeflow.
```
joeliedtke: Nit: Remove "just"

The Google developer documentation style recommends against using  "just" since it can generally be removed from sentences without affecting the meaning.

joeliedtke:
/lgtm

Up to you if you want to make a change for the nit I added.

content/docs/started/getting-started.md
sarahmaddox: Thanks Joe! Style guides are worth following. :) However, in this case, I think the word "just" does add meaning. I tried removing it, but "if you're getting started with Kubeflow" carries different connotations from "if you're just getting started with Kubeflow". The first could mean you're well along the way, whereas the second means you're just starting out. Another option would be to reword the entire sentence, but I've decided not to go that far.

sarahmaddox:
Thanks @joeliedtke for the tech writing review.

jlewi:
Thanks @sarahmaddox 

@sarahmaddox and @abhi-g should we drop GCP with basic auth from the table?

I think we should strongly nudge users towards IAP because that's what we test and validate and will support long term.

I think we should continue to provide instructions to avoid "breaking" people but I'd like to strongly push people towards IAP.

See also kubeflow/website#1571

@sarahmaddox is this the diagram you were talking about that use to render links but the links no longer work? Do we have a follow on bug for that?

abhi-g:
I am okay with dropping off the Basic auth on GCP.

sarahmaddox:
> 
> @sarahmaddox and @abhi-g should we drop GCP with basic auth from the table?

OK, I've removed the GCP basic auth row from the table.

> 
> @sarahmaddox is this the diagram you were talking about that use to render links but the links no longer work? Do we have a follow on bug for that?

Yes, this is the diagram, I've tested the links again - something (Netlify or Hugo) removes the links before serving the image. I've raised issue https://github.com/kubeflow/website/issues/1573

sarahmaddox:
/hold cancel

This PR is now ready for LGTM and approval. Any comments welcome!

I'll take a look at the possibility of adding links to the SVG file as a separate update, tracked in issue https://github.com/kubeflow/website/issues/1573.

jbottum:
@sarahmaddox hey Sarah, we have some recommendations, especially for desktop with miniKF.    We believe that MiniKF has proven to be very simple (over 7k downloads) and increasing every month.   So we think that "Note this option maybe complex" should be removed.   As for "Installing Kubeflow on Desktop or Server", we think the "Not recommended ..." should be removed.

jlewi:
@jbottum Do you have a release candidate for minikf for 1.0?  I'd like to unblock this change and generally start getting the docs updated on master as part of producing an RC.

Until we have a 1.0 RC for minikf the debate about "recommended" vs. not seems mute. The minikf page will need some sort of note saying that minikf for 1.0 is not available yet.

So my suggestion is we let this PR go in as is. When an RC for minikf 1.0 is ready you can submit a PR to update the minikf docs and potentially include an update removing the "Not recommended"

jlewi:
Filed kubeflow/kubeflow#4714 to track a minikf RC.

sarahmaddox:
Thanks for all the feedback everyone. I'm travelling and haven't had a
chance to respond to review comments yet. I'm currently still in transit.
I'm happy to remove the note, both in the diagram and in the text, based on
feedback. I'll also await other feedback on the removal of the note.

cvenets:
@jlewi MiniKF on GCP is currently available with the latest official Kubeflow stable release (0.7). MiniKF on Vagrant is with 0.6, because we thought 1.0 would be out by now, so at first we were thinking about skipping 0.7 altogether. But, since 1.0 got delayed, we are planning to release the latest Kubeflow stable release (0.7) on MiniKF on Vagrant as well.

We can never align a MiniKF release with a Kubeflow release, because we first need to have the actual Kubeflow release out to test that everything works end-to-end, and then package it inside MiniKF. We are committed on trying to do this as fast as possible, after an official Kubeflow release is out, but this needs significant work to ensure the experience is smooth and everything works as it should.

From the way the docs are written I understand that a new user coming in will get a clear picture that they should *not* be trying to install on desktop or laptop because it is either "complex" or "not recommended". I think this is definitely not case. I think that the community has proven that MiniKF is one of the easiest ways, if not the easiest, to get started with Kubeflow. We are seeing more than 7K downloads on Vagrant and growing fast, plus significantly large enterprises using it on GCP.

So, I'm not sure this discussion should revolve around MiniKF release dates (since they can never align by design), but rather if we strategically want to put the laptop/desktop versions at the end of the list and essentially discourage people from using them.

I agree completely that on all MiniKF-related pages we should have a note that points to the exact version of Kubeflow this MiniKF version includes.

jlewi:
Thanks @cvenets  all of that seems good.

My proposal is that we should be working on getting the docs into the expected state at the time of KF 1.0 release.

So given that what information should be conveyed to users about the set of installation options for Kf 1.0 when KF 1.0 will be ready?

sarahmaddox:
Hallo all, I've removed the warning that the desktop/server installation is potentially complex.

jlewi:
Thanks @sarahmaddox

sarahmaddox:
/hold

I've added the hold back again, so folks feel comfortable about LGTM/approving this file while waiting for others to approve too.

@jbottum I've removed the notes as requested. Are you ready to LGTM/approve the guide? Also @jlewi and @abhi-g are you happy to LGTM/approve?

@pdmack If you have any feedback, that would be very welcome too. (I've pinged you on Slack about this doc too.)

jlewi:
/lgtm
/approve

@jbottum could you please take a look? It would be great to get this PR merged.

jbottum:
@sarahmaddox hey Sarah, we are reviewing and may have a suggestion as MiniKF works on Desktop and on GCP.    Currently the page just referencing minikf on GCP.   @cvenets @cspavlou please comment.

sarahmaddox:
Thanks @jbottum there's been a bit of discussion about the entry for MiniKF on GCP on this page. One thing we could do is list all the options for each OS. So, for example, instead of just listing Linux, MacOS, Windows, we could add a sublist for each one. The sublist would include MicroK8s, MiniKF, MiniKube (whichever apply). So, in effect, the sublist is a summary of each OS page. Expert users can use the sublist, others can use the main OS entry.

Something like this:

Choose the desktop/server guide for your operating system or environment:

  * To use Kubeflow on Linux, follow the
    [Linux deployment guide](/docs/started/workstation/getting-started-linux/). Options include:

    * [MicroK8s](https://microk8s.io/)
    * [Multipass and MicroK8s](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/other-guides/virtual-dev/getting-started-multipass/)
    * [MiniKF](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/other-guides/virtual-dev/getting-started-minikf/)
    * [Minikub](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/other-guides/virtual-dev/getting-started-minikube/)

(The above is a very rough mockup of the first bullet item. I'm at a conference, attending to reviews between sessions. Other OS bullet points would look similar.)

Then we can also include the MiniKF on GCP as already shown in the draft.

What do people think?

cvenets:
@sarahmaddox I think the laptop/desktop section can stay as is, and after one clicks on their preferred OS, they get correctly redirected to the corresponding page with all the different options. I don't think there needs to be any change there.

The only confusing part is that in the desktop/laptop section, there is a fourth bullet for MiniKF on GCP, which has nothing to do with desktop/laptop/server. I think this bullet should be removed from that section and incorporated in the first bullet of the "Installing Kubeflow on a public cloud" section, which refers to Google Cloud.

We can even explain that MiniKF is a getting-started tool, and only runs on one node, so for a multi-node Kubeflow deployment one should look at the "GCP deployment guide", so that it is 100% to the inexperienced user. Although I think that this is already clear, this is how the docs looked liked before as well.

What do you think?

P.S.: We are already working on producing MiniKF with KF 1.0 and after Jeremy's suggestion we will be adding a banner to the MiniKF page to note which version of Kubeflow it includes, so it is clear to users from there as well.

sarahmaddox:
@cvenets Thanks, done: I've moved the "MiniKF on GCP" list item back into the cloud section. There was some concern that MiniKF on GCP isn't a cloud solution, but is rather a server solution that's running on GCP. But I think there's enough justification both ways. Users will probably look in the cloud section for MiniKF on GCP.

cvenets:
Thanks @sarahmaddox. Yes, I think the users will first decide cloud vs. laptop/desktop/on-prem server, and then move on. So, it is much easier for users to navigate this way.

abhi-g:
@cvenets @jbottum 
I actually don't think MiniKF on GCP should be a recommended cloud deployment of Kubeflow. By definition MiniKF is a local deployment equivalent and therefore we'd be happy to put in that section as soon as you have tested out the 1.0 RC on the vagrant vm.

Please note that this PR is not going to be the final version of the doc. We can still update it, as soon as you have a tested version of MiniKF with 1.0.

cvenets:
@abhi-g I'm not sure I'm following your suggestion. MiniKF on GCP is not based on Vagrant at all, it is based on a GCP VM. We will be running separate tests of 1.0 RC for MiniKF on GCP, and separate tests for MiniKF on Vagrant, which is the laptop/desktop version.

I understand putting MiniKF on Vagrant (laptop/desktop) under the Desktop/server section, which I don't think is really necessary because the users will get re-directed to it either way, once they choose their OS, if that is what you are suggesting.

But how can MiniKF on GCP go under Desktop/server, when there is a section for "Installing Kubeflow on a public cloud" and MiniKF on GCP is tailored to run only on GCP? Shouldn't this be a second option for GCP? I also commented above that we may even want to be specific that this is a one-node Kubeflow deployment for experimentation, whereas someone wanting a full-blown, multi-node deployment should follow the "GCP Deployment Guide". Am I missing something?

jlewi:
All of the workstation/desktop solutions (MicroK8s, MiniKF, Minikube, Docker For Desktop etc...) could be run inside a VM on any cloud. That's probably not what I would recommend for anyone looking to run Kubernetes on Cloud (I would defer to the other cloud owners for their respective clouds).

I think the primary purpose of this page is to help users choose. So I don't think simply enumerating the different space of options is the goal. We should be pushing helping people identify the  recommended way to run things. On GCP the recommended way to run Kubernetes is GKE and the recommended way to run Kubeflow is on GKE.

Users do not need to use MiniKF to get a development, 1 node cluster they can just configure a GKE cluster to do that.

IIUC on GCP the primary reason to use MiniKF would be if someone wanted to use Rok with KF. IIUC people actually pay a fee (on top of the compute) for using MiniKF on GCP.

Since MiniKF isn't using GKE or even a multi-node capable cluster, users are giving up a lot (stackdriver, IAM integration, easy integration with GCR, etc...). What they get in return is Rok.

Rok is a great product but its a commercial, proprietary product that's not part of Kubeflow. I don't think driving people towards products integrated with Kubeflow is really in line with the goals of this page.

We should figure out a way to let people know about various (commerical and opensource) integrations with Kubeflow but I think that is out of scope for this PR and page.

sarahmaddox:
How about something like this:

**Cloud**
* GCP
* AWS
* Microsoft
* IBM

**Existing Kubernetes**
* Existing Kubernetes

**Lightweight Kubeflow**
* [MiniKF](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/other-guides/virtual-dev/getting-started-minikf/) for Linux, MacOS, and Windows.
* [MiniKF on GCP](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/started/workstation/minikf-gcp/)

**Installing Kubeflow on a desktop or server**
* Linux
* MacOS
* Windows

cvenets:
@jlewi I have to say that I'm a bit surprised, because it seems we are completely misaligned on the purpose of MiniKF.

MiniKF was never intended to be a Kubeflow deployment vehicle. Our goal, like yours, is the success of Kubeflow – to see Kubeflow adopted as the default platform of choice for ML deployments on Kubernetes.

We identified early on that it was too hard for users to instantiate a Kubeflow experimentation/testing environment in public cloud environments. MiniKF on GCP (or on local machines with Vagrant) allows users to come up to speed on Kubeflow much more rapidly than going straight to a public cloud environment. It is a stepping stone that will enable users to move more quickly to real, full deployments on GKE, and other supported public clouds.

MiniKF has nothing to do with promoting any proprietary technology. It is all about making Kubeflow successful.

@sarahmaddox thank you for the suggestion, it is definitely better than the current iteration. I will leave it to you to decide what's the best way going forward.

sarahmaddox:
@cvenets  and @jbottum  I've held discussions offline with @jlewi , @abhi-g , and @theadactyl . Please would you take a look at this updated PR and preview.

The preview is at the same URL as before: [https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/started/getting-started/](https://deploy-preview-1566--competent-brattain-de2d6d.netlify.com/docs/started/getting-started/)

In a nutshell, the change is:

+   Changed wording of the heading from "**Installing Kubeflow on a desktop or server**" to "**Installing Kubeflow on a single-node Kubernetes cluster**".
+   Added an explanation below the heading:  
"You can use the following options to run Kubeflow on a single-node Kubernetes cluster, which you can use on a desktop or server or in the cloud."
+   Moved MiniKF back into the section now titled "Installing Kubeflow on a single-node Kubernetes cluster".
+   Changed a few other bits of wording that refer to "desktop", to generalize the description to "a single-node cluster which you can use on a desktop or server or in the cloud". 
+   Updated the wording in the first horizontal row of the diagram to match the updated text.

We discussed using the word "local" instead of "single-node" in the heading ("Installing Kubeflow on a **local** Kubernetes cluster"), because "local" is technically more correct in the sense that the master and workers are all on the same machine. But we think that Kubeflow users reading this doc may understand "local" in the more common sense, meaning "on my local machine". We looked at the [Minikube docs](https://kubernetes.io/docs/tasks/tools/install-minikube/), which refer to a single-node Kubernetes cluster.

We think the new wording allows for including MiniKF in this section.

Please would you take a look and let us know your thoughts.

sarahmaddox:
@cvenets and @jbottum Do you have any thoughts about the latest version of the page? See my [earlier comment](https://github.com/kubeflow/website/pull/1566#issuecomment-583205514) for details.

cvenets:
@sarahmaddox IMHO we are making the life of inexperienced users harder. I think "Laptop/Desktop/Server" is by far the easiest way to guide people on what to do. I would actually add "Laptop" in, not remove "Desktop/Server" altogether.

Referring to a "single-node K8s cluster" makes the wording even more technical in my opinion for a getting started page, where a lot of user will have no experience with K8s at all. I don't want to make things harder for the users, so that we, internally, are 100% technically correct. This page should be much more marketing oriented, than technical. The only thing we care is to onboard more people faster, and let them play, until they decide where/how to do a full-blown deployment.

This is why I said previously, that I'll leave it up to you to decide where to put MiniKF. I don't think I have much to add.

sarahmaddox:
@cvenets thanks for your input. Let me think a little more on this. I'm currently busy with the doc sprint, so it may be a couple of days before I get back to you.

sarahmaddox:
@cvenets You're right about keeping the wording around desktop/server in the forefront. I've updated the heading for that section to the following:
**Installing Kubeflow on desktop, server, or cloud in a single-node Kubernetes cluster**

sarahmaddox:
/hold cancel

This PR is now ready for LGTM and merge.

PR #1572, opened 2020-01-29
Update getting-started-linux

ammarn911:
Redirecting to updated MicroK8s Kubeflow docs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1572)

### DIFF
--- a/content/docs/started/workstation/getting-started-linux.md
+++ b/content/docs/started/workstation/getting-started-linux.md
@@ -19,9 +19,8 @@ GCP, AWS and Azure.
 [snapd](https://snapcraft.io/docs/getting-started), which is pre-installed on Ubuntu
 and Ubuntu derivative operating systems.
 
-1. Install MicroK8s - `snap install microk8s --classic --channel=1.14/stable`
-2. Follow the getting started guide for Kubeflow on an
-[existing Kubernetes cluster](/docs/started/k8s/overview/).
+1. Install MicroK8s - `snap install microk8s --classic`
+2. Follow the getting started guide for Kubeflow [here](https://www.kubeflow.org/docs/other-guides/virtual-dev/getting-started-multipass/).
 
 ## Linux desktop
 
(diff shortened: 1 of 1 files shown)


ammarn911:
@sarahmaddox, updating this page.

content/docs/started/workstation/getting-started-linux.md
```diff
@@ -19,9 +19,8 @@ GCP, AWS and Azure.
 [snapd](https://snapcraft.io/docs/getting-started), which is pre-installed on Ubuntu
 and Ubuntu derivative operating systems.
 
-1. Install MicroK8s - `snap install microk8s --classic --channel=1.14/stable`
-2. Follow the getting started guide for Kubeflow on an
-[existing Kubernetes cluster](/docs/started/k8s/overview/).
+1. Install MicroK8s - `snap install microk8s --classic`
+2. Follow the getting started guide for Kubeflow [here](https://www.kubeflow.org/docs/other-guides/virtual-dev/getting-started-multipass/).
```
jlewi: Can we avoid using the URL in the link i.e. just

(/docs/other-guides/virtual-dev/getting-started-multipass/)

I think we want to use relative links otherwise it won't be consistent; i.e we have multple versions of the  docs. If we use relative links it will point to the correct version.

sarahmaddox: +1 to using relative links, for the reason given by @jlewi. I also think it'd be good to refer the reader directly to the page about Microk8s, so that they see all the setup in one place. How about this (no list, just straight text):

```
Follow the installation guide for [Kubeflow with MicroK8s](/docs/other-guides/virtual-dev/getting-started-multipass/) to set up MicroK8s and enable Kubeflow.
```

ammarn911: sounds neat @sarahmaddox , committed the change, thank you!

sarahmaddox:
/lgtm
/approve

Thanks!

Issue #1581, opened 2020-01-31
Publish v1-0 version of the docs


jlewi:
We'd like to do the following

1. Start publishing master on v1-0.kubeflow.org
1. Have www.kubeflow.org continue to direct to the v0.7 version of the docs
1. When the docs on master are ready redirect www.kubeflow.org to the v1 version of the docs.

I think doing the above should be fairly straightforward based on the [docs](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#creating-a-website-branch-for-the-latest-major-or-minor-release)

We basically want to follow the [process](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#creating-a-website-branch-for-the-latest-major-or-minor-release) for cutting release branches.

The only difference is in netlify. We should point master/latest at the version of the site corresponding to the v0.7 branch not v1.0.

jlewi:
In Netlify I used [deploy contexts](https://docs.netlify.com/site-deploys/overview/#deploy-contexts)
to switch the production branch from master to v0-7

I also enabled branch deploys for all branches

jlewi:
I believe the current [process](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#creating-a-website-branch-for-the-latest-major-or-minor-release) is creating a new netlify site for each branch. I'm not sure we want to do that. I think all of those cites are being owned by the individuals who created them; they don't show up in my netlify team when I log in.

jlewi:
I enabled deploy branches but they don't seem to be deployed; maybe we need to push a commit to trigger deploy?

At the very least the production version of the website www.kubeflow.org is now pointing to the v0.7-branch so we should be able to push changes to master.

jlewi:
When I pushed a commit to the v0.5-branch that triggered a deploy but it doesn't appear to be showing up at https://v0.5-branch--competent-brattain-de2d6d.netlify.com/

Could the period in the name be problematic?

jlewi:
Looks like the period is the problem. I pushed a commit to the branch will-may and that branch is now visible at
https://will-may--competent-brattain-de2d6d.netlify.com/

So I think next steps

* It looks like https://master--competent-brattain-de2d6d.netlify.com/ is already available
* I updated the CNAME record for v1-0.kubeflow.org so it should point to master
[...]

PR #1520, opened 2020-01-12
Added a support page for Kubeflow v1.0.0

sarahmaddox:
This PR adds a page describing **support options** for customers using Kubeflow v1.0 or later.

This PR is now ready for approval.

Preview: https://deploy-preview-1520--competent-brattain-de2d6d.netlify.com/docs/other-guides/support/

Fixes https://github.com/kubeflow/website/issues/1131

Reference: [Kubeflow v1.0 Docs and Processes Design](https://bit.ly/2OX11vw)

DO NOT MERGE UNTIL v1.0.0.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1520)

PR #1308, opened 2019-11-04
Kubeflow v1.0: Versioning policy and app stability indicators

sarahmaddox:
This PR adds a page describing the Kubeflow v1.0 versioning policy and adds status banners to all the applications included in the Kubeflow v1.0 release.

In summary, the PR implements the following:

* Vocabulary for Kubeflow application status: **Stable**; **beta**; **alpha**. See preview of the [versioning policy doc](https://deploy-preview-1308--competent-brattain-de2d6d.netlify.com/docs/reference/version-policy/).

* A matrix of Kubeflow apps and components for each Kubeflow version, starting from v1.0.0. See preview of the [versioning policy doc](https://deploy-preview-1308--competent-brattain-de2d6d.netlify.com/docs/reference/version-policy/#application-versioning-and-stable-status).

* A matrix for [SDKs and CLIs](https://deploy-preview-1308--competent-brattain-de2d6d.netlify.com/docs/reference/version-policy/#kubeflow-sdks-and-clis).

* A banner indicating the status of each application on the relevant pages. For example: 

    * [Notebooks](https://deploy-preview-1308--competent-brattain-de2d6d.netlify.com/docs/notebooks/why-use-jupyter-notebook/)
    * [TFJob](https://deploy-preview-1308--competent-brattain-de2d6d.netlify.com/docs/components/training/tftraining/)
    * And more

References:

* [Kubeflow v1.0 Docs and Processes Design](https://bit.ly/2OX11vw)
* [Kubeflow roadmap](https://github.com/kubeflow/kubeflow/blob/master/ROADMAP.md)

DO NOT MERGE UNTIL v1.0.0.

Fixes  https://github.com/kubeflow/website/issues/1236
Fixes https://github.com/kubeflow/website/issues/1129

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1308)

PR #1584, opened 2020-02-01
Revert "Added a support page for Kubeflow v1.0.0"

sarahmaddox:
Reverts kubeflow/website#1520

The *hold* on PR # 1520 was removed prematurely. The PR has dependencies on PR https://github.com/kubeflow/website/pull/1308, in terms of content and a link. We should submit the two PRs at the same time.

In addition, the website versioning is currently broken - the support page doesn't show up on the following site as predicted in a comment on the merged PR:
https://competent-brattain-de2d6d.netlify.com/docs/other-guides/
I can't open the page on either v1-0.kubeflow.org/ or master.kubeflow.org.
I'm concerned that if we revert to publishing master on www.kubeflow.org (as I think we should - see issue https://github.com/kubeflow/website/issues/1581) then this page will show up but without the versioning-policy page to back it up.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1584)


jlewi:
@sarahmaddox Apologies for the churn it took me a while to get netlify working.

The naming schema in Netlify is

https://${BRANCH}--competent-brattain-de2d6d.netlify.com/docs/other-guides/

So the support guide is available at
https://master--competent-brattain-de2d6d.netlify.com/docs/other-guides/support/

We also have a vanity URL
https://master.kubeflow.org/docs/other-guides/support/

As noted in #1586 there is an ugly SSL cert warning but you can work around it.

https://www.kubeflow.org  points to the stable version of the v0.7 docs on the v0.7 branch

I sent you #1585 to remove the **archived version banner**

So with #1586

* Users should see the stable version of the v0.7 docs until we are ready to flip to the v1 docs
* PRs can be merged on master without them being directly shown to users
* master can be viewed by developers and provides a preview

So I think at this point PRs for 1.0 can be unblocked and merged into master. I think this is preferable to putting holds on 1.0 PRs and then trying to get them all merged at once. 

Since users are redirected to the v0.7 branch of the docs, I think its ok if there is some inconsistencies in the master version of the docs while we work on getting them ready for 1.0.

Do we still need to revert this PR?

sarahmaddox:
Yes, we should still revert this PR. The docs on the website should retain their integrity as much as possible, even when on master.kubeflow.org. That is, if we know two PRs belong together, we should attempt to merge them as close to the same time as possible. The support page has a broken link and its content (app status as stable/alpha/beta) depends heavily on a page which does not yet exist.

sarahmaddox:
Thanks. I'll recreate the original PR tomorrow.

Issue #1554, opened 2020-01-24
End-to-end example needs to be changed to deploy tfjob into user profile namespace


amygdala:
https://www.kubeflow.org/docs/gke/gcp-e2e/
(I'll file a kubeflow/examples bug too).

Looks like this example is broken right now; looks like the tfjob needs to be deployed into a user profile namespace now. The quick workaround is to just direct the output of 'kustomize build' to a yaml file rather than piping directly to kubectl; then editing that file to change the namespace, then applying it.  The more principled fix would be to change the set of kustomize commands.

(for googlers, see [this discussion](https://groups.google.com/a/google.com/g/google-kubeflow-support/c/ZQixo8knLiA/m/QI9JijprEQAJ) for background)

amygdala:
Also, I assume that we need to update the tfjob docs as well (https://www.kubeflow.org/docs/components/training/tftraining/).  Looks like either they need to deploy into a `kubeflow-<user>` namespace, or the `default-editor` k8s service account needs to be added to the namespace they want to use. 
(Perhaps this should be a separate bug).

amygdala:
I'd put this as a 'p0' since it's the "getting started" tutorial.

jal06:
When I run the examples for Katib (doc https://www.kubeflow.org/docs/components/hyperparameter-tuning/hyperparameter/ ), they run in the namespace kubeflow. Looks like they also must also run in a user namespace. Do you agree ? Do we need to open another ticket ? (sorry for asking those questions, I am new to kubeflow)

jlewi:
@jal06 please open a separate issue for katib examples.

jlewi:
Bump to P0 because we should fix for 1.0.

sarahmaddox:
@amygdala Are you working on this tutorial and if so, can I assign this issue to you?

jal06:
> 
> 
> @jal06 please open a separate issue for katib examples.

Thanks for your answer
I open a new issue for Katib : [https://github.com/kubeflow/website/issues/1599](url)

sarahmaddox:
/cc @DanSanche 

@DanSanche I'm cc-ing you on this bug about the end-to-end tutorial (not the afore-mentioned Katib issue). @amygdala is planning to chat to you about it.

sarahmaddox:
Related issue in the kubeflow/samples repo: https://github.com/kubeflow/examples/issues/713

jlewi:
[...]

PR #1602, opened 2020-02-04
change e2e example to update to user profile namespace before tfjob deployment

amygdala:
This is necessary b/c the kubeflow namespace, by default, doesn't have sufficient permissions.
Addresses https://github.com/kubeflow/website/issues/1554.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1602)


amygdala:
/assign @sarahmaddox 
cc @DanSanche 
Here's what I'm thinking for a quick fix to the e2e example, just so it continues to work.  When there's time, we can figure out how to address it via kustomize config instead.  What do you think?

sarahmaddox:
Thanks @amygdala that looks good to me!

/lgtm
/approve

Issue #1316, opened 2019-11-04
Update samples page to show which version of Kubelow each sample works with


jlewi:
This page 
https://www.kubeflow.org/docs/examples/kubeflow-samples/

lists a lot of samples that I don't think are updated for 0.7.

* Should we remove samples until they have been updated for 0.7?
* Should we add the [xgboost_synthetic](https://github.com/kubeflow/examples/tree/master/xgboost_synthetic) ?

   * That is probably the sample most up to date for 0.7

* [code_search](https://github.com/kubeflow/examples/tree/master/code_search) is pretty stale at this point as are some other samples so we should probably remove that one.

* I think @amygdala might have made some updates to github issue summarization recently

jlewi:
@sarahmaddox suggested not removing the list of samples but instead with each one listing the Kubeflow versions they work with.

janeman98:
@sarahmaddox Is there an easy way to find out what version these sample work with?

I have a quick look on these samples, most of them just point to "Getting Started with Kubeflow"
Semantic code search -- > Getting Started with Kubeflow (https://www.kubeflow.org/docs/started/getting-started/)
Financial time series --> KUBEFLOW_VERSION=0.2.5 ??
GitHub issue summarization --> v0.4.0-rc.2.
MNIST image classification -- > Getting Started with Kubeflow
Object detection - cats and dogs -- > Getting Started with Kubeflow
PyTorch MNIST -->  Getting Started with Kubeflow
Ames housing value prediction --> Getting Started with Kubeflow

sarahmaddox:
@texasmichelle and @amygdala Can you help identify the Kubeflow version for each sample?

amygdala:
I should have some time to tackle it week after next.  (Most aren't "mine" and I don't know their current status). 
The github summarization/pipelines one _is_ mine, and has been updated to work with 0.7. (Note: that's the `pipelines` subdir of that example.)

hefedev:
I'd like to jump on this one if possible. 
To figure out which version works, I'll look at the CONFIG_FILE url and which master branch it points toward. This is always worked for me before.

sarahmaddox:
/assign @hefedev 

Thanks @hefedev!

sarahmaddox:
@hefedev I've moved this issue to **In progress** in the [doc sprint Kanban board](https://github.com/orgs/kubeflow/projects/21), because the issue is assigned to you. Happy sprinting!

hefedev:
[...]

PR #1620, opened 2020-02-08
Fixes kubeflow/website#1316

hefedev:
- The sample page now can read the date of the last commit for a reference project. 
- The Kubeflow versions are still not 100% correct but based on `app.yaml` `README` `Comment` a new label shows minimum tested version.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1620)

### DIFF
--- a/content/docs/examples/kubeflow-samples.md
+++ b/content/docs/examples/kubeflow-samples.md
@@ -4,50 +4,58 @@ description = "Examples that demonstrate machine learning with Kubeflow"
 weight = 10
 +++
 
-{{% blocks/content-item %}}
-This section introduces the examples in the
-[kubeflow/examples](https://github.com/kubeflow/examples) repo.
-{{% /blocks/content-item %}}
+This section introduces the examples in the [Kubeflows/examples](https://github.com/kubeflow/example) Rep.
 
-{{% blocks/content-item title="Semantic code search"
-  url="https://github.com/kubeflow/examples/tree/master/code_search" %}}
-Use a Sequence to Sequence natural language processing model to perform a semantic code search. This tutorial runs in a Jupyter notebook and uses Google Cloud Platform (GCP).
-{{% /blocks/content-item %}}
+{{% blocks/sample-section title="Financial time series"
+  kfctl="v0.7"
+  url="https://github.com/kubeflow/examples/tree/master/financial_time_series"
+  api="https://api.github.com/repos/kubeflow/examples/commits?path=financial_time_series&sha=master" %}}
+Train and serve a model for financial time series analysis using TensorFlow on GCP.
+{{% /blocks/sample-section %}}
 
-{{% blocks/content-item title="Financial time series"
-  url="https://github.com/kubeflow/examples/tree/master/financial_time_series" %}}
-Train and serve a model for financial time series analysis using TensorFlow
-on GCP.
-{{% /blocks/content-item %}}
+{{% blocks/sample-section title="GitHub issue summarization"
+  kfctl="v0.3.0-rc.3"
+  url="https://github.com/kubeflow/examples/tree/master/github_issue_summarization"
+  api="https://api.github.com/repos/kubeflow/examples/commits?path=github_issue_summarization&sha=master" %}}
+Infer summaries of GitHub issues from the descriptions, using a Sequence to Sequence natural language processing model.
+You can run the tutorial in a Jupyter notebook or using TFJob. You use Seldon Core to serve the model.
+Train and serve a model for financial time series analysis using TensorFlow on GCP.
+{{% /blocks/sample-section %}}
 
+{{% blocks/sample-section   title="MNIST image classification"
+  kfctl=""
+  url="https://github.com/kubeflow/examples/tree/master/mnist"
+  api="https://api.github.com/repos/kubeflow/examples/commits?path=mnist&sha=master" %}}
+Train and serve an image classification model using the MNIST dataset.
+You can choose to train the model locally, using GCP, or using Amazon S3. Serve the model using TensorFlow.
+{{% /blocks/sample-section %}}
 
-{{% blocks/content-item title="GitHub issue summarization"
-  url="https://github.com/kubeflow/examples/tree/master/github_issue_summarization" %}}
-Infer summaries of GitHub issues from the descriptions, using a Sequence to
-Sequence natural language processing model. You can run the tutorial in a
-Jupyter notebook or using TFJob. You use Seldon Core to serve the model.
-{{% /blocks/content-item %}}
+{{% blocks/sample-section   title="Object detection - cats and dogs"
+  kfctl=""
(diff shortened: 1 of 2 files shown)


sarahmaddox:
Preview: https://deploy-preview-1620--competent-brattain-de2d6d.netlify.com/docs/examples/kubeflow-samples/

content/docs/examples/kubeflow-samples.md
```diff
@@ -9,45 +9,37 @@ This section introduces the examples in the
 [kubeflow/examples](https://github.com/kubeflow/examples) repo.
```
sarahmaddox: For some reason this link is now broken.

layouts/shortcodes/blocks/samples-section.html
```diff
@@ -0,0 +1,11 @@
+{{ $url_text := .Get "url_text" }}
+{{ $api := .Get "api" }}
+{{ $last_updated := "" }} 
+{{ $version := .Get "kfctl" }} 
+<div class="col-lg-12 mb-5 mb-lg-0 ">
+  {{ with .Get "title" }}<h4 class="h3 mt-3">{{ . }}</h4>{{ end }}
+  {{ with getJSON $api }} {{ $last_updated = (index (index (index (index . 0) "commit") "committer") "date") | dateFormat "2006/01/02" }}
+  <p class="text-muted">{{ "Last update " }} {{ $last_updated }} {{ with $version }}{{ "Kubeflow " }}{{ . }}{{ end }}</p>{{ end }} 
+  <p class="mb-0">{{ .Inner }}</p>
+  {{ with .Get "url" }}<p><a href="{{ . }}">{{ with $url_text }}{{ $url_text }}{{ else }}{{ "Read" }}{{ end }}</a></p>{{ end }}
```
sarahmaddox: Would it be more meaningful to change `"Read"` to `"Go to sample"` now that we know this particular block is specifically for samples?

Another suggestion: I think it's unlikely we'll ever use the `url_text` param. Should we remove it, to make the code simpler?

config.toml
```diff
@@ -210,6 +210,54 @@ no = 'Sorry to hear that. Please <a href="https://github.com/kubeflow/website/is
 	icon = "fa fa-calendar"
         desc = "Discussion and help from your fellow users"
 
+[[params.examples.demo]]
```
sarahmaddox: I'm wondering why you've moved the content into the `config.toml` file. Overall, I'd prefer to keep the content in the page rather than in `config.toml`. I think it's simpler for contributors to find the content on the page rather than having to open the file corresponding to the page then find the `config.toml` file. Also, I'm concerned that the `config.toml` file will soon become large and difficult to navigate, when we start adding more samples.

README.md
```diff
@@ -183,6 +183,7 @@ You can override the default styles and add new ones:
   You can also use `_variables_project.scss` to specify your own values for any 
   of the default 
   [Bootstrap 4 variables](https://getbootstrap.com/docs/4.0/getting-started/theming/).
+* You can update the Kubeflow example projects variables in the [website configuration file](https://github.com/hefedev/website/blob/7406b6f0b248d4881537cdb957dc5b27ae71c3a5/config.toml).
```
sarahmaddox: This comment doesn't fit too well here, because this section of the README is about CSS styling. Also, the link points to the `hefedev` repo instead of the `kubeflow/website` repo.

See my comment below about keeping the content in the file corresponding to the page rather than in the config file.

content/docs/examples/kubeflow-samples.md
```diff
@@ -8,46 +8,4 @@ weight = 10
 This section introduces the examples in the
 [kubeflow/examples](https://github.com/kubeflow/examples) repo.
```
sarahmaddox: This link is still broken. See preview: https://deploy-preview-1620--competent-brattain-de2d6d.netlify.com/docs/examples/kubeflow-samples/

config.toml
hefedev: Hi Sarah, I understand your concern. I, too, like to see contributors focused on creating content. I removed statics values form the documentation file because I don't think our approach is permanent and  with a bit of work and adequately structured documents, we can automize this process entirely. Do you think we can stay with this for a time and work on document stuctures? I would love to help.

Also We use the `config.toml.` file for global static values in the past,  it seemed like the obvious place. :)

README.md
hefedev: Agree 💯

config.toml
sarahmaddox: Hallo @hefedev :) Thanks for the explanation and +1 to automating the rest of the process. I'm happy to have the parameters in the `config.yaml` file given that it's a step towards fully-automated data retrieval from the GitHub repo. 

It'd be great to have your help with the process going forward.

... and now I see in the latest commit that you've decided to put the content back into the page. No worries, either way works. 

I think that if you do decide to put the content in the config file, then you should put that comment (the one explaining that the content is in the config file) in the page itself, rather than in the README.

hefedev: Hello @sarahmaddox! , thank you for your support. I'd love to help this fantastic team-building platform of the future :) It's the least we can do to show appreciation for your efforts.
Indeed I changed the implementation and reverted to the original project structure, and I am quite happy with what I had.
 As discussed, the versioning is not reliable, but I feel that with date label will provide some hint to the user, and we now have a plan.

I will send my comment on  other points  asap,  I try to stick to this [instructions](https://github.com/kubeflow/website#creating-and-updating-the-kubeflow-docs) while working at website PRs. 
Thank you once again for helping me get up to speed.

content/docs/examples/kubeflow-samples.md
hefedev: I saw this as well, and it happens on the Master branch. But the link shows up correct on Github Markup and when I build-docs locally. I think it might be a bug but will have to investigate. Do you suggest we do it as part of this review or on a separate issue?

config.toml
sarahmaddox: Sounds good!

content/docs/examples/kubeflow-samples.md
sarahmaddox: Hmm the link is working on the master branch: https://master.kubeflow.org/docs/examples/kubeflow-samples/

It'd be nice if we can fix this before merging the PR. Here are some suggestions to try:

* Try changing from Markdown to HTML for the link. (You can still use a relative URL even in the HTML.
* If that doesn't work, I wonder if the has something to do with the fact that you're using two different types of content block? I don't think that should be a problem, but you could try changing the introductory paragraph to just plain Markdown, not in a block. Or you could put it in one of your new `samples-section` blocks, and make sure the `samples-section` block works even if you pass it no parameters.

sarahmaddox: Thanks for fixing it!

sarahmaddox:
Thank you @hefedev This change greatly improves the usability of the samples page.

/lgtm
/approve

Issue #1611, opened 2020-02-07
Pipelines End-to-end on GCP: fix example & update screenshots


amygdala:
https://www.kubeflow.org/docs/gke/pipelines/pipelines-tutorial/

got a reminder from someone (doing a hosted pipelines-related bug bash) that the screenshots here need to be updated.  And maybe more tweaks? Let's check it during the docs sprint.  I'll try to do it.

sarahmaddox:
/assign @amygdala 

Thanks @amygdala!

sarahmaddox:
@amygdala I've moved this issue to **In progress** in the [doc sprint Kanban board](https://github.com/orgs/kubeflow/projects/21), because the issue is assigned to you. Happy sprinting!

amygdala:
This example no longer runs properly either; will look into it.  Getting some kind of memory reference error related to ksonnet (!).

amygdala:
hi @jinchihe , not sure if you're the right one to ping on this: when I tried running the example, I got the following ksonnet-related segfault error in the `serve` step. 
I don't have visibility as to what's going on in the docker image that this step is using (`gcr.io/ml-pipeline/ml-pipeline-kubeflow-deployer:7775692adf28d6f79098e76e839986c9ee55dd61`)

Are you the right person to look into this? (I'm thinking it might be best to find a non-ksonnet implementation of this step.)

```
Switched to context "kubeflow".
+ cd /src/github.com/kubeflow/kubeflow
+ git checkout v0.2.5
HEAD is now at a4486ccd... Fix bug with updating an existing deployment. (#1475)
+ cd /opt
+ echo 'Initializing KSonnet app...'
+ ks init tf-serving-app
Initializing KSonnet app...
level=info msg="Using context \"kubeflow\" from kubeconfig file \"/root/.kube/config\""
level=info msg="Creating environment \"default\" with namespace \"default\", pointing to \"version:v1.14.8\" cluster at address \"https://kubernetes.default\""
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x456452]

goroutine 1 [running]:
github.com/ksonnet/ksonnet/vendor/github.com/ksonnet/ksonnet-lib/ksonnet-gen/ksonnet.parsePaths(0xc4204c2300, 0x6, 0x1, 0xe)
[...]

PR #1647, opened 2020-02-10
pipelines example: images update

amygdala:
The [example itself](https://github.com/kubeflow/examples/blob/master/pipelines/mnist-pipelines/mnist_pipeline.py) appears to be broken right now though, so this does NOT close the bug.

Partially addresses: https://github.com/kubeflow/website/issues/1611 

cc @sarahmaddox , @joeliedtke

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1647)


joeliedtke:
Thanks for working on this issue @amygdala! Since the tutorial is broken, I have removed it from the site. I will close this PR for now and will re-evaluate merging it when the tutorial is published again.

PR #1698, opened 2020-02-18
fix: curl short --output to small case

portellaa:
The `-O` option is used to download the file using the last path component for the name.
This is just a small change to use the small case option which allows you to specify the output file.

Reference: https://curl.haxx.se/docs/manpage.html#-o

Cheers

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1698)


portellaa:
@googlebot I signed it!

sarahmaddox:
Thanks!
/lgtm
/approve

PR #1699, opened 2020-02-18
fix: add ' ' to yq command spec keypath

portellaa:
Because we are using array notation to select the first position in the plugins under spec section, we need to enclose the keypath with ' ' otherwise bash will complain that can't find a match.

For the sake of consistency i have added it to `metadata.name` path.

Reference from yq help: `write       yq w [--inplace/-i] [--script/-s script_file] [--doc/-d index] sample.yaml 'b.e(name==fr*).value' newValue`

Cheers

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1699)


sarahmaddox:
/ok-to-test
/assign
/lgtm
/approve

Thanks @portellaa!

Issue #1764, opened 2020-03-03
Clarify that GCP basic auth is not supported and remove most references


sarahmaddox:
The basic auth (username/password) option on GCP is already marked as not supported in the docs, but we currently give people instructions on how to use it if they want to. This is leading to problems for some, where the option does not work.

Instead, we should just say it's not supported, without giving usage instructions. In addition, we should remove most references to the basic auth config, leaving only the one in the troubleshooting guide.

PR #1765, opened 2020-03-03
Clarified that GCP basic auth is not supported and removed most references

sarahmaddox:
Fixes https://github.com/kubeflow/website/issues/1764

The basic auth (username/password) option on GCP is already marked as not supported in the docs, but we currently give people instructions on how to use it if they want to. This is leading to problems for some, where the option does not work.

Instead, we should just say it's not supported, without giving usage instructions. In addition, we should remove most references to the basic auth config, leaving only the one in the troubleshooting guide.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1765)

### DIFF
--- a/content/docs/gke/cloud-filestore.md
+++ b/content/docs/gke/cloud-filestore.md
@@ -34,12 +34,6 @@ This guide assumes the following settings:
   export CONFIG_FILE=${KF_DIR}/{{% config-file-gcp-iap %}}
   ```
 
-    Or:
-
-  ```
-  export CONFIG_FILE=${KF_DIR}/{{% config-file-gcp-basic-auth %}}
-  ```
-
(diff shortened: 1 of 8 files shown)


sarahmaddox:
Preview: 
* https://deploy-preview-1765--competent-brattain-de2d6d.netlify.com//docs/gke/deploy/deploy-cli/
* https://deploy-preview-1765--competent-brattain-de2d6d.netlify.com/docs/gke/deploy/deploy-ui/
* https://deploy-preview-1765--competent-brattain-de2d6d.netlify.com/docs/gke/deploy/reasons/
* https://deploy-preview-1765--competent-brattain-de2d6d.netlify.com/docs/upgrading/upgrade/#prerequisites

/assign @richardsliu 

@richardsliu Please would you review this PR.

Issue #1769, opened 2020-03-04
Securing Your Clusters = missing cert-manager.yaml file


anifort:
https://www.kubeflow.org/docs/gke/private-clusters/

It seems that this documentation is not compatible with KFv1. Following the instructions step 7 under Deploy Kubeflow with Private GKE

Remove components which are not useful in private clusters:
cd ${KF_DIR}/kustomize
kubectl delete -f cert-manager.yaml

There is not such a file -> cert-manager.yaml

sarahmaddox:
/assign @zhenghuiwang 

Assigning this to @zhenghuiwang as he's working on an update to the docs for private GKE clusters.

PR #1781, opened 2020-03-06
Update instruction on deleting cert-manager

zhenghuiwang:
Fixes https://github.com/kubeflow/website/issues/1769

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1781)


abhi-g:
Sadly it may not be possilble. https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#removal-directives

So this might be okay to add as instructions in the interim, but we'd really need a cleaned up kfdef for Private GKE.

/lgtm
/approve

PR #1822, opened 2020-03-19
Remove onnx link for seldon in overview page

cliveseldon:
The example has been removed as we need to update the available tech here. Will add a new link when available.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1822)


janeman98:
Thanks @cliveseldon !

/approve
/lgtm

PR #1831, opened 2020-03-22
Fixed program.py when num of line requested is more than actual num

### LINKED ISSUE #1826: Create Reusable Components
https://www.kubeflow.org/docs/pipelines/sdk/component-development/

program.py doesn't work, suggested change:

```#!/usr/bin/env python3
import argparse
from pathlib import Path


# Function doing the actual work (Outputs first N lines from a text file)
def do_work(input1_file, output1_file, param1):
    for line in input1_file.readlines()[:param1]:
        if not line:
            break
        _ = output1_file.write(line)


# Defining and parsing the command-line arguments
parser = argparse.ArgumentParser(description='My program description')
parser.add_argument('--input1-path', type=str,
                    help='Path of the local file containing the Input 1 data.')  # Paths should be passed in, not hardcoded
parser.add_argument('--param1', type=int, default=100, help='Parameter 1.')
parser.add_argument('--output1-path', type=str,
                    help='Path of the local file where the Output 1 data should be written.')  # Paths should be passed in, not hardcoded
args = parser.parse_args()

# Creating the directory where the output file will be created (the directory may or may not exist).
Path(args.output1_path).parent.mkdir(parents=True, exist_ok=True)

with open(args.input1_path, 'r') as input1_file:
    with open(args.output1_path, 'w') as output1_file:
        do_work(input1_file, output1_file, args.param1)
```

janeman98:
Fixes #1826

https://www.kubeflow.org/docs/pipelines/sdk/component-development/

program.py will give error when number of line requested is more than the actual number of line in the input file.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/website/1831)

### DIFF
--- a/content/docs/pipelines/sdk/component-development.md
+++ b/content/docs/pipelines/sdk/component-development.md
@@ -80,7 +80,8 @@ from pathlib import Path
 # Function doing the actual work (Outputs first N lines from a text file)
 def do_work(input1_file, output1_file, param1):
   for x, line in enumerate(input1_file):
-    if x >= param1:                                                                                                      break
+    if x >= param1:                                                                                                      
+      break
     _ = output1_file.write(line)
   
 # Defining and parsing the command-line arguments
--- a/content/docs/pipelines/sdk/component-development.md
+++ b/content/docs/pipelines/sdk/component-development.md
@@ -79,8 +79,7 @@ from pathlib import Path
 
 # Function doing the actual work (Outputs first N lines from a text file)
 def do_work(input1_file, output1_file, param1):
-  for x in range(param1):
-    line = next(input1_file)
+  for line in input1_file.readlines()[:param1]:
     if not line:
       break
     _ = output1_file.write(line)
--- a/content/docs/pipelines/sdk/component-development.md
+++ b/content/docs/pipelines/sdk/component-development.md
@@ -79,11 +79,10 @@ from pathlib import Path
 
 # Function doing the actual work (Outputs first N lines from a text file)
 def do_work(input1_file, output1_file, param1):
-  for line in input1_file.readlines()[:param1]:
-    if not line:
-      break
+  for x, line in enumerate(input1_file):
+    if x >= param1:                                                                                                      break
     _ = output1_file.write(line)
-
+  
 # Defining and parsing the command-line arguments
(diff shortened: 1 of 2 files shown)


joeliedtke:
Preview: https://deploy-preview-1831--competent-brattain-de2d6d.netlify.com/docs/pipelines/sdk/component-development/

joeliedtke:
/assign @Ark-kun 

@Ark-kun, would you please take a look at the proposed change? It would be great to get your thoughts on: 

-  Is the proposed solution readable?
-  In the proposed solution, the input file is read into memory and then sliced into the chunk that will go into the output file. This seems like it could create problems when dealing with potentially huge files. Should this example model how to deal with really large files, or should it mainly show how components accept small data and file paths as arguments?

janeman98:
@joeliedtke @Ark-kun 

If there is a concern on proposed solution on dealing with potentially huge files, instead of changing the program, as Joe suggested, I can add a remark mentioning the program accept small data only.

What do you think?

janeman98:
Hi @joeliedtke @Ark-kun

Any update on this?

content/docs/pipelines/sdk/component-development.md
```diff
@@ -79,12 +79,10 @@ from pathlib import Path
 
 # Function doing the actual work (Outputs first N lines from a text file)
 def do_work(input1_file, output1_file, param1):
-  for x in range(param1):
-    line = next(input1_file)
-    if not line:
-      break
+  for x, line in enumerate(input1_file):
+    if x >= param1:                                                                                                      break
```
joeliedtke, line 83: Looks like a newline was lost. Please move the break to the next line, indented two spaces past the "if"

```
    if x >= param1:
      break
```

janeman98: Thanks @joeliedtke 

Look like the break statement is lost when I merge. And I am afraid that I messed up my branch during merge. May need to revert and push again :-(

joeliedtke: Looks like I was able to update this. Thanks for working on this!   :)

joeliedtke:
/lgtm
/approve

Thanks for working on this @janeman98!

janeman98:
Thanks @joeliedtke !

PR #1842, opened 2020-03-25
Updating cluster config for IBM Cloud

psschwei:
Per https://www.kubeflow.org/docs/started/k8s/overview/: "Kubeflow does not work on Kubernetes 1.16" (assuming this is the case for 1.17 and 1.18 too).


sarahmaddox:
/lgtm
/approve

Thanks!

PR #1845, opened 2020-03-25
Thinned out the website OWNERS file

sarahmaddox:
Reducing the number of **approvers** to a handful, and removing some **reviewers** who have either left the project or not shown much activity in the last year.

Also added @Jeffwan as reviewer, so that he can LGTM the AWS-related PRs.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -7,19 +7,15 @@ approvers:
   - animeshsingh
   - rmgogogo
 reviewers:
-  - abhi-g
   - aronchick
   - carmine
   - dansanche
   - Jeffwan
   - jinchihe
-  - jlewi
   - kunmingg
   - lluunn
   - nickchase
   - pdmack
-  - richardsliu
-  - sarahmaddox
   - terrytangyuan
   - willingc
   - dsdinter
@@ -32,6 +28,4 @@ reviewers:
   - qimingj
   - rileyjbauer
   - vicaire
-  - animeshsingh
-  - rmgogogo
   - janeman98
--- a/OWNERS
+++ b/OWNERS
@@ -1,28 +1,8 @@
 approvers:
   - abhi-g
-  - aronchick
-  - dansanche
-  - ewilderj
-  - jinchihe
   - jlewi
-  - kunmingg
-  - lluunn
-  - nickchase
-  - pdmack
   - richardsliu
   - sarahmaddox
-  - texasmichelle
-  - willingc
-  - dsdinter
-  - Ark-kun
-  - gaoning777
-  - hongye-sun
-  - IronPan
-  - neuromage 
-  - paveldournov
-  - qimingj
-  - rileyjbauer
-  - vicaire
   - joeliedtke
   - animeshsingh
   - rmgogogo
@@ -31,7 +11,7 @@ reviewers:
   - aronchick
   - carmine
   - dansanche
-  - inc0
+  - Jeffwan
   - jinchihe
   - jlewi
   - kunmingg
@@ -40,7 +20,6 @@ reviewers:
   - pdmack
   - richardsliu
   - sarahmaddox
-  - texasmichelle
   - willingc
   - dsdinter
   - Ark-kun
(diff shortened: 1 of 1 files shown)


sarahmaddox:
/assign @abhi-g 
/assign @animeshsingh 
/assign @Jeffwan 
/assign @joeliedtke 

Please would you take a look and let me know if you think I should add/remove any IDs. The goal is to to trim the number of **approvers** to people who have responsibility for the website as a whole, and to include people as reviewers if they're looking after a specific Kubeflow component or section of the docs.

Applying a hold to make sure all reviewers have time to comment:
/hold

joeliedtke:
Would it make sense to keep @neuromage as an approver since his team works on the DSL?

abhi-g:
/lgtm

@joeliedtke  As Sarah says we can leave people as reviewers if needed. Most of the people are not familiar with the current website setup enough to qualify as owners.

OWNERS
```diff
@@ -1,28 +1,8 @@
...
-  - Ark-kun
-  - gaoning777
-  - hongye-sun
-  - IronPan
-  - neuromage 
-  - paveldournov
-  - qimingj
-  - rileyjbauer
-  - vicaire
   - joeliedtke
```
animeshsingh: Not sure of the role for joeliedtke?

OWNERS
```diff
@@ -1,28 +1,8 @@
...
   - animeshsingh
   - rmgogogo
```
animeshsingh: Same above for rmgogogo

animeshsingh: Maybe part of Tech writers team?

OWNERS
```diff
@@ -52,7 +31,6 @@ reviewers:
   - paveldournov
   - qimingj
   - rileyjbauer
-  - thedriftofwords
   - vicaire
```
animeshsingh: We can remove the Approvers from the Reviewers list - ultimately when it comes to approval, you are forced to approve. Also I think reviewers list is quite long - and its meant for automatically rerouting reviewing requests. I believe some folks like 'paveldournov' etc would not want to be assigned review requests until they are asked explictly? So you would have a better way of knowing who are folks actively reviewing here?

animeshsingh: Other way is to go with a longer list, and thin down in future based on frequency of folks reviewing

OWNERS
animeshsingh: Maybe part of Tech writers team?

abhi-g:
@animeshsingh Our general philosophy with owners is to keep it to a minimum
and not expand it without reasons. I think more people can remain in the
"reviewers" list but the owners list should contain people who are
genuinely familiar with the repository and the setup.

OWNERS
sarahmaddox: @animeshsingh That's right, Joe is the tech writer for Kubeflow Pipelines.

OWNERS
sarahmaddox: Agreed that it'd be nice to remove the approvers from the reviewers list, to make the list less overwhelming. @abhi-g do you know: does an approver have the ability to `/lgtm` if they're not listed as a reviewer?

A reason for having people in the reviewers list is to relieve the burden on the approvers - if a reviewer gives an LGTM for their area of the docs, then the approver doesn't need to do a full review. A quick sanity check is usually enough.

abhi-g:
Yes approvers can also lgtm a PR

OWNERS
sarahmaddox: @animeshsingh This is a good question. `rmgogogo` is a lead engineer for Kubeflow Pipelines. I left them in as approver, because they're an active contributor and they're in a different timezone from most others, which is useful if a PR needs to be pushed through. (I'm the only other approver who's not on the US West Coast.)

sarahmaddox:
> Yes approvers can also lgtm a PR
Thanks @abhi-g!

OWNERS
sarahmaddox: I've removed the approvers from the reviewers list.

PR #1860, opened 2020-04-03
Update private-clusters.md

anifort:
I encounter an error creating the service perimeter. ended up being due to insufficient roles. Recommendation to add a note that this role is required


anifort:
@googlebot I signed it!

PR #1861, opened 2020-04-03
Update private-clusters.md

anifort:
PROJECT parameter is set above, there is no need for replication.


anifort:
@googlebot I signed it!

PR #1876, opened 2020-04-07
Correct the get storageclass command

janeman98:
In https://www.kubeflow.org/docs/ibm/install-kubeflow/#ibm-cloud-block-storage-setup
image

The correct command should be 
`kubectl get storageclasses | grep block`

(w/o shell in the front)


janeman98:
Thanks @Jeffwan and @sarahmaddox !

PR #1870, opened 2020-04-06
Upgraded Hugo and Docsy

sarahmaddox:
This PR upgrades the website to the latest version of the Docsy theme and to Hugo 0.68.3. 

Preview: https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/

For special attention:

* From Hugo 0.60 onwards, the default Markdown renderer is Goldmark rather than the current default BlackFriday. I've created issue https://github.com/kubeflow/website/issues/1891 to track any breakages we find due to the new renderer. Then we can decide whether to upgrade and subsequently fix the oddities, or to force the use of the BlackFriday renderer.
* I've enabled the **Create project issue** link, which appears now below **Create documentation issue**. This will help steer people towards adding an issue in kubeflow/kubeflow rather than kubeflow/website when the issue isn't a doc issue. (The link was added to the Docsy templates a while ago but we haven't made use of it up to now.)

Other notes:

* I had a lot of trouble to get the Netlify deployment working. I was getting the following build error: `Error building site: POSTCSS: failed to transform "scss/main.css" (text/css): resource "scss/scss/main.scss_4853eb546e7a6c0898ed71feae7357c0" not found in file cache`. The eventual fix was: Remove all generated `resources/` files then regenerate them (using the `hugo` build command as usual) and force-commit the new `resources/` files.
* I've updated all the custom layouts and resources files that we still need, and removed those that we no longer need. (The reason for no longer needing them is that Docsy has now merged our suggested updates into the theme.)
* I had to add a few more custom files, because previous PRs had made changes to the Docsy theme files in the kubeflow/website repo, and those changes were therefore overwritten when I copied over the new Docsy theme files.

Fixes https://github.com/kubeflow/website/issues/1546
Fixes https://github.com/kubeflow/website/issues/1588

PR #1893, opened 2020-04-13
Remove space between [] and (url)

janeman98:
Ref : https://github.com/kubeflow/website/pull/1870
Remove the invalid space between [] and (url) which cause the following in upgraded Hugo and Docsy:

image

/assign @sarahmaddox


janeman98:
Thanks @sarahmaddox !

PR #1895, opened 2020-04-14
Remove space between [] and (url) in TroubleShooting

janeman98:
Ref : #1870
Remove the invalid space between [] and (url) which cause the [following in upgraded Hugo and Docsy](https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/other-guides/troubleshooting/#pods-stuck-in-pending-state):

image

/assign @sarahmaddox


janeman98:
Thanks @sarahmaddox !

PR #1901, opened 2020-04-16
Update kfctl-istio-dex.md

sachua:
fix "LDAP Seed Users and Groups" dropdown

just one line break that caused the whole text to be rendered incorrectly, tutorial to seed users and groups cannot be followed properly.


sachua:
@googlebot I signed it!

janeman98:
Thanks @sachua !

/ok-to-test

Issue #1891, opened 2020-04-12
Tracking issue: Rendering oddities cause by change to Goldmark renderer


sarahmaddox:
We're upgrading to the latest version of the Docsy theme and to Hugo 0.68.3. From Hugo 0.60 onwards, the default Markdown renderer is Goldmark rather than the current default BlackFriday (see [Hugo docs](https://gohugo.io/getting-started/configuration-markup#goldmark)). We can choose to force the use of BlackFriday rather than Goldmark if necessary, but we're told that Goldmark gives a better overall experience (although both Goldmark and BlackFriday have their own set of oddities).

Let's list any breakages we find in the [preview from PR # 1870](https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/) so that we can decide whether to upgrade then fix the oddities, or to force the use of BlackFriday renderer.

* Numbered list is broken when it includes code snippets: https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/other-guides/kustomize/#applying-the-configuration-to-your-kubeflow-cluster
* Numbered list is broken when it includes an indented bulleted list: https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/gke/deploy/project-setup/

janeman98:
Numbered list is broken here : https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/components/central-dash/overview/#using-kubectl-and-port-forwarding

janeman98:
Link for "feedback" display source in [Beta section](https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/components/metadata/)
![image](https://user-images.githubusercontent.com/52723717/79148875-1aa9b380-7d7b-11ea-9b6b-04eefa863ea8.png)

janeman98:
Same for [Alpha section](https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/components/training/chainer/): 

![image](https://user-images.githubusercontent.com/52723717/79159182-79c3f400-7d8c-11ea-9541-f377a76991ac.png)

janeman98:
Content outside the code block [here](https://deploy-preview-1870--competent-brattain-de2d6d.netlify.com/docs/components/training/tftraining/#troubleshooting): 
![image](https://user-images.githubusercontent.com/52723717/79159795-914fac80-7d8d-11ea-9a24-5b97d90976c7.png)

[...]

PR #1904, opened 2020-04-17
Removed space between [] (url) in ibm iks page

janeman98:
Ref issue #1891 
Removed the space between [] (url) caused by the restriction of Goldmark renderer.

In https://www.kubeflow.org/docs/ibm/iks-e2e/#overview-of-iks
image


janeman98:
Thanks @sarahmaddox !

PR #1913, opened 2020-04-20
Adding Reviewers for Azure Documentation

berndverst:
Adding myself as a reviewer so I can help review docs, especially for Azure.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -29,4 +29,5 @@ reviewers:
   - qimingj
   - rileyjbauer
   - vicaire
+  - berndverst
 


berndverst:
Blocked by https://github.com/kubeflow/internal-acls/pull/239

content/docs/azure/OWNERS
```diff
@@ -0,0 +1,5 @@
+# Azure Docs Owners File
+approvers:  # No new approvers
+reviewers:
+  - berndverst
```
sarahmaddox, line 4: @berndverst I think we should add you as a reviewer in the top-level OWNERS file for the kubeflow/website repository. Would you be happy with that solution? My thinking is that you're adding yourself as a reviewer, not an approver, which means that you're not acquiring the extra permissions that the approver role brings to the doc set as a whole. WDYT?

berndverst: Works for me. Let me do that instead!

berndverst:
@sarahmaddox updated the PR! Thanks!

sarahmaddox:
Thanks!
/lgtm
/approve

PR #1920, opened 2020-04-22
update pipeline version to 0.5.0

hongye-sun:


layouts/shortcodes/pipelines/latest-version.html
```diff
@@ -1 +1 @@
-0.4.0
\ No newline at end of file
+0.5.0
```
sarahmaddox, line 1: Please would you remove the empty line at the end of the file. I think you need to do this update via the command line, as the web UI always puts the empty line back.

The empty line causes problems in rendering the version number on the page.

hongye-sun: Fixed, thanks.

PR #1970, opened 2020-05-27
OWNERS for pipelines folder

Bobgy:
/assign @rmgogogo @joeliedtke 
Propose adding engs to approvers of pipelines folder, so that we can fix minor issues much faster.

### DIFF
--- a/content/en/docs/pipelines/OWNERS
+++ b/content/en/docs/pipelines/OWNERS
@@ -1,0 +1,13 @@
+approvers:
+  - Ark-kun
+  - Bobgy
+  - hongye-sun
+  - IronPan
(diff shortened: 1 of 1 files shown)


content/en/docs/pipelines/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+  - Bobgy
+  - IronPan
+  - neuromage
```
rmgogogo, line 4: add Joe to approver

Bobgy: joe is already root folder approver, there's no need to duplicate here

PR #1971, opened 2020-05-27
Make KFP install command zsh compatible

Bobgy:
/assign @joeliedtke @rmgogogo 
Fixes https://github.com/kubeflow/pipelines/issues/2767#issuecomment-571881081

As zsh is becoming the default for macos, we should expect more users for zsh. Update the instruction to be zsh compatible. (I'm also using zsh now.)

### DIFF
--- a/content/en/docs/pipelines/installation/standalone-deployment.md
+++ b/content/en/docs/pipelines/installation/standalone-deployment.md
@@ -72,9 +72,9 @@ See the Google Kubernetes Engine (GKE) guide to
 
      ```
      export PIPELINE_VERSION={{% pipelines/latest-version %}}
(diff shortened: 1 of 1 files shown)


rmgogogo:
wondering whether it's the fix the for the issue.
Without this fix, you can reproduce the issue in mac OS (zsh)?

/lgtm

Bobgy:
@rmgogogo Yes, without the fix, I can reproduce it in zsh for both macos and linux.

joeliedtke:
Preview: https://deploy-preview-1971--competent-brattain-de2d6d.netlify.app/docs/pipelines/installation/standalone-deployment/

joeliedtke:
/approve

Thanks for updating this, @Bobgy!

PR #1975, opened 2020-06-02
Auto-generate api reference doc

jingzhang36:
Auto-generated api reference doc from files in pipelines repo: https://github.com/kubeflow/pipelines/blob/master/backend/api/swagger/kfp_api_single_file.swagger.json and the proto files in folder https://github.com/kubeflow/pipelines/tree/master/backend/api.

This html doc is auto generated. Therefore we don't usually make manual revision to it, since any manual revision could be easily overridden when this file is updated via another auto-generation. If anything needs to be changed in this html, they should be done in pipelines repo and then re-generate this html from the files in pipeline repo.

We recently updated our api related files in  pipeline repo (https://github.com/kubeflow/pipelines/pull/3758) and therefore regenerate this html.

/assign @joeliedtke 
/cc @rmgogogo 
/cc @Bobgy


Bobgy:
Thanks @jingzhang36!
As already reviewed in upstream repo. We can get this in.

Bobgy:
The doc can be previewed at https://deploy-preview-1975--competent-brattain-de2d6d.netlify.app/docs/pipelines/reference/api/kubeflow-pipeline-api-spec/.
looks great

/lgtm
/approve

PR #1976, opened 2020-06-04
kf dashboard

yahavb:
access to kubeflow dashboard with kfctl_aws.v1.0.2 supports:
istio-system svc/istio-ingressgateway instead of ingress istio-ingress -n istio-system


janeman98:
Thanks @yahavb for submitting this PR. 
Can you please sign the CLA?

/assign @Jeffwan

Jeffwan:
@yahavb 

Thanks for the contribution. While, I don't think we expose LB by default. This is not the default setting, right? Have you changed service type to LoadBalancer?

yahavb:
I dont think we need this change anymore. @dalbhanj helped me understand the secure setup with auth.

PR #1992, opened 2020-06-10
Add a small example of listing pipeline with a filter

jingzhang36:
This PR adds a small example of listing pipelines with a filter. Moreover, since we now have multiple sdk examples on this page, I also added explicit example id.


jingzhang36:
Preview at https://deploy-preview-1992--competent-brattain-de2d6d.netlify.app/docs/pipelines/tutorials/sdk-examples/
/assign @joeliedtke

Issue #2001, opened 2020-06-16
Deploy failing for v1.0 branch missing npm package


jlewi:
Per #1984 I cut a 1.0 branch. Netlify auto-triggered a deploy but the deploy is failing. The logs are
below.

It looks like an npm package `postcss-cli` is missing. 

So far it looks like PR previews and master is deploying fine. Not sure why that is.

```
3:24:42 PM: Build ready to start
3:24:44 PM: build-image version: c23500190e408b04bdfcf9703d16c460ef03b813
3:24:44 PM: build-image tag: v3.3.15
3:24:44 PM: buildbot version: 85ab9be17c6555a1d991d1cf7b0dbb73c18d0b92
3:24:44 PM: Fetching cached dependencies
3:24:44 PM: Starting to download cache of 426.6MB
3:24:49 PM: Finished downloading cache in 4.857444615s
3:24:49 PM: Starting to extract cache
3:25:01 PM: Finished extracting cache in 12.269516078s
3:25:01 PM: Finished fetching cache in 17.226443229s
3:25:01 PM: Starting to prepare the repo for build
3:25:02 PM: Preparing Git Reference refs/heads/v1.0-branch
3:25:04 PM: Starting build script
3:25:04 PM: Installing dependencies
3:25:04 PM: Python version set to 2.7
3:25:04 PM: Started restoring cached node version
3:25:08 PM: Finished restoring cached node version
3:25:08 PM: v8.17.0 is already installed.
3:25:09 PM: Now using node v8.17.0 (npm v6.13.4)
3:25:09 PM: Started restoring cached build plugins
3:25:09 PM: Finished restoring cached build plugins
3:25:09 PM: Attempting ruby version 2.3.6, read from environment
3:25:09 PM: Started restoring cached ruby version
3:25:10 PM: Finished restoring cached ruby version
3:25:11 PM: Using ruby version 2.3.6
3:25:12 PM: Successfully installed bundler-2.1.4
3:25:12 PM: 1 gem installed
3:25:12 PM: Using PHP version 5.6
3:25:12 PM: 5.2 is already installed.
3:25:12 PM: Using Swift version 5.2
3:25:12 PM: Started restoring cached go cache
3:25:13 PM: Finished restoring cached go cache
3:25:13 PM: Installing Go version 1.10
3:25:13 PM: unset GOOS;
3:25:13 PM: unset GOARCH;
[...]

PR #2004, opened 2020-06-18
Adds configuration for Netlify deployments to branches

joeliedtke:
Adds configuration to specify the version of Hugo that is used to deploy branches.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,3 +8,6 @@
 [context.production.environment]
   HUGO_VERSION = "0.68.3"
   HUGO_ENV = "production"
+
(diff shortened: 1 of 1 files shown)


joeliedtke:
/assign @jlewi 

@jlewi, Do you have bandwidth to review this PR?  This is for Kubeflow/website #2001. The issue appears to be that branches are being built with an old version of Hugo. This update specifies the version of Hugo to use when building branches.

I'm not sure if we can fully test this without merging the change and cherry picking it into the 1.0 branch. It seems to be harmless for previews, so it should be safe to try.

Issue #2011, opened 2020-06-23
Updates TFJob and PytorchJob document to disable istio sidecar inject


shawnzhu:
## Problem

YAML resource description from this article doesn't work: https://www.kubeflow.org/docs/components/training/tftraining/#what-is-tfjob

So as that from https://www.kubeflow.org/docs/components/training/pytorch/#monitoring-a-pytorch-job

It leads to several issue reports including: 

* https://github.com/kubeflow/pytorch-operator/issues/258
* https://github.com/kubeflow/katib/issues/1042
* https://github.com/kubeflow/kubeflow/issues/4935

## Solution

So adding one annotation `sidecar.istio.io/inject: "false"` will get the examples work in above two articles. But it would need some note like what's expected from https://github.com/kubeflow/kubeflow/issues/4935

shawnzhu:
Given `tfjob` example, the pods will be in `Error` status when following the doc because of this problem:

```
 kubectl logs dist-mnist-for-e2e-test-ps-0 -n <my-profile-name> -c tensorflow
/usr/local/lib/python2.7/dist-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Traceback (most recent call last):
  File "/var/tf_dist_mnist/dist_mnist.py", line 303, in <module>
    tf.app.run()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 124, in run
    _sys.exit(main(argv))
  File "/var/tf_dist_mnist/dist_mnist.py", line 110, in main
    mnist = input_data.read_data_sets(FLAGS.data_dir, one_hot=True)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/mnist.py", line 240, in read_data_sets
    source_url + TRAIN_IMAGES)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/base.py", line 208, in maybe_download
    temp_file_name, _ = urlretrieve_with_retry(source_url)
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/contrib/learn/python/learn/datasets/base.py", line 165, in wrapped_fn
    return fn(*args, **kwargs)
[...]

PR #2017, opened 2020-06-24
Adds note and example annotation for tfjob

shawnzhu:
Supports #2011


shawnzhu:
@animeshsingh @gaocegege Thank you!

PR #2025, opened 2020-06-28
update: any of the following ~~are~~ (is) true

liusy182:


liusy182:
@googlebot I signed it!

joeliedtke:
@liusy182, thank you for suggesting this change. In the following sentence, is/are refers to "any of the following" which provides several options. Since the noun that is/are refers to is plural, "are" is correct in this case.

Port-forwarding typically does not work if any of the following [is/are] true:

PR #2030, opened 2020-07-07
Update metadata.md docs to show images

Mithul:
Fixed  tags being displayed as raw HTML source instead of displaying the actual image


Mithul:
I can confirm that it works in the deployment preview at `/docs/components/metadata/`

PR #2041, opened 2020-07-16
add namespace warning at top of standalone upgrade section

amygdala:

PR #2049, opened 2020-07-23
Revert "add namespace warning at top of standalone upgrade section"

Bobgy:
Reverts kubeflow/website#2041

/assign @amygdala 
Shall we revert this first before resolving https://github.com/kubeflow/pipelines/issues/4261


amygdala:
thanks, makes sense. 
(I wonder if in the meantime we should add a note explicitly indicating that this doesn't work with marketplace installs).

PR #2059, opened 2020-07-29
Fix management cluster setup instructions

jlewi:
* Instructions should reference the Makefile
* We will now install CNRM in workload identity mode not namespace mode
  per kubeflow/gcp-blueprints#13


jlewi:
/approve

@Bobgy Looks like you might not be in the OWNERs file; you probably should be

PR #2074, opened 2020-08-03
Update oidc auth secret to match the 1.1 release in the IBM multi user doc

Tomcli:
The oidc secret in the IBM multi user doc is not matching with the default oidc secret for the Kubeflow 1.1 release. Thus, updating the doc to have to correct default secret.

Secret reference: https://github.com/kubeflow/manifests/blob/master/stacks/ibm/application/oidc-authservice/kustomization.yaml#L20

### DIFF
--- a/content/en/docs/ibm/deploy/install-kubeflow.md
+++ b/content/en/docs/ibm/deploy/install-kubeflow.md
@@ -260,7 +260,8 @@ The scenario is a GitHub organization owner can authorize its organization membe
         - id: kubeflow-oidc-authservice
           redirectURIs: ["/login/oidc"]
           name: 'Dex Login Application'
-          secret: pUCnCOY80SnXgjibTYM0ZWNzY3xreNGQok
+          # Update the secret below to match with the oidc authservice.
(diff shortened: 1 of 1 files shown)


content/en/docs/ibm/deploy/install-kubeflow.md
```diff
@@ -260,7 +260,8 @@ The scenario is a GitHub organization owner can authorize its organization membe
         - id: kubeflow-oidc-authservice
           redirectURIs: ["/login/oidc"]
           name: 'Dex Login Application'
-          secret: pUCnCOY80SnXgjibTYM0ZWNzY3xreNGQok
+          # Update the secret below to match with the oidc authservice.
+          secret: pUBnBOY80SnXgjibTYM9ZWNzY2xreNGQok
```
shawnzhu, line 264: 👍 it matches that from https://github.com/kubeflow/manifests/blob/14de0c7dc27231c0a5840f99f125acc73a351ce1/stacks/ibm/application/oidc-authservice/kustomization.yaml#L20

Tomcli:
@animeshsingh can you approve this so people won't get blocked when they copy the config from our instructions?

PR #2083, opened 2020-08-04
Add Slack link to WG Trainining channel

terrytangyuan:
Also removed link for tf-operator since we will want to start promoting the WG channel instead.


jlewi:
Instead of duplicating the information here should we just refer people to the appropriate community/wgs page for working groups?

terrytangyuan:
I thought about doing that too but the channel name is in the yaml file of the specific working group, which may not be intuitive to many people. There has to be some sort of duplication somewhere.

PR #2089, opened 2020-08-06
Hide upgrade instructions for AI Platform Pipelines managed storage instructions

Bobgy:
/assign @rmgogogo 

before we figure out a more detailed instructions that can make sure there's no user data loss


rmgogogo:
Maybe you can make the sentence be more user friendly as current it can work. You can point it out that current in-cluster PVC is not deleted but no garentee in the future and we suggest managed.

Bobgy:
> Maybe you can make the sentence be more user friendly as current it can work. You can point it out that current in-cluster PVC is not deleted but no garentee in the future and we suggest managed.

I want to at least confirm with marketplace team we can rely on this for a while first.

PR #2102, opened 2020-08-10
Update pipeline version to the 1.0.0

JohanWork:
update export PIPELINE_VERSION=0.5.1 to point to 1.0.0 instead of the old version 0.5.1


layouts/shortcodes/pipelines/latest-version.html
```diff
@@ -1 +1 @@
-0.5.1
\ No newline at end of file
+1.0.0
```
NikeNano, line 1: Why version 1.0.0 and not the latest?

JohanWork: https://github.com/kubeflow/pipelines 1.0.0 seams to be the latest one to me

Bobgy:
master branch is already 1.0.0 https://github.com/kubeflow/website/blob/master/layouts/shortcodes/pipelines/latest-version.html

JohanWork:
> master branch is already 1.0.0 https://github.com/kubeflow/website/blob/master/layouts/shortcodes/pipelines/latest-version.html

sorry your right, got confused since https://www.kubeflow.org/docs/pipelines/installation/standalone-deployment/#deploying-kubeflow-pipelines is still showing 0.5.1 . Why is that @Bobgy ?

Bobgy:
kubeflow.org is not showing master, because everyone is working on kubeflow 1.1

JohanWork:
aa ok thnx!

Issue #1984, opened 2020-06-09
[Release 1.1] Release Website 1.1


jlewi:
Opening this issue to track doing a release of the website for Kubeflow 1.1

* We still need to find an owner to drive this (see kubeflow/kubeflow#5022)

* The first step is cutting a 1.0 branch from master providing a stable link to the docs so we can begin updating the docs on master.

* Docs for versioning the website: https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#version-the-website

jlewi:
It looks like we might want to branch 1.0 on
https://github.com/kubeflow/website/tree/61b5ed89590fe961c26d18da30969f3841dbf93a

After that it looks like we start getting commits related to 1.1

RFMVasconcelos:
Hi Jeremy,

As soon as I get access to the repo I'll push that branch up.

As far as netlify, is that automatic or are there any additional steps?

jlewi:
@RFMVasconcelos 

> As soon as I get access to the repo I'll push that branch up.

Please read: 
https://groups.google.com/forum/#!topic/kubeflow-discuss/pGflW6m9Xrg

And please file PRs.

> As far as netlify, is that automatic or are there any additional steps?
Is what automatic?

RFMVasconcelos:
Hi @jlewi, thank you for the guidance!!

> Please read:
I'm going through all the guidelines. Signed CLA, ready to start contributing :) 

> And please file PRs.
Will do! Started out with this one: https://github.com/kubeflow/website/pull/1999

> Is what automatic?
I haven't used netlify before. Was wondering what I needed to know to render the website, or if it was automatic.

8bitmp3:
Hey @RFMVasconcelos I left some suggestions in your PR https://github.com/kubeflow/website/pull/1999. (I'm not an official editor but a few things may need to be reviewed by you before the official PR review.)

jlewi:
@RFMVasconcelos The website is automatically updated once PR is merged.

Checkout the quickstart: https://github.com/kubeflow/website#quickstart

When you create a PR it will automatically create preview for you of the rendered website.

8bitmp3:
[...]

PR #2111, opened 2020-08-12
Add v1.0 to Kubeflow docs config.toml (v1.1 is the latest)

8bitmp3:
This PR updates Kubeflow docs' `config.toml` so the versions drop down shows the link to the 1.0 docs (v1.1 is the latest) @jlewi . Note that the link points to `https://v1-0-branch.kubeflow.org` (as per https://github.com/kubeflow/website/issues/1984#issuecomment-671375882).

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -171,6 +171,11 @@ githubbranch = "v1.1-branch"
   githubbranch = "v0.7-branch"
   url = "https://v0-7.kubeflow.org"
 
+[[params.versions]]
+  version = "v1.0"
+  githubbranch = "v1.0-branch"
+  url = "https://v1-0-branch.kubeflow.org"
+
 # Docsy: User interface configuration
 [params.ui]
 # Docsy: Enable to show the side bar menu in its compact state.


jlewi:
https://deploy-preview-2111--competent-brattain-de2d6d.netlify.app/
Looks great!

/lgtm

I can no longer approve PRs;

@animeshsingh @Bobgy Can one of you approve?

/assign @animeshsingh 
/assign @Bobgy

config.toml
```diff
@@ -171,6 +171,11 @@ githubbranch = "v1.1-branch"
   githubbranch = "v0.7-branch"
   url = "https://v0-7.kubeflow.org"
 
+[[params.versions]]
+  version = "v1.0"
+  githubbranch = "v1.0-branch"
+  url = "https://v1-0-branch.kubeflow.org"
```
Bobgy, line 177: Why do we use `v1-0-branch` as url prefix? It seems previous versions were using `v0-7.kubeflow.org`

Bobgy: I think @jlewi meant https://v1-0-branch.kubeflow.org is available even before you set up this config, but this if we follow previous conventions, we can make https://v1-0.kubeflow.org also available.

Please correct me if there's something I missed.

8bitmp3: @Bobgy Let's make `v1-0.kubeflow.org` available - do you know how? I'm looking at https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#version-the-website. Thank you!

(fyi I linked v1.0 to `https://v1-0-branch.kubeflow.org` because it was the only URL that worked and fixing `config.toml` seemed urgent - some folks were looking for the v1.0 docs.) cc @jlewi

Bobgy:
Makes sense
let's get this in first

/lgtm
/approve

PR #2124, opened 2020-08-12
Add more robust check for empty doc pages

knkski:
Checks that there is actual content after front matter section when adding outdated banner to docs. Treats docs with only "strippable" characters after the front matter section as having no content.


8bitmp3 (commented):
can you PTAL, this has gone stale

/assign @animeshsingh @Bobgy

Bobgy:
/lgtm
/approve
Thanks!

PR #2139, opened 2020-08-18
Fix bad link in Fairing Docs

jinchihe:
Fix bad link in Fairing Docs


jinchihe:
Checked `deploy preview`. Looks fine. Thanks.

PR #2170, opened 2020-09-01
doc(gcp): fix broken format in management cluster setup

Bobgy:
/assign @rmgogogo


Bobgy:
Bypassing review, because I can verify format is already fixed in https://deploy-preview-2170--competent-brattain-de2d6d.netlify.app/docs/gke/deploy/management-setup/.

PR #2175, opened 2020-09-01
doc(kfp): update multi-user isolation doc title and supported platforms

Bobgy:
/assign @joeliedtke @8bitmp3

### DIFF
--- a/content/en/docs/pipelines/multi-user.md
+++ b/content/en/docs/pipelines/multi-user.md
@@ -15,7 +15,7 @@ for the common Kubeflow multi-user operations including the following:
 
 Note, Kubeflow Pipelines multi-user isolation is only supported in
 [the full Kubeflow deployment](/docs/pipelines/installation/overview/#full-kubeflow-deployment)
-starting from Kubeflow v1.1 and **currently** on all platforms except OpenShift. For latest status about platform support, read [kubeflow/manifests#1364](https://github.com/kubeflow/manifests/issues/1364#issuecomment-668415871).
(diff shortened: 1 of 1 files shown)


8bitmp3 (changes requested):
@Bobgy Hi, what do you think about this proposal: 

`For the latest status... refer to...`

- Adding a definite article "the" (I'm not 100% sure here but I think it's right)
- Replacing "read" with "refer to" for a11y


(Additionally, I'd keep each line at 80 characters max)

WDYT?

Thank you!

Bobgy:
@8bitmp3 Thank you for the suggestions! It does sound better.

> (Additionally, I'd keep each line at 80 characters max)

I don't really want to enforce this, because editing existing paragraphs will incur non-trivial edits to keep the rule.
Do you find it difficult to read in some situations? So far, I've been using VS Code with wrapping enable to edit markdown files and not manually wrapping lines make the whole editing a lot easier.

8bitmp3:
> (Additionally, I'd keep each line at 80 characters max)
I don't really want to enforce this, because editing existing paragraphs will incur non-trivial edits to keep the rule.
Do you find it difficult to read in some situations? So far, I've been using VS Code with wrapping enabled to edit markdown files. Not manually wrapping lines make the whole editing a lot easier.

@Bobgy Ok! Whatever makes it easier to review for you.

I think some open-source projects, including at Google, have a linter in docs tests that sort of checks for the 80 chart limit, many don't have it so it's not enforced.

Certain IDEs may not have the default setting set to auto-wrap (do the less modern ones with less GUI have them?). So if a user forks the repo and starts working via the CLI, their view may not be the same as ours. the GitHub web UI, VSCode, IDEA and Xcode have done an excellent job at this.

Check this out - https://about.gitlab.com/blog/2016/10/11/wrapping-text/ - to wrap or not to wrap dos and don'ts 🙂

image

PR #2181, opened 2020-09-09
Remove issue GoogleContainerTools/kpt#539 from the docs (solved)

Saul-R:
Seems fixed:
https://github.com/GoogleContainerTools/kpt/issues/510 / https://github.com/kubernetes-sigs/kustomize/pull/2636

The issue described in the guide is not appearing anymore:

```
> cd ./management && make get-pkg
mkdir -p  ./upstream
kpt pkg get https://github.com/kubeflow/manifests.git/gcp/v2/management@1f948ed ./upstream/management
fetching package /gcp/v2/management from https://github.com/kubeflow/manifests to upstream/management
> echo $?
0
```

### DIFF
--- a/content/en/docs/gke/deploy/deploy-cli.md
+++ b/content/en/docs/gke/deploy/deploy-cli.md
@@ -94,16 +94,6 @@ one if you haven't already.
    make get-pkg
    ```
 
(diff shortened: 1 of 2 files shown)


Saul-R:
Still going on, my bad.

PR #2183, opened 2020-09-10
Fix TLS certificate usage in IBM cloud Kubernetes

shawnzhu:
I went through these steps after the certificate expired, then discovered this fix.

Notice that IBM Cloud will only save renewed certificates stored in `default` namespace, so it will to change the secret name and its namespace when copying it into `istio-system` namespace.


animeshsingh (approved):
/lgtm
/approve

PR #2188, opened 2020-09-12
Fix links for cluster manifests in localcluster-deployment

iancoffey:
These docs seem to be broken as the manifest locations seems to have changed.

These updates worked for me. Thanks!


alfsuse:
Hi @iancoffey  manifest seems to be in the right place from what I see : I can find them in the master branch  [here](https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/env/platform-agnostic-pns).
Which version of KF pipelines where you trying to install?
Notice the -pns version is to install KF pipelines with non-docker runtimes, so if you were installing on docker based the standard platform-agnostic works fine.

PR #2212, opened 2020-09-25
Correct metadata inaccuracy in the output viewer documentation

lucinvitae:
This corrects a misleading sentence in the output viewer documentation, where the docs suggest the file name matters, but in reality the artifact name is what matters.

### DIFF
--- a/content/en/docs/pipelines/sdk/output-viewer.md
+++ b/content/en/docs/pipelines/sdk/output-viewer.md
@@ -61,9 +61,13 @@ See the [sample description and links below](#example-source).
 ## Writing out metadata for the output viewers
 
 The pipeline component must write a JSON file specifying metadata for the
-output viewer(s) that you want to use for visualizing the results. The file name 
-must be `/mlpipeline-ui-metadata.json`, and the component must write the file
-to the root level of the container filesystem.
+output viewer(s) that you want to use for visualizing the results. The component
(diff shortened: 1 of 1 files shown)


lucinvitae:
I'm not sure why CI is failing for this PR. It runs locally for me with `hugo server -D` just fine.

Screenshot:
Screen Shot 2020-09-25 at 10 42 10 AM

alfsuse (approved):

lucinvitae:
Closing and reopening to try to fix CI issues as suggested by reviewer.

CI issue shows `fatal error: concurrent map read and map write` which looks a bit like https://github.com/gohugoio/hugo/issues/2549

Looks like there’s some more context… I think the -race option may show it’s a data race. For this issue, @bep opened https://github.com/gohugoio/hugo/issues/5926 in May and https://github.com/golang/go/issues/39807 in June… @andreasf tried to fix in https://github.com/gohugoio/hugo/pull/7507/files but @bep rejected, citing that template rendering in hugo has always been done in parallel, and that he doesn’t want to maintain custom parser code, and appears to be waiting on a fix in golang itself.

andreasf:
Hey, I've had a patch for this in my Golang branch for a while. The fix in Golang is the same as in the Hugo copy of text/template.

In our Hugo pipeline, we wrapped `hugo` with a script that tries to rerun it a couple of times if it fails with a `concurrent map access` error message.

PR #2219, opened 2020-09-28
WIP: Add OWNERS to Upgrading Kubeflow deployment

RFMVasconcelos:
Looking into the feature requests, having an upgradeability path seems to be referred often. Wondering who should own these docs.

@jlewi do you have someone in mind?

cc @8bitmp3

### DIFF
--- a/content/en/docs/upgrading/OWNERS
+++ b/content/en/docs/upgrading/OWNERS
@@ -1,3 +1,4 @@
 approvers:
 
 reviewers:
+  - rfmvasconcelos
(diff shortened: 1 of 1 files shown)


8bitmp3:
I'm not a domain expert but happy to be just a reviewer @RFMVasconcelos Thank you!

alfsuse:
> I'm not a domain expert but happy to be just a reviewer @RFMVasconcelos Thank you!

+1 I can help here

PR #2232, opened 2020-10-01
Improved Docs

5HAD0W-P1R4T3:


joeliedtke:
Markdown renders list items with "1." as part of an ordered list. So, the HTML renders with the correct number without specifying it in the Markdown file. Which makes the content easier to maintain.

PR #2240, opened 2020-10-01
Improve docs 

MDARBAAJ:


unnown:
Please mark as invalid and spam, this is an obvious attempt to get a free t-shirt by 'contributing'(hacktoberfest)
This will also ban the user for trying to abuse the system by doing marginal effort

joeliedtke:
Thanks for your suggestion. However, in this case that section describes one example.

PR #2244, opened 2020-10-01
improve docs

sethmcagit:


joeliedtke:
Thank you for your suggestions. However, I feel that they do not improve that doc. Please review our [style guide](https://www.kubeflow.org/docs/about/style-guide/) to learn more about contributing to the Kubeflow docs.

PR #2261, opened 2020-10-06
Removed "outdated" from version agnostic "Getting started" docs

RFMVasconcelos:
The entrance for users when they click `Getting started` on Kubeflow.org is the `Installing Kubeflow` page, which appears outdated:

image

At a first glance, this is more of a gateway, which seems version agnostic. And we're telling ALL users that our main docs page is outdated. It surely has space to improve, but at the current pace, I argue that we should remove the outdated banner and later on create a coordinated effort to update it. 

Same for `Getting Started > Kubeflow Overview`.

@8bitmp3 @jlewi @jbottum what are your thoughts on this?

### DIFF
--- a/content/en/docs/started/getting-started.md
+++ b/content/en/docs/started/getting-started.md
@@ -156,25 +156,6 @@ The matrix is therefore an alternative way of accessing the information in the
   </table>
 </div>
 
-## Installing command line tools
-
-The following information is useful if you need or prefer to use command line
-tools for deploying and managing Kubeflow:
-
-* Download the kfctl binary from the
-  [Kubeflow releases page](https://github.com/kubeflow/kfctl/releases/).
-
-* Follow the kubectl installation and setup instructions from the [Kubernetes
-  documentation](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
-  As described in the Kubernetes documentation, your kubectl
-  version must be within one minor version of the Kubernetes version that you
-  use in your Kubeflow cluster.
-
-* Follow the kustomize installation and setup instructions from the guide to
-  [kustomize in Kubeflow](/docs/other-guides/kustomize/).
-  Make sure that you have the minimum required version of kustomize:
-  <b>{{% kustomize-min-version %}}</b> or later.
-
 ## Understanding the Kubeflow versioning policies
 
 With the launch of Kubeflow {{% kf-latest-version %}}, the Kubeflow community attributes 
--- a/content/en/docs/started/getting-started.md
+++ b/content/en/docs/started/getting-started.md
@@ -2,7 +2,7 @@
 title = "Installing Kubeflow"
 description = "Overview of installation choices for various environments"
 weight = 20
-                    
+
 +++
 
 This document provides information about setting up Kubeflow in various
@@ -28,8 +28,8 @@ in the "Doc diagrams" folder in the Kubeflow team drive.
 The following diagram gives an overview of the options for deploying Kubeflow:
 
 <div>
-  <object type="image/svg+xml" 
-    data="/docs/images/kubeflow-getting-started-diagram.svg" 
+  <object type="image/svg+xml"
+    data="/docs/images/kubeflow-getting-started-diagram.svg"
     alt="A diagrammatic overview of Kubeflow deployment options"
     class="mt-3 mb-3 border border-info rounded">
   </object>
@@ -84,9 +84,9 @@ Choose the guide for your operating system or environment:
   * To use MiniKF (mini Kubeflow) on Google Cloud Platform, follow the guide to
     [MiniKF on GCP](/docs/started/workstation/minikf-gcp/).
 
-## Configuration quick reference
+## Deployment quick reference
 
-Below is a matrix of the platforms where you can deploy Kubeflow and the 
+Below is a matrix of the platforms where you can deploy Kubeflow and the
 corresponding manifest files that specify the default configuration for each
 platform. The matrix shows the same manifests as the installation guides.
 The matrix is therefore an alternative way of accessing the information in the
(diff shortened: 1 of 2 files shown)


RFMVasconcelos:
@8bitmp3 since I started the PR, you have to be the one to approve :) Given that this is under our responsibility.

content/en/docs/started/getting-started.md
```diff
@@ -4,10 +4,6 @@ description = "Overview of installation choices for various environments"
 weight = 20
                     
 +++
-{{% alert title="Out of date" color="warning" %}}
```
8bitmp3, line 7: Thank you @RFMVasconcelos 

/lgtm

content/en/docs/started/kubeflow-overview.md
```diff
@@ -4,10 +4,6 @@ description = "How Kubeflow helps you organize your ML workflow"
 weight = 10
                     
 +++
-{{% alert title="Out of date" color="warning" %}}
```
8bitmp3, line 7: 👍  @RFMVasconcelos

8bitmp3:
@joeliedtke @Bobgy Can you please take a quick look - I think this is a good idea by @RFMVasconcelos

jlewi:
Thanks!

Preview: 
https://deploy-preview-2261--competent-brattain-de2d6d.netlify.app/docs/started/getting-started/

High level comment: I think the main purpose of this page should be to redirect to one of the sections with detailed instructions and clear owners.

Along those lines should we delete the section "Configuration Quick Reference"
https://github.com/kubeflow/website/pull/2261/files#diff-1705b3cb96d57a841492edf36721efc6R87

* This is duplicating information that should be in the detailed sections.
* Not all installation options are using this config files. Removing those distributions from the table could be misleading
   * i.e. users think a given distribution isn't supported when in fact it is it just uses a different deployment option.

Likewise should we remove the section "Installing Command Line Tools"

* The command line tools needed will depend on the installation option you are choosing.

For the section "existing Kubernetes Cluster"
https://deploy-preview-2261--competent-brattain-de2d6d.netlify.app/docs/started/getting-started/#installing-kubeflow-on-an-existing-kubernetes-cluster

Can we change the title to something like "conformant Kubernetes Cluster" ? and likewise add an explanation along the lines of if none of the platform/cloud specific use directions work for you, you can try these instructions for a generic cluster.

On the overview page should we delete the section "kubeflow-command-line-interface-cli"
https://deploy-preview-2261--competent-brattain-de2d6d.netlify.app/docs/started/kubeflow-overview/#kubeflow-command-line-interface-cli

kfctl is specific to certain ways of deploying Kubeflow so its not always used.

RFMVasconcelos:
Hey @jlewi,
Thank you for your always sharp commentary! Replying inline:

### Getting Started Page
> High level comment: I think the main purpose of this page should be to redirect to one of the sections with detailed instructions and clear owners.

Very much agreed! This should overarching index page that maps content to needs, not a version-specific content page.

> Along those lines **should we delete the section "Configuration Quick Reference"**
> https://github.com/kubeflow/website/pull/2261/files#diff-1705b3cb96d57a841492edf36721efc6R87
> This is duplicating information that should be in the detailed sections.
> Not all installation options are using this config files. Removing those distributions from the table could be misleading
> i.e. users think a given distribution isn't supported when in fact it is it just uses a different deployment option.

It seems that this table is still serving the purpose of mapping deployment platforms to their docs. Removing it without a replacement strategy for mapping docs might be risky IMO. 

> Likewise **should we remove the section "Installing Command Line Tools"**
> The command line tools needed will depend on the installation option you are choosing.

As you mention, these appear to be deployment-specific. I'll add a commit to remove.

> For the section "existing Kubernetes Cluster"
https://deploy-preview-2261--competent-brattain-de2d6d.netlify.app/docs/started/getting-started/#installing-kubeflow-on-an-existing-kubernetes-cluster
Can we change the title to something like "conformant Kubernetes Cluster" ? and likewise add an explanation along the lines of if none of the platform/cloud specific use directions work for you, you can try these instructions for a generic cluster.

I can do this. I'm starting to question the reason for it to be 1 full section with 1 sentence. Let me work on a proposal to summarize/compress these methods and remove cognitive overhead.

**Overall comment:**
Looking at these comments it seems that we want to have this page be a map, and at the moment it is a confusing one:
1. The diagram is mapping to anchors in the page
2. "Install Kubeflow" and "Configuration quick reference" sections seem to be somehow redundant in mapping to specific docs. 

**Actions:**
1. Delete the section "Installing Command Line Tools"
2. Find best way to display what is not in "Install Kubeflow" and "Configuration quick reference" section
3. Improve or remove section "existing Kubernetes Cluster" based on 2.


### Overview Page
> On the overview page should we delete the section "kubeflow-command-line-interface-cli"
https://deploy-preview-2261--competent-brattain-de2d6d.netlify.app/docs/started/kubeflow-overview/#kubeflow-command-line-interface-cli
kfctl is specific to certain ways of deploying Kubeflow so its not always used.

Agreed. I can also commit this removal.

LMKWYT!

RFMVasconcelos:
Ideally, IMO, we should merge this subset of changes, to remove the outdated banner ASAP, and then work to agree on the best strategy in a new PR.

DONE:
- Remove <Installing command line tools> section from getting started
- Remove <Kubeflow command line interface (CLI)> section from overview

NEW PR:
- Find best way to display what is not in "Install Kubeflow" and "Configuration quick reference" section
- Improve or remove section "existing Kubernetes Cluster" based on 2.

@8bitmp3 @jlewi WDYT?

jlewi:
Thanks

The configuration quick reference is duplicating information in the sections above. The doc even says so

> The matrix is therefore an alternative way of accessing the information in the installation section above.

The problem is the column "manifest". Not all of the deployment options use KFDef manifests anymore.  The row for GCP is problematic because its pointing at the outdated manifest.

Here are a couple options

1. Delete the manifest column
1. Delete the table all together
1. Put Not Applicable in the GCP column

I much prefer deleting the manifest column altogether. This should lower the burden of maintaining this page over time. 

I might also suggest changing the title "Configuration Quick .. " to "Deployment Quick ..."

RFMVasconcelos:
Sounds good! 

I'll remove the manifest column and switch "Configuration" to "Deployment".

Bobgy:
Thank you both @RFMVasconcelos @jlewi!
/lgtm
/approve
This looks good to me!

I'm not seeing unanswered comments, so I'll just let it merge. Feel free to add further comments and start new PRs.

PR #2265, opened 2020-10-08
Update properties headers

subodh101:
What?
Update the header for authentication.

Why?
`kubeflow-userid` doesn't work with multi-tenancy feature.
The main problem is that when a contributor tries to access Notebook, Minio for visualization or any RBACed resource it throws `RBAC: access denied` which is why the issue is also related to the centralbashboard image version but at least the header should be correct.


Bobgy:
Thank you @subodh101!
The change you are suggesting is specific to GCP. Sounds reasonable adding a comment there saying if user uses GCP, they need to change the header name, but it's probably not suitable for everyone.

Bobgy:
Thank you! @subodh101 
/lgtm
/approve

PR #2257, opened 2020-10-04
Create a list of approvers and reviewers for Upgrading Kubeflow docs

8bitmp3:
Create a list of approvers and reviewers for Upgrading Kubeflow (deployment) docs - currently, there are none

@Bobgy Can you think of anyone else who works directly on the project and can approve the doc additions/changes here? Thank you!!

cc @RFMVasconcelos

PR #2275, opened 2020-10-10
Remove the Upgrading from Kubeflow v0.7.0 or later guide

8bitmp3:
The feature is no longer maintained, the docs are no longer relevant.

Related to the doc discussion here: https://github.com/kubeflow/website/pull/2257#issuecomment-705550851

As per @jlewi and @Bobgy comments:

> I think its sufficient for now to delete upgrading.md. We can leave it to the respective distribution owners to create their own upgrade documentation if and when they desire to.

> Agree to deleting upgrade.md doc, because the feature is no longer maintained.
Sign up for free

/assign @Bobgy @joeliedtke 

PTAL!

cc @RFMVasconcelos


Bobgy:
Thank you!
/lgtm
/approve

PR #2326, opened 2020-10-30
Update getting-started-multipass.md

didier-durand:
Hi there
The dot is not needed to do 'microk8s enable gpu'
Didier


RFMVasconcelos:
Hey @didier-durand! 

That's true, thank you for the help! 

/lgtm
/ok-to-test
/approve

PR #2361, opened 2020-11-18
Adding berndverst as Azure Docs approver

berndverst:


berndverst:
/lgtm

I am a reviewer of all of docs itself, so this probably should be ok :)

berndverst:
@aronchick can you also add the LGTM label? Apparently approval alone wasn't enough :)

Issue #2322, opened 2020-10-29
Kubeflow 1.2 release doc changes


Jeffwan:
I create this issue to track coming 1.2 doc changes.  WG owners and platform owners, it would be great to link your WIG or PRs there to help on tracking. 

We will cut a 1.2 branch from masters around Nov 8.

Docs for versioning the website: https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#version-the-website

Jeffwan:
@8bitmp3 Do you have capacity to drive 1.2 website release?

andreyvelich:
@Jeffwan From AutoML WG these PRs are related to 1.2:

- https://github.com/kubeflow/website/pull/2312
- https://github.com/kubeflow/website/pull/2315
- https://github.com/kubeflow/website/pull/2316
- https://github.com/kubeflow/website/pull/2320
- https://github.com/kubeflow/website/pull/2321
- https://github.com/kubeflow/website/pull/2324

That is the main issue to track the release: https://github.com/kubeflow/katib/issues/1360.

/cc @johnugeorge @gaocegege

RFMVasconcelos:
Thank you for creating this issue @Jeffwan! 

I can help drive this effort. Do you say you'll cut a 1.1 branch on Nov 8th?

RFMVasconcelos:
Hi @Jeffwan @andreyvelich !
 
So the way we do it is that we change `master` to be the latest version, now -> `v1.2`, and we create a new branch for the previous version - `v1.1`.

If you look, today we don't have a `v1.1-branch` on the website repo as `v1.1` is now in `master`

RFMVasconcelos:
Happy to drive the cut of `v1.1-branch` this week.

RFMVasconcelos:
@jlewi, it seems that I do not have the permissions to create a new branch upstream. How should I proceed?

jlewi:
@RFMVasconcelos 
https://github.com/kubeflow/community/blob/master/how-to/github_admin.md
https://github.com/kubeflow/internal-acls/blob/4535a60993946dcb57a05cd33552518030410ac5/github-orgs/kubeflow/org.yaml#L1100
https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization

RFMVasconcelos:
Thank you!

RFMVasconcelos:
[...]

PR #2365, opened 2020-11-23
Update website versioning for v1.2

RFMVasconcelos:
See #2322


RFMVasconcelos:
cc @8bitmp3 @jlewi @Jeffwan

Jeffwan (commented):
Seems we just need to make these changes. Rest should be taken care by cloud providers

Jeffwan:
`fatal error: concurrent map read and map write` @RFMVasconcelos Are you able to start Hugo server locally? This is the first time I see this issue

RFMVasconcelos:
Hm.. first time doing this as well, there seems to be an issue with Hugo - https://github.com/gohugoio/hugo/issues/7293

Jeffwan:
@RFMVasconcelos website tests are not managed by Prow,  `/retest` won't work. Could you make tiny change in your change like adding empty line at the end in `layouts/shortcodes/kf-latest-version.html . Let's see if that's a reproducible or intermittent issue

Bobgy:
I forced a redeploy, let's see how it goes

RFMVasconcelos:
Thank you @Bobgy! Looks like it fixed the issue 🚀

RFMVasconcelos:
Hi @Bobgy or @joeliedtke, could you please verify and approve?

PR #2383, opened 2020-12-01
Install kubectl using gcloud components

mnmami:
Without kubectl the step `make create-context` in the management cluster creation fails.


8bitmp3:
/assign @Bobgy @joeliedtke 
PTAL, thanks!

Bobgy:
Thanks for the fix!

PR #2372, opened 2020-11-26
docs(gcp): Kubeflow cluster upgrade instructions and revamp of env vars

Bobgy:
Part of https://github.com/kubeflow/gcp-blueprints/issues/139
Fixes https://github.com/kubeflow/gcp-blueprints/issues/142

Major changes:
- add kubeflow cluster upgrade instructions
- revamp env var ${KF_DIR} and ${MGMT_DIR} to a unified format

Please review related pages on https://deploy-preview-2372--competent-brattain-de2d6d.netlify.app/docs/gke/deploy/

PR #2385, opened 2020-12-01
Changing MANAGED-PROJECT to MANAGED_PROJECT

mnmami:
setting MANAGED-PROJECT using exports fails, e.g., "export: `MANAGED-PROJECT=kubeflow': not a valid identifier"


Bobgy:
Thanks, I've already added this change in https://github.com/kubeflow/website/pull/2372

PR #2390, opened 2020-12-03
fix(doc): update docs that still refer to KFP latest SDK reference. Fixes kubelow/pipelines#4685

rui5i:
Update docs that still refer to KFP latest SDK reference.

https://github.com/kubeflow/pipelines/issues/4685


Bobgy:
/lgtm
/approve
Awesome, thank you @rui5i !

PR #2410, opened 2020-12-14
Update install-kubeflow.md

aronchick:


8bitmp3:
@aronchick I can `/lgtm` if this is urgent. Otherwise, I can only suggest three formatting + stylistic changes.

joeliedtke:
/assign

@aronchick is this an open issue? If so, is there a bug for it? I would recommend adding a link to the bug to the note.

RFMVasconcelos:
This is losing relevance with Kubeflow v1.3 release and has a conflict. 

Should we close it? Or fix the conflict and merge?

RFMVasconcelos (commented):
As the community has deprecated `kfctl` and given that this change is no longer true, I think we can close this PR.

/close

PR #2420, opened 2020-12-20
Fix Component spec example

munendrasn:
* help should be description
* Add missing `optional` input argument


munendrasn:
@googlebot I signed it!

Bobgy:
/cc @chensun @numerology 
Can you take a look?
Thanks!

PR #2413, opened 2020-12-15
Update Azure Deployment Docs for KF 1.2

berndverst:
Updates the deploy instructions for KF 1.2 on Azure

Also addresses some config file confusion, #2398

PR #2423, opened 2020-12-21
Update istio documentation to branch 1.2 for Kubeflow v1.2

nanduri1998:
Documentation for installation via K8s needs update to v1.2 as all URLs point to 1.1

### DIFF
--- a/layouts/shortcodes/config-file-istio-dex.html
+++ b/layouts/shortcodes/config-file-istio-dex.html
@@ -1,1 +1,1 @@
-kfctl_istio_dex.v1.1.0.yaml
+kfctl_istio_dex.v1.2.0.yaml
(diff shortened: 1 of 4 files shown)


berndverst:
FYI I had to do the same to the Azure docs (which use different shortcodes). I only caught that after the docs release for v1.2 also.
See: https://github.com/kubeflow/website/pull/2413/files

Bobgy:
/approve
Thanks!

As a follow up, can you move shortcodes into a sub-folder and add OWNERS there, so kfctl_istio_dex maintainers can approve this by themselves?

RFMVasconcelos:
@Bobgy that makes a lot of sense! I'll make an issue on this so we can get OWNERS for shortcodes ;)

PR #2428, opened 2020-12-29
Improve update instruction steps

jonasdebeukelaer:
* Two additional env variables are required for kf cluster setup
* clearer env var reference for management cluster update

Side note: is there a way to test the update was successful? aside from no errors in console?

### DIFF
--- a/content/en/docs/gke/deploy/upgrade.md
+++ b/content/en/docs/gke/deploy/upgrade.md
@@ -15,7 +15,7 @@ This guide assumes the following settings:
 
 * The `${MGMT_DIR}` and `${MGMT_NAME}` environment variables
   are the same as in [Management cluster setup](../management-setup#environment-variables).
-* The `${KF_DIR}` and `${KF_NAME}` environment variables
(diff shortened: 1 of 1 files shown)


jonasdebeukelaer:
Also it would be safer to require the user to specify the managed cluster's namespace name in the management cluster.
 
I have two clusters managed, and the default context in the management cluster points to one of the two kf clusters. When trying to update the second kf cluster, they were clashing. Hopefully it's not broken anything, i need to investigate.

8bitmp3 (commented):
/assign @Bobgy

And HNY to y'all! 🥂

Bobgy:
/lgtm
/approve
thanks for the fix!

Bobgy:
> Also it would be safer to require the user to specify the managed cluster's namespace name in the management cluster.
> 
> I have two clusters managed, and the default context in the management cluster points to one of the two kf clusters. When trying to update the second kf cluster, they were clashing. Hopefully it's not broken anything, i need to investigate.

good point, mind opening a separate issue?

jonasdebeukelaer:
@Bobgy sure np

PR #2439, opened 2021-01-07
Fix KF_DIR env. var. description

mnmami:
Replace 'management' with 'kubeflow' to describe KF_DIR env. variable, as correctly described just above there.


susan-shu-c:
Haha what, did we seriously submit the same fix in like 1 minute within each other #2440

Nice coincidence :)

mnmami:
Haha, this happens only once each 5 million years :)

8bitmp3:
/assign @Bobgy 

I assume this is good

/lgtm

Bobgy:
Thanks for the fix!
/approve

PR #2449, opened 2021-01-12
Fix url

susan-shu-c:
Without the slash, the link actually doesn't work (it misunderstands the relative link to start from the current page, not the "root")


joeliedtke:
/lgtm
/approve

Thanks for fixing this link!

PR #2451, opened 2021-01-13
Add newer k8s versions, cert-manager incompatibility <1.16

DavidSpek:
This PR adds Kubernetes 1.19 and 1.20 to the table. This PR does create 2 new issues:

1. The recommended version of Kubernetes is still stated as 1.14, even though the bundled cert-manager is not compatible and it is old. 
2. What is the process to verify 1.17 to 1.20 so that they can be listed as compatible?

/cc @yanniszark

### DIFF
--- a/content/en/docs/started/k8s/overview.md
+++ b/content/en/docs/started/k8s/overview.md
@@ -44,7 +44,8 @@ Your Kubernetes cluster must meet the following minimum requirements:
         <th>Kubeflow 0.6</th>
         <th>Kubeflow 0.7</th>
         <th>Kubeflow 1.0</th>
-        <th>Kubeflow 1.1</th>        
+        <th>Kubeflow 1.1</th>
+        <th>Kubeflow 1.2</th>         
       </tr>
     </thead>
     <tbody>
@@ -52,6 +53,7 @@ Your Kubernetes cluster must meet the following minimum requirements:
         <td>1.11</td>
         <td><b>compatible</b></td>
         <td><b>compatible</b></td>
+        <td>incompatible</td>
         <td>incompatible</td>
         <td>incompatible</td>
         <td>incompatible</td>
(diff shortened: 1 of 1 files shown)


8bitmp3:
/assign @swiftdiaries @yanniszark

cc @RFMVasconcelos @Bobgy

yanniszark:
@DavidSpek I remember there was an issue where I commented on this:

> even though the bundled cert-manager is not compatible and it is old

The bundled cert-manager is old which is why it's compatible with older K8s releases.
In this comment (https://github.com/kubeflow/kubeflow/issues/5501#issuecomment-755675531) you point to cert-manager docs. But your link points to the latest version, not the one we have installed. These is the link for the installed version:
https://docs.cert-manager.io/en/release-0.11/

Where it states (https://docs.cert-manager.io/en/release-0.11/getting-started/install/kubernetes.html):

> From cert-manager v0.11.0 onwards, the minimum supported version of Kubernetes is v1.11.0. Users still running Kubernetes > v1.10 or below should upgrade to a supported version before installing cert-manager.


We (Arrikto) have also successfully installed cert-manager both on K8s 1.14 and 1.16.

Have I missed something? It seems to me that there is no cert-manager incompatibility with versions < 1.16.

DavidSpek:
@yanniszark Indeed, I stand corrected. However, if I remember correctly this issue did happen to a user. Either way, I think it is probably a good idea to use a more up to date version of cert-manager. Along with that, I am still interested how newer versions of kubernetes are supposed to be validated. When can a new version be added to the list as having no known problem (what is the threshold for usage and testing), and what is the threshold for saying a version is compatible? Personally, I've run kubernetes 1.19 for a few months and have been running 1.20 since it's release and have not experienced any issues (though my usage of Kubeflow is quite limited). The only versions of Kubernetes that are listed as being compatible are all from 2019, and this is a problem in my view. Especially as it is causing users to choose to run a a much older version of Kubernetes than likely is necessary, just because the docs say those version are compatible.

joeliedtke:
/ok-to-test

Preview: https://deploy-preview-2451--competent-brattain-de2d6d.netlify.app/docs/started/k8s/overview/

DavidSpek:
@joeliedtke 

I’ll remove the very-manager comment as that is a mistake.

DavidSpek:
/ok-to-test
@joeliedtke I think it is good to merge now.

Bobgy:
/lgtm
/approve

Thank you @DavidSpek

Issue #2489, opened 2021-02-05
Adding "Methods and Distributions" to aggregate 


RFMVasconcelos:

RFMVasconcelos:
Given the community decision in #2492, changing this issue to `Distributions`

thesuperzapper:
@RFMVasconcelos hey, it seems like only some distributions (AWS, Operator) were moved with PR https://github.com/kubeflow/website/pull/2492, what are the plans to move the others under "Distributions" (Azure, GCP, IBM Cloud)?

RFMVasconcelos:
@thesuperzapper there is one PR per distribution - #2547 #2548 #2549 #2550. I've left some time for these to be viewed by all the leads. I believe we'll be able to merge them all tomorrow.

thesuperzapper:
I created some issues about moving a few stragglers to the new "Distributions" folder:
- https://github.com/kubeflow/website/issues/2565
- https://github.com/kubeflow/website/issues/2567

thesuperzapper:
@RFMVasconcelos I created some issues about moving a few stragglers to the new "Distributions" folder:

#2565
#2567

RFMVasconcelos:
Given that only #2565 is missing, let's track it there. 

/close

PR #2492, opened 2021-02-08
Create "Distributions" with kfctl + Kubeflow Operator

RFMVasconcelos:
PR to tackle issue #2489, as part of the Kubeflow docs restructuring [project](https://github.com/kubeflow/website/projects/2).


RFMVasconcelos:
Hey @Bobgy could you please take a look into this PR? 

This creates the concept of "Methods & Distributions" as discussed in the last community meetings, to clarify what docs belong to what methods or distro. 

/assign @Bobgy

RFMVasconcelos:
Hey @8bitmp3! Could you please take a look into this PR?

This creates the concept of "Methods & Distributions" as discussed in the last community meetings, to clarify what docs belong to what methods or distro.

content/en/docs/methods/_index.md
```diff
@@ -0,0 +1,5 @@
++++
+title = "Methods & Distributions"
+description = "A list available methods for Kubeflow deployment and operations"
```
8bitmp3, line 3: Hey @RFMVasconcelos : WDYT

```suggestion
description = "A list of available methods for Kubeflow deployment and operations"
```

content/en/docs/methods/kfctl/_index.md
```diff
@@ -0,0 +1,5 @@
++++
+title = "kfctl"
+description = "Documentation of how to use the kfctl method for Kubeflow"
```
8bitmp3, line 3: hey @RFMVasconcelos, since it's all docs here under `/docs`, maybe there's no need for saying "Documentation":

```suggestion
description = "How to use the kfctl method for Kubeflow"
```

8bitmp3:
@RFMVasconcelos Catching up with PRs 🙏 See my comments above, LMKWYT. If you're happy with the links, I'm good and will LGTM

RFMVasconcelos:
Hey @8bitmp3 ! 

Thank you for the review! Let me take a look :)

RFMVasconcelos:
Nice catches! Accepted both suggestions :)

8bitmp3:
/lgtm

/assign @animeshsingh @Bobgy @joeliedtke
(please `/approve` if 👍, thanks)

content/en/_redirects
```diff
@@ -91,7 +91,7 @@ docs/started/requirements/                    /docs/started/getting-started/
 # Restructured the getting-started and other-guides sections.
 /docs/started/getting-started-k8s/             /docs/started/k8s/
 /docs/started/getting-started-minikf/                      /docs/started/workstation/getting-started-minikf/
-/docs/started/getting-started-minikube/                    /docs/started/workstation/minikube-linux/
+/docs/started/getting-started-minikube/                    /docs/started/methods/kfctl/minikube/
```
Bobgy, line 94: I believe for each file move, you need to set up a new redirection rule, because there can be links around the internet pointing to these docs.

RFMVasconcelos: Hey @Bobgy! That's a very good point! Just created #2511 to tackle this

joeliedtke:
My understanding was that this change was still in discussion in [kubeflow/community #440](https://github.com/kubeflow/community/pull/440). Would it be alright for us to put this on hold until that discussion is complete?

content/en/_redirects
joeliedtke: This issue should not be addressed as a separate issue. Each move must contain the relevant redirects to keep from breaking incoming links and on-site links.

RFMVasconcelos: That is a good point! Let's merge first the redirects change and then this one.

I will keep them separate so it's easier to review, this one is starting to be big. :)

RFMVasconcelos: [Forget prior msg] 

I think it's easier to do as you propose. Include redirects on each change, so it's done in bite-sized chunks

RFMVasconcelos: Added redirects in these commits - [operator](https://github.com/kubeflow/website/pull/2492/commits/54e2b002e4b340fe7a34e7790c43ef68d6c3af47), [kfctl](https://github.com/kubeflow/website/pull/2492/commits/fedabc8f93e1a5f471ac472c81b1b4390e16eb1d)

animeshsingh:
thanks @RFMVasconcelos 
/lgtm

RFMVasconcelos:
@Bobgy could you please review and verify if I've been able to fix all issues so we can approve?

content/en/docs/methods/_index.md
```diff
@@ -0,0 +1,5 @@
++++
+title = "Methods & Distributions"
```
Bobgy, line 2: In the meetings, if I recall properly, people prefer `Distributions` more than `Methods`. I'm OK with naming the section "Methods & Distributions" for now, but I'd prefer at least change the path to `/docs/distributions/xxx`.

Bobgy (commented):
My current thoughts, I'll sync with @joeliedtke tomorrow.

content/en/docs/methods/_index.md
RFMVasconcelos: Hey @Bobgy ! Thank you for the comment, I will bring this up for discussion in the next meeting, I am happy with either naming convention. :) 

My understanding was that `kfctl` would be more of a method and e.g. `minikf` more regarded as a distribution, hence the 2 names, but I agree that may not be too stylish and that 1 name would be better than 2.

joeliedtke: "Methods" is too generic in this context. The term _method_ doesn't imply that it is related to the installation process, just that it is related to _a_ process.

If  "Distritbutions" does not cover all the installation cases, then I would recommend something like "Installing Kubeflow" or "Deploying Kubeflow". The "Installing|Deploying Kubeflow" section could have sub-sections for distributions and methods.

joeliedtke:
/hold

Adding a hold while we discuss the section name.

joeliedtke:
I just had a quick chat with @Bobgy , here are the open items from our discussion:

1. Not all distributions are currently included in the planned "Methods and distributions" section. For example, the Google Cloud distribution is documented in the platform docs. Users who start with the Distributions docs will not know about these other distributions. 
2. There seems to be overlap between the Platforms and Distributions sections. For example, the Platforms section seems to largely be about installing and configuring a Kubeflow distribution on a cloud platform. Why not consolidate the platforms content under distributions?
3. The _Methods_ in "Methods and distributions" is a generic term that could be confusing. If this section is about installation, then consider renaming to "Installing Kubeflow" with subsections for _methods_ and _distributions_. Otherwise, if this is primarily about distributions and there are different methods used by the distributions, then I think that "Distributions" is the primary concept and the better section name.

Also, @Bobgy  suggested that we create the v1.2 docs branch so we can start work on the v1.3 docs without affecting the users.

Bobgy:
Two options I can think of to tackle problems @joeliedtke has mentioned. Shall we do a quick poll here using thumb ups?

1. Build a "installing kubeflow" page, which should be source of truth for finding platforms and distributions, similar to https://www.kubeflow.org/docs/started/getting-started/. We should avoid letting users find distributions by browsing the distributions section (because platforms is a separate section).

If you agree with this, thumbs up on this comment.

Bobgy:
2. Merge platforms and distributions into a single section called "Distributions", which is what I prefer too, a platform is also an opinionated distribution. I believe there were more people agreeing to this during the community meeting.

Please add a thumbs up here if this is what you prefer.

content/en/docs/methods/_index.md
RFMVasconcelos: @joeliedtke thank you for your comment! You're probably right that "Methods" alone might be too ambiguous.

RFMVasconcelos:
Thank you for the clarity @Bobgy and @joeliedtke. 

I brought this up in the community meeting and asked that members vote in your Poll or comment on this PR. 

I see your point of "If there are platforms and distributions, which one do I choose as a user?". Given this, I voted on the second option you propose.

RFMVasconcelos:
It seems that we mostly have consensus for merging platforms into distributions, let's give it a few more days for people to express if they think otherwise.

RFMVasconcelos:
Great! 

To tackle this change in bite-sized, reviewable chunks I'll modify this PR including the `kfctl` and `Kubeflow operator` under `Distributions` and make additional PRs to move each of the other `Distributions` under this same umbrella.

RFMVasconcelos:
@8bitmp3 I had to rebase this PR and since you co-authored a couple of commits, we need your consent :)

thesuperzapper:
@8bitmp3 can you please write `@googlebot I consent.`

Otherwise, @RFMVasconcelos you can just squash out the commits by @8bitmp3

RFMVasconcelos:
@thesuperzapper @8bitmp3 I squashed commits so no action required. 

I think we're good to review and merge these. @joeliedtke @Bobgy would you be keen to take a look? :)

Bobgy (commented):
/lgtm

Thank you @RFMVasconcelos !

/assign @joeliedtke 
to make the final decision

joeliedtke:
Let's update the Netlify configuration to use the 1.2 docs branch (we may need to create this branch) so these changes can be made without affecting the published docs until the 1.3 release. Once that change is in place, I will remove the hold and approve this PR.

Bobgy:
I've set up v1.2-branch in https://github.com/kubeflow/website/issues/2546.
/unhold

@RFMVasconcelos can you resolve the conflicts in this PR?

RFMVasconcelos:
I've resolved the conflict @Bobgy, need a new LGTM now as I made changes.

PR #2509, opened 2021-02-18
doc(kfp): remove outdated banner for generated reference

Bobgy:
/assign @joeliedtke


zijianjoy (approved):
Thank you Yuan! Besides this page, do we also need to remove the outdated banner in https://github.com/kubeflow/website/blame/master/content/en/docs/pipelines/pipelines-quickstart.md#L7-L10?

PR #2518, opened 2021-02-24
Update IBM storage config

Tomcli:
The IKS with classic VM is getting group ID support for the file storage on Kubernetes 1.17+. Since file storage will continue to be the default for classic VM, we will change the recommended classic VM storage to `ibmc-file-gold-gid` to minimize maintenance support.

For OpenShift and other types of VM, block storage is the default and we don't need any extra instructions to configure it.

### DIFF
--- a/content/en/docs/ibm/create-cluster.md
+++ b/content/en/docs/ibm/create-cluster.md
@@ -56,7 +56,7 @@ Follow these steps to create and setup a new [IBM Cloud Kubernetes Service(IKS) 
 Choose the region and the worker node provider for your cluster, and set the environment variables.
 
 ```shell
-export KUBERNERTES_VERSION=1.16
+export KUBERNERTES_VERSION=1.17
 export CLUSTER_ZONE=dal13
 export WORKER_NODE_PROVIDER=classic
 export CLUSTER_NAME=kubeflow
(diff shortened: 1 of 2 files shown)


pvaneck (commented):
/lgtm

ScrapCodes:
I have tested this out for classic infrastructure. This storage class has support for ReadWriteMany.
In VPC infrastructure, there is no such equivalent - available.

Tomcli:
@ScrapCodes this document is focus on classic vm cluster. Do you mind creating a new page for VPC gen 2 cluster? We can start with adding instructions to create-cluster.md on how to create a VPC cluster with proper network setup.

moficodes:
Seems like vpc only supports a ReadWriteOnce volume. Would kubeflow components work properly in such env?

Tomcli:
> Seems like vpc only supports a ReadWriteOnce volume. Would kubeflow components work properly in such env?

ReadWriteMany only matters when running distributed workload on one volume. It mostly matters for training operators like tfjob where the advantage is to leverage resources in multiple nodes.

ScrapCodes:
@Tomcli, yeah, I am on it.

moficodes (approved):
/lgtm

animeshsingh:
@ScrapCodes if distributed training may not work with ReadWriteOnce - do we need to add instructions for that?

animeshsingh (approved):

PR #2523, opened 2021-03-02
Azure: Update single-user deployment instructions

berndverst:
The Azure stack does not add functionality needed for deployment to Azure in the single user use case. To avoid duplication in documentation and to ensure faster updates we will refer to the standard / vanilla kubernetes deployment files for the single user use case.


yilun-msft:
A meta comment: on the top of Azure doc, it says install Docker is optional. Let's remove `(optional)` to avoid confusion.

PR #2527, opened 2021-03-04
docs(gke): Delete submit-docker-image.md

Bobgy:
/assign @DavidSpek 

After taking another look, I think this is no longer needed.

Auth in GKE will happen transparently after workload identity became the default, so there's no need for a page to introduce that.


DavidSpek:
@Bobgy Sounds good! I have some plans for documentation and potentially a web-app for users to create custom images in the cluster using Kaniko.

PR #2542, opened 2021-03-16
Limit Kustomize version to 3.x.x

benjaminwinokur:
Kustomize version 4.x is incompatible with the current syntax.

### DIFF
--- a/content/en/docs/distributions/gke/deploy/deploy-cli.md
+++ b/content/en/docs/distributions/gke/deploy/deploy-cli.md
@@ -78,13 +78,14 @@ purpose. No tools will assume they actually exists in your terminal environment.
 
 1. Install [Kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/).
 
-    **Note:** Prior to Kubeflow v1.2, Kubeflow was compatible only with Kustomize `v3.2.1`. Starting from Kubeflow v1.2, you can now use the latest Kustomize versions to install Kubeflow.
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@benjaminwinokur can you please sign the google CLA, and rebase (we have moved all the distributions under the `content/en/docs/distributions/gke` folder now)

benjaminwinokur:
@googlebot I signed it!

yanniszark:
@benjaminwinokur I would wait for the latest instructions from the manifests working group (https://github.com/kubeflow/manifests/issues/1776) so we can announce the compatible versions correctly. The current version we use is `v3.2.0`.

Furthermore, is there a reason you found for this:

> Kustomize version 4.x is incompatible with the current syntax.

benjaminwinokur:
`kustomize build --load_restrictor none -o $(BUILD_DIR)/gcp_config $(GCP_CONFIG)`

--load_restrictor is no longer supported in kustomize v4

yanniszark:
> --load_restrictor is no longer supported in kustomize v4

I think the flag has been renamed. Can you take another look?

benjaminwinokur:
Ah you are correct. Thanks!
```
{Version:kustomize/v4.0.5 GitCommit:9e8e7a7fe99ec9fbf801463e8607928322fc5245 BuildDate:2021-03-08T20:53:03Z GoOs:linux GoArch:amd64}
kustomize build --help
      --load-restrictor string   if set to 'LoadRestrictionsNone', local kustomizations may load files from outside their root. This does, however, break the relocatability of the kustomization. (default "LoadRestrictionsRootOnly")
```

Do you want to wait on the results of (kubeflow/manifests#1776) or I can update this PR to fix the syntax in the Makefiles? This 2nd option will require the user to use kustomize v4.

PR #2543, opened 2021-03-16
Add Tommy to the IBM doc approvers

Tomcli:


8bitmp3:
@animeshsingh Currently, there's only one approving `OWNERS` member. If we add @Tomcli , we can have two and you'll have a backup! 👍 👍 😃

PR #2544, opened 2021-03-16
KFP - Update Python function-based component doc with param naming rules

joeliedtke:
When the KFP SDK converts a Python function into a component, there are a few cases where the argument name is not the same as the input/output parameter name. These updates describe the rules that are applied when the function is converted to a component.

Also, there is a small change to specify that users can pass `list` and `dict` instances by value.


joeliedtke:
The notebook and markdown files contain the same content. (The markdown file is generated using scripts/nb_to_md.py.) You need to review only one of the files. I'll apply all changes to the notebook file, and regenerate the markdown file with the updates applied.

Preview: https://deploy-preview-2544--competent-brattain-de2d6d.netlify.app/docs/components/pipelines/sdk/python-function-components/

chensun:
/lgtm 

Thanks!

content/en/docs/components/pipelines/sdk/python-function-components.ipynb
```diff
@@ -287,14 +287,55 @@
...
     "<a name=\"pass-by-value\"></a>\n",
     "#### Passing parameters by value\n",
     "\n",
     "Python function-based components make it easier to pass parameters between\n",
     "components by value (such as numbers, booleans, and short strings), by letting\n",
     "you define your component’s interface by annotating your Python function. The\n",
-    "supported types are `int`, `float`, `bool`, and `string`. If you do not\n",
-    "annotate your function, these input parameters are passed as strings.\n",
+    "supported types are `int`, `float`, `bool`, and `str`. You can also pass \n",
+    "`list` or `dict` instances by value, if they contain small values, such as\n",
```
Ark-kun, line 336: >if they contain small values, such as `int`, `float`, `bool`, or `str` values.

Small correction: Lists and dictionaries can contain other lists and dictionaries. Basically they can contain anything that can be serialized/deserialized with `json.dumps`/`json.loads`.

Long version:
This is almost, but not 100% correct.
There are two core requirements:
* The total size of arguments consumed by value (inputValue) in the pipeline should not be big (<<1MiB per pipeline run).
* The constant arguments passed to a component needs to be serializable to string.
* The serialized string cannot contain NULL-bytes (`\x00`)

There are built-in serializers for the following types: `str` (`"String"`), `int` (`"Integer"`), `float` (`"Float"`), `bool`(`"Bool"`, `"Boolean"`), `list` (`List`, `"JsonArray"`), `dict` (`Dict`, `"JsonObject"`).

Here is what follows:
* Function parameters can be annotated with any python types or type names.
* By default, the function will receive the value of each parameter as a string.
* However if the function parameter is annotated with a known supported type (e.g. `int`, `float`, `bool`, `list`, `dict`), the parameter value will be automatically deserialized and the function will receive the value with a proper type (e.g. `int`).
* When returning the outputs by value, the returned object needs to be serialized to a string (if it's not already a string). Most values are serialized using the `str(value)` function. However for `list` and `dict`-typed outputs the `json.dumps(value)` serialization is used (if the output value is not a string already). `json.dumps` has certain limitations and quirks. It can only serialize a limited number of types. Also JSON considers all dictionary keys to be strings.

Some relevant code:
https://github.com/kubeflow/pipelines/blob/a0fdb92413c4dd97ff7be554b4df680907d2c56d/sdk/python/kfp/components/_data_passing.py#L108..L117

PR #2550, opened 2021-03-17
Move `Openshift` platform under /distributions

RFMVasconcelos:
Following the latest discussions on platforms and distributions starting a PR to move Kubeflow on Openshift.

cc @joeliedtke @Bobgy @animeshsingh @8bitmp3


8bitmp3:
cc @vpavlin @crobby @pdmack @animeshsingh

RFMVasconcelos:
@animeshsingh could you please review this? I have made a similar request to all the platforms so we get them moved to distributions.

This is merging to master and we can make sure we only point the main website to master once all the platforms have moved, so there is no weird scenario where one of the clouds would be outside and others inside the distributions folder.

/assign @animeshsingh

cc @Bobgy @joeliedtke

Bobgy:
Also
/cc @Tomcli 
Who would be the best POC going forward?

PR #2562, opened 2021-03-23
Make kfctl first distribution

RFMVasconcelos:
Looking at the current distributions list, I believe that kfctl should always be the first on the list, as it is the community endorsed method.


thesuperzapper:
@RFMVasconcelos makes sense, but we will need to re-write those docs as kfctl wont be required for Kubeflow 1.3, only kustomize and git-cloning the `kubeflow/manifets` repo.

/lgtm

RFMVasconcelos:
I see @thesuperzapper! Thank you for merging this for now. Let's look into rewriting the docs then :)

PR #2563, opened 2021-03-24
cherry pick master into v1.2-branch

thesuperzapper:
This PR cherry picks all commits from `master` into `v1.2-branch`, as otherwise the "distribution" changes are not present, which makes the sections on the left hand side messy.


Bobgy:
I believe it was intentional that we keep the restructuring changes not in 1.2 to avoid unknown issues.

thesuperzapper:
@Bobgy makes sense

I think I was getting confused by the drop down for versions, (which is currently missing v1.2)
Screen Shot 2021-03-24 at 12 48 49

At very least we should ensure that https://www.kubeflow.org/docs/ lets you select v1.2, and master separately.

Issue #2567, opened 2021-03-24
Move "Kubernetes Installation" under "Distributions"


thesuperzapper:
Currently there is a confusingly named `Kubernetes Installation` section under `Getting Started`, which contains ​information about the "Vanilla Istio-Dex" distribution.

In the interest of not treaing it any differnt to the other distributions, it should be moved under "Distributions / kfctl".

## propsoed structure is 
```
Distributions:
- kftcl
   - Overview [moved from "Getting Started / Kubernetes Installation / Overview of Deployment on an Existing Kubernetes"] Cluster
   - ... existing stuff...

- ...
```

thesuperzapper:
What are your thoughts @RFMVasconcelos @Bobgy?

RFMVasconcelos:
@thesuperzapper I fully agree with this, it is an overview of `kfctl`. 

It includes a compatibility table which we might want to move to a more central place.

RFMVasconcelos:
Solved in #2569

PR #2569, opened 2021-03-24
Move getting started on K8s page to under kfctl distribution

RFMVasconcelos:
This solves #2567


RFMVasconcelos:
cc @thesuperzapper @Bobgy @joeliedtke @animeshsingh

thesuperzapper:
Looks good, I don't think there is any controversy here.

/lgtm

Issue #2546, opened 2021-03-17
Kubeflow 1.3 release doc updates


Bobgy:
Creating an issue to track updates preparing for Kubeflow 1.3 doc release.

Bobgy:
I'm following https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#lifecycle.

I just forked current master branch as `v1.2-branch`, and pinned www.kubeflow.org to `v1.2-branch`.
It's OK to move forward with large changes preparing for the next release now.

RFMVasconcelos:
Thank you for taking on this work @Bobgy !

RFMVasconcelos:
I have just created #2547 #2548 #2549 #2550 #2551 to move all `platforms` under `distributions` 

cc @Bobgy @joeliedtke @animeshsingh @8bitmp3 @yanniszark

thesuperzapper:
@RFMVasconcelos before we close this, I think we should make https://www.kubeflow.org/ default to master branch, as there is nothing 1.3 specific in master, (and it is much better structured).

Also we should resolve the issue about not having `v1.2` in the dropdown (https://github.com/kubeflow/website/issues/2564)

RFMVasconcelos:
@thesuperzapper this is part of the [docs release process](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#version-the-website). I will follow these steps to have Kubeflow.org point to master and there will be v1.2 in the selector.

This is a state in transition. When we point it to `master` (by EOW latest) this will be resolved.

RFMVasconcelos:
In fact, trying to do this now I realize this is [done on Netlify](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#changing-the-version-wwwkubefloworg-points-to) which I have not played with yet, I am not sure this is googlers only. I think @Bobgy did this for us in the previous release.

I think the best thing to do is for me to get the `master branch` ready to be pointed to, and @Bobgy, to point kubeflow.org to `master` once he feels comfortable with the status.

RFMVasconcelos:
Sounds good @Bobgy ?

thesuperzapper:
I think we can try and resolve https://github.com/kubeflow/website/issues/2536 before doing the master switch.

If your willing to make a PR @RFMVasconcelos?

Bobgy:
Sure, for any netlify related steps. You can ping @joeliedtke or me about them.

[...]

PR #2604, opened 2021-04-13
Remove deprecated kfctl pages

castrojo:
These techniques are now deprecated, starting removal.

/cc @RFMVasconcelos


thesuperzapper:
Lets merge this after @Bobgy has cherry-picked/recreated the `v1.2-branch` from master (as this change is specific to 1.3), as discussed here https://github.com/kubeflow/website/issues/2546#issuecomment-819151434

RFMVasconcelos:
+1 @thesuperzapper

RFMVasconcelos (commented):
This PR LGTM, but as @thesuperzapper mentions, we need to make sure it only becomes effective for v1.3 docs, preserving the `kfctl` documentation for v1.2 users.

thesuperzapper:
@castrojo there are other references to `kfctl` which should be removed, for example under:
* lots of stuff under "Further Setup and Troubleshooting"
* surprisingly, the katib component docs
* most of the distribution docs (but hopefully those distributions will update those!)

castrojo:
Ok I've removed some of the non-distribution mentions, however I am not sure how that affects the instructions around the removal -- I need an engineer to check to make sure I didn't cut too much or if what I cut doesn't break the world.

shannonbradshaw:
@RFMVasconcelos I'm wondering where we stand with this one.

RFMVasconcelos:
@shannonbradshaw I think that this PR is in a good shape, but first we need to get #2612 approved as the highest priority, then make the website point from `v1.2-branch` to `master` (google needs to do this on netlify). 

Once these 2 steps are done, we're good to merge this PR.

Bobgy:
Does this also mean deployment WG is should be turned down?

content/en/docs/components/katib/experiment.md
```diff
@@ -754,11 +754,7 @@ kubectl apply -f <your-path/your-experiment-config.yaml>
 
 - (Optional) Katib's experiments don't work with
   [Istio sidecar injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection).
-  If you install Kubeflow using
-  [Istio config](https://www.kubeflow.org/docs/methods/kfctl/deployment),
-  you have to disable sidecar injection. To do that, specify this annotation:
-  `sidecar.istio.io/inject: "false"` in your experiment's trial template. For
-  examples on how to do it for `Job`, `TFJob` (TensorFlow) or
```
andreyvelich, line 761: We should not delete this information since the default installation is also used Istio: https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml#L9.
Maybe we can just rephrase it ? For instance: 
"If you are running Kubeflow with Istio config, you have to disable sidecar injection."

content/en/docs/components/katib/hyperparameter.md
```diff
@@ -141,9 +141,6 @@ an experiment using the random algorithm example:
 1. (Optional) **Note:** Katib's experiments don't work with
    [Istio sidecar injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection).
    If you installed Kubeflow using
-   [Istio config](/docs/methods/kfctl/deployment),
-   you have to disable sidecar injection. To do that, specify this annotation:
-   `sidecar.istio.io/inject: "false"` in your experiment's trial template.
```
andreyvelich, line 146: Similar comment here.

content/en/docs/components/katib/hyperparameter.md
```diff
@@ -393,10 +390,6 @@ the Kubeflow's TensorFlow training job operator, TFJob:
 
 1. (Optional) **Note:** Katib's experiments don't work with
    [Istio sidecar injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection).
-   If you installed Kubeflow using
-   [Istio config](/docs/methods/kfctl/deployment),
-   you have to disable sidecar injection. To do that, specify this annotation:
-   `sidecar.istio.io/inject: "false"` in your experiment's trial template.
```
andreyvelich, line 399: Similar comment here.

content/en/docs/components/katib/hyperparameter.md
```diff
@@ -437,10 +430,6 @@ using Kubeflow's PyTorch training job operator, PyTorchJob:
...
```
andreyvelich, line 443: Similar comment here.

andreyvelich (commented):
Thank you for taking this @castrojo!
I left few comments.
cc @gaocegege @johnugeorge

cvenets:
> Does this also mean deployment WG should be turned down?

@Bobgy I think that this would make sense to discuss after the release, yes. It seems that almost all distributions are moving to pure kustomize and their own tooling on top starting 1.3, and no one seems to have a strong appetite to maintain `kfctl` any more. We can ask in kubeflow-discuss once the release is out, and if no one steps up to maintain `kfctl`, then there doesn't seem to be a need for the Deployments WG to exist.

RFMVasconcelos (commented):
We are now in the [condition](https://github.com/kubeflow/website/pull/2604#issuecomment-819153806) we set for merging of this PR. 

It looks good to me! Thank you for contributing this @castrojo :tada: 

/lgtm
/approve

PR #2627, opened 2021-04-20
docs(kfp): bump version to 1.5.0

Bobgy:


Bobgy:
Self LGTMing, because this is part of release

PR #2626, opened 2021-04-19
Distinguish components from external add-ons

shannonbradshaw:
This fixes #2536 I'm expecting to make small fixes depending on reviews.

PR #2630, opened 2021-04-21
Components v external add ons

### LINKED ISSUE #2536: Split "Components" section into "Core Components" and "External Components"
While it's awesome we now have all components living under one section, this has caused users to become confused about what is actually part of Kubeflow, and what is external.

### My proposal is to split "Components" into:
* `Core Components` (all components owned by a Kubeflow WG)
* `External Components` (all other components we want to highlight)

### Proposed Structure:
```
Core Components:
- Central Dashboard
   - ... same ...
   - + Multi-Tenancy [moved]
- Notebooks [rename from "Notebook Servers"]
   - ... same ...
- Pipelines [rename from "Kubeflow Pipelines"]
   - ... same ...
- Metadata
   - ... same ...
- Katib
   - ... same ...
- KFServing
   - ... same ...
- Training Operators [rename from "Frameworks for Training"]
   - ... same ...
- Fairing
   - ... same ...
```
```
External Components:
- Istio
   - ... same as before ...
- Tools for Serving
   - ... same as before ...
- Feature Store
   - ... same as before ...
```

shannonbradshaw:
This is fixes #2536. It is a replacement for PR: #2626 (which I will now close).


RFMVasconcelos (commented):
@shannonbradshaw this is great work! Thank you for your contribution :) 

Given that the community has agreed on this step already in #2536, and that we are releasing v1.3 providing urgency, I will approve this. cc @thesuperzapper @castrojo

/lgtm
/approve

PR #2655, opened 2021-04-26
change title of page and video to Introduction to Kubeflow

### LINKED ISSUE #2618: Google search results show "An error occurred" as the title for about/kubeflow
See this screenshot:
Screen Shot 2021-04-15 at 9 10 04 PM

Potentially relevant:
https://stackoverflow.com/questions/38943067/snippet-in-google-search-contains-an-error-occurred-try-watching-this-video-on

shannonbradshaw:
This is an effort to fix #2618  The only place the expression "An error occurred." occurs on the page is inside the <noscript> tag for the embedded video. 

My theory is that either:
1. Google is attempting to find a better title for this page than a title that is exactly the same as the domain name (the current title) and ended up using the `<h1>An error occurred.</h1>` inside the <noscript> tag. 
2. There was a transient error when Google crawled the page.

In either case, I think these changes slightly improve the page and should encourage Google to reindex the page.

We should also directly request a re-index from the Google Search Console.

cc: @RFMVasconcelos

### DIFF
--- a/content/en/docs/about/kubeflow.md
+++ b/content/en/docs/about/kubeflow.md
@@ -23,7 +23,7 @@ your environment and install Kubeflow.
 
 Watch the following video which provides an introduction to Kubeflow.
 
-{{< youtube cTZArDgbIWw title="Introduction to Kubeflow">}}
+{{< youtube id="cTZArDgbIWw" title="Introduction to Kubeflow">}}
 
 ## What is Kubeflow?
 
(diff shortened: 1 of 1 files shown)


Bobgy:
/ok-to-test
Thank you! That seems a good guess!

shannonbradshaw:
@RFMVasconcelos note @Bobgy's comment above. This is pretty ugly in search results. Would like to move this forward if we can.

RFMVasconcelos (commented):
@shannonbradshaw thanks a lot for taking the time to spot and fix this issue! 

/lgtm
/approve

RFMVasconcelos:
Let's hope this fixes it :)

PR #2661, opened 2021-04-27
Fix working directory path for make commands

plang85:
The makefile is in `${MGMT_DIR}/management`

### DIFF
--- a/content/en/docs/distributions/gke/deploy/management-setup.md
+++ b/content/en/docs/distributions/gke/deploy/management-setup.md
@@ -106,10 +106,10 @@ To deploy a management cluster:
 1. Change to the Kubeflow directory
 
     ```bash
-    cd "${MGMT_DIR}"
+    cd "${MGMT_DIR}/management"
     ```
 
-    Note, all the instructions below assume your current working directory is `${MGMT_DIR}`.
+    Note, all the instructions below assume your current working directory is `${MGMT_DIR}/management`.
 
 1. Fetch the upstream management package
 


RFMVasconcelos (commented):
/assign @joeliedtke

RFMVasconcelos (commented):
Thank you again @plang85 ! 

@Bobgy @zijianjoy same here

zijianjoy:
/approve

Thank you @plang85 for the fix! You might want to resolve the conflict before merging.

And thanks @RFMVasconcelos for directing the PR to us.

Bobgy:
/hold
I think there are some complexities here, behavior is different between MGMT_DIR ends with management or not.

If MGMT_DIR=management, the folder itself will be used, if MGMT_DIR=mgmt, then mgmt/management folder will be created.

That's what I recall, I think we need to verify that

zijianjoy:
@Bobgy That is a good catch! I misread the PR change and the latest document here, the latest `MGMT_DIR` definition is documented in https://www.kubeflow.org/docs/distributions/gke/deploy/management-setup/#configure-environment-variables, so it ends with `management`. Moreover, we use the actual value instead of variable name for directory navigation in the latest instruction https://www.kubeflow.org/docs/distributions/gke/deploy/management-setup/#fetch-kubeflowgcp-blueprints-package.

@plang85 Would you like to confirm if the latest document is accurate? If so, please close this PR. Thanks again for creating PR!

Arhell (commented):
There are conflicts

PR #2669, opened 2021-04-27
fixed env variable in cloud config connector step

mindthevirt:
One step referred to ${MANAGED_PROJECT}" which is not a valid variable and should really be ${MGMT_NAME}

### DIFF
--- a/content/en/docs/distributions/gke/deploy/management-setup.md
+++ b/content/en/docs/distributions/gke/deploy/management-setup.md
@@ -193,7 +193,7 @@ The easiest way to do this is to grant the Google Cloud service account owner pe
 1. Set the managed project:
 
    ```bash
-   kpt cfg set ./instance managed-project "${MANAGED_PROJECT}"
+   kpt cfg set ./instance managed-project "${MGMT_PROJECT}"
    ```
 
 1. Update the policy:


mindthevirt:
@googlebot I signed it!

RFMVasconcelos (commented):
Thank you for this fix @mindthevirt! 

@Bobgy this is GKE specific, do you confirm this fix? :)

zijianjoy:
Hello @mindthevirt , thank you for the PR. However, this variable ${MANAGED_PROJECT} is made a different name than ${MGMT_NAME} on purpose, because __Management cluster__ can live in a GCP project different from __Kubeflow cluster__. We want to ask user to use `kpt` to set the __Kubeflow cluster__'s GCP project ID here. So the __Management cluster__ can manage __Kubeflow cluster__. The lack of explanation about this env variable is causing confusion here. 

But if you deploy both Management cluster and Kubeflow cluster in the same GCP project, then naturally the value of ${MANAGED_PROJECT} and ${MGMT_NAME} is the same. We should explain clearly about their meanings, and probably use ${KUBEFLOW_PROJECT} instead of ${MANAGED_PROJECT} for the Kubeflow cluster project ID. Because `MANAGED`, `MANAGEMENT` and `MGMT` are confusing terms when they stay near each other!

So I suggest making the change like below:

```
Note that Management cluster can live in a different Google Cloud project from Kubeflow clusters, admin should assign owner permission to Management cluster’s service account. 

1. Assign Kubeflow cluster project ID to an environment variable
KUBEFLOW_PROJECT=<YOUR-KUBEFLOW-CLUSTER-GCP-PROJECT-ID>

2. Configure the kpt setter with Kubeflow cluster project:
kpt cfg set ./instance managed-project "${KUBEFLOW_PROJECT}"

Remaining steps....
```

RFMVasconcelos:
Thank you for your comment @zijianjoy ! 

@mindthevirt would you be able to edit this PR? Or should @zijianjoy create a separate one?

RFMVasconcelos:
Thank you for your comment @zijianjoy ! 

@mindthevirt would you be able to edit this PR? Or should @zijianjoy create a separate one?

zijianjoy:
Now that the new deployment steps have been updated. You can find the content in https://www.kubeflow.org/docs/distributions/gke/deploy/deploy-cli/#authorize-cloud-config-connector-for-each-kubeflow-project.

I suggest to close this PR since we have addressed the issue altogether with the deployment doc update. How do you think? @RFMVasconcelos 

Thank you Jan @mindthevirt again for creating the PR to improve kubeflow doc! Please feel free to try out the new document and submit new PR if you spotted anything.

PR #2676, opened 2021-04-29
Small fix - Remove non-rendered note

RFMVasconcelos:
cc @evilnick

### DIFF
--- a/content/en/docs/distributions/charmed/install-kubeflow.md
+++ b/content/en/docs/distributions/charmed/install-kubeflow.md
@@ -63,10 +63,8 @@ You can list your models with the `juju models` command.
 
 #### 5. Deploy Kubeflow
 
-[note type="caution" status="MIN RESOURCES"]
 To deploy `kubeflow`, you'll need at least 50Gb available of disk, 14Gb of RAM, and 2 CPUs available in your machine/VM.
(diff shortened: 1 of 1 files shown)


evilnick (commented):
I think the style in these docs for such notes is to use the site shortcode. For example:

```

{{% alert color="warning" %}}
note content 
{{% /alert %}}
```

RFMVasconcelos:
Thank you for the review and suggestion @evilnick ! Can you please lgtm?

evilnick:
I'm really sorry - now that the warning note is rendering properly, it seems that the style for <code> inside a warning is different. I will file an issue on that because it is barely readable.

Screenshot from 2021-04-29 13-04-19

For now I would just remove the backticks around those terms.

evilnick:
Added issue - https://github.com/kubeflow/website/issues/2677

RFMVasconcelos:
Fixed!

evilnick:
@RFMVasconcelos did we ever figure out why these tests were failing? doesn't look like it is anything to do with the content of this PR

RFMVasconcelos:
@evilnick no idea why the tests are falling and it appears we are not being able to trigger retests. #2698 replaces this PR.

PR #2692, opened 2021-05-04
Patch no longer needed in v1.3

RFMVasconcelos:
The documented patch was only required in v1.2, we can now remove it from this docs page. 

@DomFleischmann can you ptal?


DomFleischmann (approved):
LGTM

PR #2696, opened 2021-05-05
(GKE) Update GKE related outdated docs

zijianjoy:
Partial: https://github.com/kubeflow/gcp-blueprints/issues/258

Update GKE doc and remove banner:

1. Kubeflow customization on GKE
2. Cloud Filestore


content/en/docs/distributions/gke/cloud-filestore.md
```diff
@@ -2,15 +2,9 @@
 title = "Using Cloud Filestore"
 description = "Using Cloud Filestore with Kubeflow"
 weight = 60
-                    
-+++
```
Bobgy, line 6: Note, it's related to the title.
Removing this "+++" caused your tests to fail.

content/en/docs/distributions/gke/cloud-filestore.md
```diff
@@ -2,15 +2,9 @@
...
-{{% alert title="Out of date" color="warning" %}}
-This guide contains outdated information pertaining to Kubeflow 1.0. This guide
-needs to be updated for Kubeflow 1.1.
-{{% /alert %}}
 
 This guide describes how to set up and use Cloud Filestore with Kubeflow on 
-Google Cloud Platform (GCP).
+Google Cloud (GCP).
```
Bobgy, line 7: nit: remove (GCP)

content/en/docs/distributions/gke/cloud-filestore.md
```diff
@@ -70,44 +55,49 @@ This guide assumes the following settings:
...
--->
+3. Apply the configuration to filestore deployment `${KF_NAME}-nfs`
 
-```
-cd ${KF_DIR}/gcp_config
-gcloud --project=${PROJECT} deployment-manager deployments create ${KF_NAME}-nfs --config=gcfs.yaml
-```
+    ```
+    cd ${KF_DIR}/common/cnrm
+    gcloud --project=${PROJECT} deployment-manager deployments create ${KF_NAME}-nfs --config=gcfs.yaml
```
Bobgy, line 97: In fact, because we no longer recommend deployment-manager. I'd suggest still removing cloud-filestore doc, people can simply read cloud filestore docs to learn how to create it. (If we do decide to update cloud-filestore doc, we should use cnrm to install it)

content/en/docs/distributions/gke/customizing-gke.md
```diff
@@ -14,16 +14,12 @@ Kubernetes Engine (GKE) on Google Cloud.
...
 **apply**, so that you can modify your configuration before deploying your 
 Kubeflow cluster.
 
-Follow the guide to [deploying Kubeflow on Google Cloud](/docs/gke/deploy/deploy-cli/).
-When you reach the 
-[setup and deploy step](/docs/gke/deploy/deploy-cli/#deploy-kubeflow),
-**skip the `kfctl apply` command** and run the **`kfctl build`** command 
-instead, as  described in that step. Now you can edit the configuration files 
-before deploying Kubeflow.
+Follow the guide to [deploying Kubeflow on Google Cloud](/docs/gke/deploy/deploy-cli/). You can add your patches in corresponding component folder, and include those patches in `kustomization.yaml` file. Learn more about the usage of [kustomize](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/). You can also find the exisitng kustomization in [kubeflow/gcp-blueprints](https://github.com/kubeflow/gcp-blueprints) as example. After adding the patches, you can run `make hydrate` to validate the resulting resources. Finally, you can run `make apply` to deploy the customized Kubeflow.
```
Bobgy, line 21: For discussion: I don't think you need to address these comments for this release.

If they add patches in `kustomization.yaml` directly, how do they upgrade?
my current thoughts to support upgrade is -- when user needs customizations, they move gcp manifests to a subfolder and write their own kustomization.yaml there instead, for example:

```
cd apps/pipelines
mkdir gcp
mv ./* gcp/
# build user customizations in apps/pipelines, including Makefile, kustomization.yaml etc
```

1 month later, user wants to upgrade pipelines:

```
kpt pkg update apps/pipelines/gcp <latest-repo-url>
```

In this way, upgrade and customization do not conflict.

Tradeoff: this approach needs a bit more manual work for upgrade, it's not worth it when customization is simple and can easily be upgraded and resolve merge conflicts.

To save the folder move step, we might consider creating a `gcp` folder upfront for every app/common, and keep minimum wrapper outside of it, but this seems like pre-mature optimization. Asking people who need to customize to move their folders seem more reasonable.

content/en/docs/distributions/gke/customizing-gke.md
```diff
@@ -38,68 +34,46 @@ This guide assumes the following settings:
...
-  ```
-  export KF_NAME=<the name of your Kubeflow deployment>
-  ```
+To customize Google Cloud resources, such as your Kubernetes Engine cluster, you can 
+modify the Deployment settings starting in `${KF_DIR}/common/cnrm`.
 
-* The `${PROJECT}` environment variable contains the ID of your Google Cloud project. 
-  You can find the project ID in your
-  `${CONFIG_FILE}` configuration file, as the value for the `project` key.
+This folder is the folder which contains multiple dependencies on sibling directories for Google Cloud resources. So you can start from here by reviewing `kustomization.yaml`. Depends on the type of Google Cloud resources you want to customize, you can add patches in corresponding directory.
```
Bobgy, line 52: nit: simplify to `This folder contains`

content/en/docs/distributions/gke/customizing-gke.md
```diff
@@ -200,19 +160,53 @@ command:
...
+          count: 1
+      metadata:
+        disable-legacy-endpoints: "true"
+    management:
+      autoRepair: true
+      autoUpgrade: true
+    clusterRef:
+      name: "PROJECT/LOCATION/KUBEFLOW-NAME" # {"$kpt-set":"cluster-name"}
+```
+
```
Bobgy, line 209: there's another outdated section "Add GPU node pool to an existing kubeflow cluster" below, remove it?

note, there's a step of installing nvidia driver to each node, you should keep it here.

Bobgy (commented):
/lgtm
thanks for updating these!

content/en/docs/distributions/gke/cloud-filestore.md
zijianjoy: Thank you for spotting it! Fixed.

content/en/docs/distributions/gke/cloud-filestore.md
zijianjoy: I will apply this change to all docs.

content/en/docs/distributions/gke/cloud-filestore.md
zijianjoy: Sounds good, I will remove the cloud firestore doc.

content/en/docs/distributions/gke/customizing-gke.md
zijianjoy: Thank you Yuan for the detailed discussion and thoughtful idea!

I agree that it might be a bit overhead for the need of moving folders. Especially when it comes to mismatch of file location between local and remote repos.

It is also a good point of creating a `gcp` folder, but it mean creating a lot fo `gcp` folders. And creating difficulty when it comes to Kubeflow verion upgrade.

I want to add one idea to this discussion:

We provide a simple kustomization.yaml file to allow users to edit, initially the content in this file references to our gcp-blueprints resource using `remote URL`: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#url-format. This approach makes sure that we don't need to change the kustomization.yaml file on the user side, so they can use it for their own customization. During every upgrade, they just need to change the branch name in the remote URL, and evaluate the result in `build` folder. If they want to pull the manifest from remote, they can do so by following our instruction, but by default their local directory is lightweight and contains only their customization.

However, this will be a large refactoring as well. So I agree that we need more discussion around it.

content/en/docs/distributions/gke/customizing-gke.md
zijianjoy: Done

content/en/docs/distributions/gke/customizing-gke.md
zijianjoy: Sounds good, updated. 

Should I keep the "Add GPU node pool to an existing kubeflow cluster" for commandline instruction?

Bobgy: You can decide, I probably prefer removing it -- to make doc consistent, always use cnrm resources to deployo

content/en/docs/distributions/gke/customizing-gke.md
Bobgy: discussed the idea over gvc, it's against best practices for using off-the-shelf manifests with kustomize

zijianjoy: Agreed, let's direct user for individual component upgrade, and figure out future plan for full kubeflow batch upgrade later.

content/en/docs/distributions/gke/customizing-gke.md
zijianjoy: Sounds good, I have removed the CLI instruction to recommend resource flie approach.

Bobgy (commented):
/lgtm
let's wait for https://github.com/kubeflow/website/pull/2686 to merge at the same time

PR #2698, opened 2021-05-06
Fix un-rendered note

RFMVasconcelos:
Replaces #2676, which is failing tests.


evilnick:
@RFMVasconcelos good work, renders fine now

/lgtm

PR #2699, opened 2021-05-07
Update IP from `xip` to `nip`

RFMVasconcelos:
cc @DomFleischmann


DomFleischmann (changes requested):
There is another mention of `xip.io` in line 64

RFMVasconcelos:
Nice catch! Fixed

RFMVasconcelos:
The tests are failing and automation to relaunch seems not to be working.. 

I'll make another PR. 

@Bobgy @joeliedtke do you know if we should fix the test triggers? Or is it working and I am simply not finding the right command? :)

PR #2700, opened 2021-05-07
Replaces `xip.io` with `nip.io`

RFMVasconcelos:
replaces #2699


RFMVasconcelos:
@DomFleischmann can you please review?

PR #2716, opened 2021-05-14
Upgrading the instance type in cluster launch

goswamig:
m5.large has 2 vCPU. When we do full kubeflow installation
most of its vCPU resources will be occupied and notebook server
will fail with error on unavailbale of vCPU resource. Upgrade the
instance type to m5.2xlarge so that vCPU limit will increase from
2 to 8.

### DIFF
--- a/content/en/docs/distributions/aws/deploy/install-kubeflow.md
+++ b/content/en/docs/distributions/aws/deploy/install-kubeflow.md
@@ -59,7 +59,7 @@ For example:
 export AWS_CLUSTER_NAME=kubeflow-demo
 export AWS_REGION=us-west-2
 export K8S_VERSION=1.18
-export EC2_INSTANCE_TYPE=m5.large
+export EC2_INSTANCE_TYPE=m5.2xlarge
 ```
 
 Now, create a cluster configuration file for use with `eksctl`.


RFMVasconcelos (commented):
Thank you for this remark @goswamig ! 

This is up to AWS distribution leaders what they decide to recommend. Costs will be higher in stronger machines, but if, as you mention, it breaks with current setup, I guess this needs to be reviewed. 

cc @kubeflow/aws

PatrickXYS:
@goswamig Thank you contributions, but m5.large is used because it can serve normal use cases, including individual users, demo, workshop. 

So let's not change default instance type, instead, you can add a note:

```
If users are using kubeflow massively in production, it's recommended to use m5.2xlarge instance.
```

goswamig:
@PatrickXYS  I am getting error "CPU resource not available" when I try to launch notebook server with m5.large.

goswamig:
```
 Warning  FailedScheduling  62s   default-scheduler  0/3 nodes are available: 1 Too many pods, 3 Insufficient cpu.
  Warning  FailedScheduling  62s   default-scheduler  0/3 nodes are available: 1 Too many pods, 3 Insufficient cpu.
```

PatrickXYS:
You can use more than one m5.large instance for your kubeflow testing / development

goswamig:
there is already 3 m5 launched in the config ...

```
cat << EOF > cluster.yaml
---
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: ${AWS_CLUSTER_NAME}
  version: "${K8S_VERSION}"
  region: ${AWS_REGION}

managedNodeGroups:
- name: kubeflow-mng
  desiredCapacity: 3
  instanceType: ${EC2_INSTANCE_TYPE}
EOF
```

PR #2723, opened 2021-05-18
chore: fix getting started button link to not require redirect

Bobgy:
/assign @RFMVasconcelos


Bobgy:
The get started button requires a redirect, when testing locally I noticed redirection do not work, so it's not very convenient.
Also, the links should better use current real urls without redirection to improve browsing experience.

Therefore, I'm sending this PR to fix this.

RFMVasconcelos (commented):
Nice catch @Bobgy ! 

/lgtm
/approve

PR #2748, opened 2021-05-31
Temporary fix

RFMVasconcelos:
Fixes currently arrive faster in the charm store `cs:kubeflow` than in the charmhub version which is what you get when you `juju deploy kubeflow`. 

@DomFleischmann can you please confirm?


RFMVasconcelos:
nevermind, the Microk8s script already accounts for this. 

/close

PR #2752, opened 2021-06-04
Update component-development.md for which types are treated as parameters instead of artifacts

chensun:
- After https://github.com/kubeflow/pipelines/pull/5665, we now treat `bool`, `dict`, and `list` as parameters as well.

### DIFF
--- a/content/en/docs/components/pipelines/sdk/v2/component-development.md
+++ b/content/en/docs/components/pipelines/sdk/v2/component-development.md
@@ -18,10 +18,10 @@ building and running pipelines that are compatible with the Pipelines SDK v2.
 
 ## Before you begin
 
-Run the following command to install the Kubeflow Pipelines SDK v1.6 or higher.
+Run the following command to install the Kubeflow Pipelines SDK v1.6.1 or higher.
 
 ```bash
-pip install kfp==1.6.0rc0
+pip install kfp --upgrade
 ```
 
 For more information about the Kubeflow Pipelines SDK, see the [SDK reference guide][sdk-ref].
--- a/content/en/docs/components/pipelines/sdk/v2/component-development.md
+++ b/content/en/docs/components/pipelines/sdk/v2/component-development.md
@@ -75,7 +75,7 @@ When you design your component's code, consider the following:
 
     _Parameters_ typically represent settings that affect the behavior of your pipeline.
     Parameters are passed into your component by value, and can be of any of
-    the following types: `int`, `float`, `str`, `bool`, `dict`, `list`. Since parameters are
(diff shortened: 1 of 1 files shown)


content/en/docs/components/pipelines/sdk/v2/component-development.md
```diff
@@ -75,7 +75,7 @@ When you design your component's code, consider the following:
 
     _Parameters_ typically represent settings that affect the behavior of your pipeline.
     Parameters are passed into your component by value, and can be of any of
-    the following types: `int`, `double`, `float`, or `str`. Since parameters are
```
chensun, line 78: I'm removing `double` as it's not a Python build-in type. We do recognize `"double"` (as a string) though, but it's more like variation of `float`, similarly as we treat `"integer"` as `int`, `"string"` and `"text"` as `str`.
I feel like keep this list shorter and cleaner (with all Python build-in types) is probably better than listing all variations. WDYT? @joeliedtke

chensun: Also we treat `typing.Dict[...]` and `"JsonObject"` the same as `dict`, and treat `typing.List[...]` and `"JsonArray"` the same as `list`. Let me know if you think we should document all these variations.

Here's the almost complete list in our code: https://github.com/kubeflow/pipelines/blob/14cb98b662bc88a28122dcc5d4bf7d577c7b218f/sdk/python/kfp/dsl/type_utils.py#L35-L49
It's almost complete because it doesn't reflect that we treat `typing.Dict`, `typing.Dict[str, Any]`, etc. as `dict`.

Bobgy:
/lgtm

Thank you for the update! Looks enough to me. Agree, no need to enumerate every corner case.

content/en/docs/components/pipelines/sdk/v2/component-development.md
```diff
@@ -75,7 +75,7 @@ When you design your component's code, consider the following:
...
+    the following types: `int`, `float`, `str`, `bool`, `dict`, `list`. Since parameters are
```
Bobgy, line 78: Forgot to ask, is there a KFP SDK version requirement for this change?

chensun: This is available in 1.6.1 -- the very first stable release we made for Vertex AI preview.
But good point, I do see this doc still mentions 1.6.0rc0. Will update that version.

joeliedtke: Please make the following change to this list. (Otherwise, LGTM.)

`dict`, `list` -> `dict`, or `list`

chensun: Done.

PR #2769, opened 2021-06-15
Update v2-compatibility.md on which types are recognized as parameter types.

chensun:


Bobgy:
/lgtm
Thank you for the fix!

PR #2785, opened 2021-06-21
fix(pipelines): add visualization image back for existing reference

zijianjoy:
Current link is broken because table visualization also references this image: https://www.kubeflow.org/docs/components/pipelines/sdk/output-viewer/#table.


Bobgy:
/lgtm
/approve
Thanks!

PR #2791, opened 2021-06-22
fix(gke) Add explanation of GKE locations: available zones/regions

zijianjoy:
Explain how to select GKE locations for management cluster selection.


Arhell (commented):
/lgtm

PR #2801, opened 2021-06-29
fix(pipelines): image reference path correction

zijianjoy:
The link to table visualization image was not correct before.


Bobgy:
/lgtm
/approve

Thank you!

PR #2858, opened 2021-08-06
(pipelines) Update KFP version requirement for v2-compatible

zijianjoy:
Reference: https://github.com/kubeflow/pipelines/issues/6241


Bobgy (commented):
/lgtm
/approve

PR #2862, opened 2021-08-09
make variable names pythonic

romeokienzler:
replaced Input 1 => input_1, Output 1 => output_1, Parameter 1 => parameter_1 because although usage of white space strings nicely shows flexibility provided it is very uncommon to do so. Variables should be written in camelCase or the_pythonic_way


Arhell (commented):
/assign @RFMVasconcelos

PR #2867, opened 2021-08-11
docs(pipelines): KFP V2 visualization doc. Partial: kubeflow/pipelines#5666

zijianjoy:
Partial: https://github.com/kubeflow/pipelines/issues/5666

Note: Missing HTML/Markdown visualization doc.

Question: Should we specify SDK and backend version for using new visualization approach?


Bobgy:
I think we have the version statement in v2 sdk docs, so it's probably not required here.

Bobgy:
/approve
consider adding an OWNERS file for https://github.com/kubeflow/website/tree/master/content/en/docs/images/pipelines? so that I won't be the only person that can approve this

PR #2870, opened 2021-08-12
Add specific version of Kustomize that works

benjamintanweihao:
I've tested that Kustomize 4.2.0 (the latest that the installation script pulls) works with Kubeflow 1.3.1. However, it is possible that not all paths would work. For example, some of the scripts use `--load_restrictor` instead of `--load-restrictor`. For the outlined steps Kustomize 4.2.0 works fine, though this is not mentioned anywhere. (The impression is that the user should use Kustomize 3.2.0 which didn't work for me)


Arhell (commented):
/lgtm

RFMVasconcelos (commented):
Thank you for testing and correcting this in the docs @benjamintanweihao ! :rocket: 

/lgtm
/approve

PR #2872, opened 2021-08-12
Fixing bug in mlrun serving

alexiguazio:
We've updated the code so we're updating documentation here as well.


terrytangyuan (approved):
Thanks!

/lgtm

RFMVasconcelos (commented):
Thank you @alexiguazio for this fix!

/lgtm
/approve

Issue #2522, opened 2021-03-02
Adding a community-managed GA code to the website


castrojo:
In the Community Meeting on 23 Feb we asked Google for some access to the Google Analytics so that we can have the data to make better decisions when restructuring docs. 

Google told us that corporate policy would make this unfeasible. David Aronchick proposed just adding a 2nd GA code to the website/docs so that there can be a clear separation between a corporate account and one run by the community. There was lazy concensus during the meeting on this. 

As far as I can tell there's no policies in the project on how "the community" can manage a google account (bus factor, trust, etc.), so that needs to be solved still, so just opening this up as a placeholder.

cc @RFMVasconcelos

thesuperzapper:
@castrojo do you want to raise a proposal on `kubeflow/community` with a solution for how to assign access to this new Google analytics account? (I am not sure how GA works with assigning other people as 'viewers')

Currently most ACL is assigned in https://github.com/kubeflow/internal-acls. However, we still need a super-admin account to bootstrap the Google Analytics account, the password for which, will need to be somehow owned by the community.

Obviously this issue will come up again (and more significantly) after the formation of the "Technical Steering Committee"(https://github.com/kubeflow/community/issues/495) as all current super-admin accounts (for things like: GitHub and Google Groups) will need be handed over to the members of that committee.

jlewi:
> David Aronchick proposed just adding a 2nd GA code

Wouldn't a 2nd GA code run into the same problem with Google corporate policies? 

> There was lazy concensus during the meeting on this.

Did anyone from Google actually say a 2nd GA code would be ok?

castrojo:
> Did anyone from Google actually say a 2nd GA code would be ok?

There were googlers at the meeting but none that could speak to this, which is why I filed this in github as an item.

RFMVasconcelos:
One potential path to get this done is:
1. Create a generic gmail address e.g. `kubeflow-docs-tracking@gmail.com`
2. Provide credentials of this account to a few leads in the community
3. Generate a tracking code from the new GA account
4. Make a PR that adds this code to the website
5. Provide view access to community members upon request

I'm happy to proceed if the community feels this is the way to go.

thesuperzapper:
@RFMVasconcelos can you add this issue to the 1.3 project https://github.com/kubeflow/website/projects/2

If you think it makes sense? Having analytics will definitely help us prioritise the docs, which I imagine will get more traffic after 1.3.

thesuperzapper:
[...]

PR #2875, opened 2021-08-13
setup community-owned Google Analytics v4

thesuperzapper:
fixes https://github.com/kubeflow/website/issues/2522 (which has been open since March 2021)

__This PR does two things:__
1. Upgrade the Google Analytics code to V4
2. Sets up a community-owned Google Analytics

This PR is important because currently no one has access to important information like "page views", and the "feedback" buttons at the bottom of the pages.

I am currently the only admin on the new Google Analytics dashboard, but we can easily add the approvers from [OWNERS](https://github.com/kubeflow/website/blob/master/OWNERS) based on their Gmail account.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -72,7 +72,7 @@ anchor = "smart"
 
 [services]
 [services.googleAnalytics]
-id = "UA-135379910-1"
+id = "G-DG00RNX3X1"
 
 # Docsy: Language configuration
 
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@RFMVasconcelos @Bobgy what do you think?

Arhell (commented):
/lgtm

RFMVasconcelos:
Thank you for moving this forward @thesuperzapper ! 

I believe this is a step in the right direction. Once we have the data, we should present it publicly in a community meeting. 

/lgtm

Bobgy:
/hold
I need to confirm, but I believe this is against Google's policy for the same reason that we cannot grant access to the existing Google analytics account.

PR #2876, opened 2021-08-14
Update component-development.md to showcase v2 compatible data passing.

chensun:
The current tutorial contains legacy data passing -- specifically passing a constant value to an input expect `inputPath`. This is not compatible with KFP v2 and Vertex Pipelines.
Update the sample to use compatible data passing that works on both KFP v1 and Vertex Pipelines.

Fixes: https://github.com/kubeflow/pipelines/issues/6112


RFMVasconcelos:
@kubeflow/wg-pipeline-leads could you PTAL?

neuromage:
Thanks @RFMVasconcelos 
/approve

/cc @joeliedtke you may need to check/approve this

PR #2899, opened 2021-08-26
Add instructions to add AWS Fargate profiles before deploying

dLobatog:
Without a fargate profile, the deployment will fail, as at some point `cert-manager` won't be able to start. 

An error like this shows up, which doesn't have any clues as to what's happening.
```
WARN[0261] Encountered error applying application cert-manager:  (kubeflow.error): Code 500 with message: Apply.Run : error when creating "/tmp/kout108981562": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request  filename="kustomize/kustomize.go:284
```

But when I looked at the deployments `kfctl` was creating, specifically at `kubectl get deployments -n cert-manager` - I saw none of them were available, because all of them were lacking a Fargate profile in the new namespace.

I'll check in kfctl itself if there's any way to bootstrap the profile automatically, is this something that would be accepted? Perhaps some users wouldn't want `kfctl` to automatically create a Fargate profile (it's an equivalent procedure of creating nodes in a k8s cluster, so basically running the command could cost you $ without realizing it). Thoughts?

### DIFF
--- a/content/en/docs/distributions/aws/deploy/install-kubeflow.md
+++ b/content/en/docs/distributions/aws/deploy/install-kubeflow.md
@@ -187,6 +187,16 @@ The method of attaching required IAM policies to the EKS node instance role is s
 
 ## Deploy Kubeflow
 
+1. (Skip if your EKS nodes are not managed by AWS Fargate). If your EKS cluster is running on AWS Fargate, you will need to create a Fargate profile for your nodes.
+    ```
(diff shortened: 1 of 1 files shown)


Arhell (commented):
@kubeflow/aws

surajkota:
Hi @dLobatog can you provide more context around your deployment environment? you trying to deploy Kubeflow on EKS with Fargate?
Kubeflow version, EKS version?

Please also take at the latest docs for 1.3.1 on the [Kubeflow website](https://www.kubeflow.org/docs/distributions/aws/), ref #3169

PR #2934, opened 2021-09-12
Update getting-started.md

jbottum:
I am suggesting a name change for authorizationpolicy.yaml.    I think that AuthZ is provided by K8s RBAC (edit, view) whereas AuthN is provided by Istio with underlying cluster and LDAP/AD credentials.  So I am suggesting that instead of an istio based authorizationpolicy.yaml that we should call this authenticationpolicy.yaml.  I am not sure the policy technically works in K8s or K8 but the name is more description (and does not overlap/confuse with K8s RBAC, which is used for authorization).  I also proposed some changes to fixe some small typos (administration to administrative, belogings to belongings.

### DIFF
--- a/content/en/docs/components/multi-tenancy/getting-started.md
+++ b/content/en/docs/components/multi-tenancy/getting-started.md
@@ -69,7 +69,7 @@ user in the Kubeflow cluster.  Here an administrator is a person who has
 [*cluster-admin*](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles)
 role binding in the Kubernetes cluster. This person has permissions to create
 and modify Kubernetes resources in the cluster. For example, the person who
-deployed Kubeflow will have administration privileges in the cluster.
+deployed Kubeflow will have administrative privileges in the cluster.
 
 ### Pre-requisites: grant user minimal Kubernetes cluster access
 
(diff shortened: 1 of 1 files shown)


RFMVasconcelos (commented):
Thank you @jbottum! 

/lgtm
/approve

content/en/docs/components/multi-tenancy/getting-started.md
```diff
@@ -316,11 +315,11 @@ subjects:
   name: userid@email.com   # replace with the email of the user from your Active Directory case sensitive
 ```
 
-Create an authorizationpolicy.yaml file with the following content on your local machine:
+Create an authenticationpolicy.yaml file with the following content on your local machine:
 
 ```
 apiVersion: security.istio.io/v1beta1
-kind: AuthorizationPolicy
+kind: AuthenticationPolicy
```
yhwang, line 322: unfortunately, you can't change the `kind` here.

jbottum:
@yhwang thanks for clarifying, I have deleted the change.    It is confusing to me that this Kind uses the Authorizationpolicy with istio.     It seems to me that KF uses K8s RBAC for AuthZ  and Istio for AuthN.

thesuperzapper:
@jbottum @RFMVasconcelos this needs to be reverted, as it suggests using a Kuberetnes `kind` which does not exist.

The following is not a valid resource `kind`:
```yaml
apiVersion: security.istio.io/v1beta1
kind: AuthenticationPolicy
```

Issue #2937, opened 2021-09-13
Cut 1.3 branch for website


kimwnasptd:
We need to cut a new branch for the website, in order to be able to merge PRs for the 1.4 release https://github.com/kubeflow/website/issues/2879. You can see the process we follow here https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md#version-the-website.

@RFMVasconcelos @knkski what is the state of this? Is there something I can help with?

cc @kubeflow/release-team

RFMVasconcelos:
As I mentioned I haven't been able to do this - @kimwnasptd would you be able to cut the branch? 

@Bobgy could you please tackle or find another Googler to tackle the release steps of docs pointing to the new `v1.3-branch` once it is there?

kimwnasptd:
> @kimwnasptd would you be able to cut the branch?

Definitely! Since I'm in the release team I have permissions to create branches in the `website` repo.

Before cutting the branch I want to ensure that the `v1.3-branch` will not have commits that correspond to future versions of Kubeflow. Specifically:
* @Bobgy @zijianjoy I see commits in the current `master` that are for Pipelines 1.7.0. I guess these should not be included in the 1.3 version of the docs right?
* @andreyvelich I see some commits that update the Katib docs https://github.com/kubeflow/website/pull/2880 https://github.com/kubeflow/website/pull/2851. Are the commits intended for 1.4 and should not be present in 1.3?

andreyvelich:
Yes, that commits should be part of Kubeflow 1.4

zijianjoy:
@kimwnasptd I am tempting to say that users can upgrade KFP independently in Kubeflow 1.3, therefore KFP 1.7.0 is possible to be included as part of `kubeflow doc` 1.3 branch, and we have called out the version requirement for KFP in those KFP 1.7.0 related documents. But defer to @Bobgy to make the decision.

Reference: [KFP website doc history](https://github.com/kubeflow/website/commits/master?before=8ce6d0cf93840b28a072cbfd749a530089fe0053+35&branch=master&path%5B%5D=content&path%5B%5D=en&path%5B%5D=docs&path%5B%5D=components&path%5B%5D=pipelines)

Bobgy:
Agree with @zijianjoy.
For configuring the website, I work with @zijianjoy to do it. Let us know once you cut 1.3 branch.

kimwnasptd:
[...]

PR #2953, opened 2021-09-15
pipelines: KFP Version for Kubeflow 1.3 table update. Partial #2937

zijianjoy:
Partial #2937.

We should make sure to keep updating this in the future, or write it down for future Kubeflow releases.


Bobgy:
/lgtm
thanks!

PR #2956, opened 2021-09-16
fix broken Slack link

tpepper:
I noticed the kubeflow.org footer link for Slack did not work when I
first clicked on it, ie:
	"This link is no longer active"

Now, I was not a member of the Slack workspace when I first clicked
it.  Having subsequently found a working link elsewhere and
successfully signed up for the Slack workspace for Kubeflow, that
link does actually try to redirect me to the workspace.

I'm not sure how these redirects are implemented, or if this patch
represents an appropriate fix, but...

Using the url to the one listed for an invite in the
content/en/docs/about/contributing.md did succeed for me, both before
(ie: takes me to a sign up page) and after (takes me to the workspace)
membership.

Signed-off-by: Tim Pepper <tpepper@vmware.com>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -220,7 +220,7 @@ no = 'Sorry to hear that. Please <a href="https://github.com/kubeflow/website/is
         desc = "Development takes place here!"
 [[params.links.user]]
 	name = "Slack"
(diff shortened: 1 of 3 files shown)


Arhell (commented):
/assign @RFMVasconcelos

PR #2962, opened 2021-09-20
Katib: Add components guide

andreyvelich:
Fixes: https://github.com/kubeflow/katib/issues/1668.

This PR: https://github.com/kubeflow/katib/pull/1667 removes components guide from the README, we should add it in the Kubeflow guides.

@tenzen-y I am not sure that we need to mention `kustomize` in the installation guides, as we discussed here: https://github.com/kubeflow/katib/pull/1667#issuecomment-921109438.
I noticed that Kubeflow Pipelines also uses `kubectl apply -k`: https://www.kubeflow.org/docs/components/pipelines/installation/standalone-deployment/#deploying-kubeflow-pipelines.
More advance users can always clone the Katib repo and use `kustomize` to modify our manifests before installation, we can explain it in the [developer guide](https://github.com/kubeflow/katib/blob/master/docs/developer-guide.md).

What do you think about it @tenzen-y @gaocegege @johnugeorge ?

cc @RFMVasconcelos

/hold for the review

### DIFF
--- a/content/en/docs/components/katib/hyperparameter.md
+++ b/content/en/docs/components/katib/hyperparameter.md
@@ -34,7 +34,7 @@ To install Katib as part of Kubeflow, follow the
 
 If you want to install Katib separately from Kubeflow, or to get a later version
 of Katib, you can use one of the following Katib installs. To install the specific
-Katib release (e.g. `v0.11.1`), modify `?ref=master` to `?ref=v0.11.1`.
+Katib release (e.g. `v0.11.1`), modify `ref=master` to `ref=v0.11.1`.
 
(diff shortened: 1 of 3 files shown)


tenzen-y:
> More advance users can always clone the Katib repo and use kustomize to modify our manifests before installation, we can explain it in the developer guide.

LGTM! @andreyvelich 

It might be better to change the required kustomize version in the following section.

https://github.com/kubeflow/katib/blob/a6cf571ea026fec6b9572f0b3cae4d8864771dda/docs/developer-guide.md#requirements

gaocegege (approved):
LGTM :+1:
/lgtm

PR #2985, opened 2021-09-30
Remove reference to insecure password-hashing site

reed9999:
The site referenced in the previous revision, passwordhashing.com, is profoundly insecure. It seems to the display last 10 BCrypt requests, *from any user worldwide*, in plain text. Clearly this is not an appropriate page to which users should be directed. 

Should references to this site also be forcibly removed from older revisions of this document?


reed9999:
@googlebot I signed it!

Arhell (commented):
@kubeflow/AWS

surajkota:
@reed9999 Thank you for the contribution. We are removed this reference as part of #3169. Please reopen the PR if you find further changes are needed

/close

PR #3012, opened 2021-10-06
remove mention of Kubeflow 1.3 on OpenShift distro page

shannonbradshaw:
It seems out place to mention 1.3 on this page. Maintenance of docs easier if we keep version information for distributions in one location here.
https://www.kubeflow.org/docs/started/installing-kubeflow/


shannonbradshaw:
@jbottum @kimwnasptd

PR #3032, opened 2021-10-17
State how to install standalone training operators

davidxia:
without installing rest of Kubeflow by linking to
https://github.com/kubeflow/training-operator#installation.

Use case: I have a cluster with some core Kubeflow components installed
already. I just want to test out the training operators and don't want
to upgrade my existing Kubeflow components. Or I have a cluster without
Kubeflow and want to try out the training operators by itself.


terrytangyuan (approved):
/lgtm

PR #3022, opened 2021-10-12
pipelines: Follow up on connecting to KFP in multi-user mode.

zijianjoy:
To address the comments in https://github.com/kubeflow/website/pull/2905:

* Process about process of AuthN and AuthZ
* Cross-namespace section title
* ServiceAccountToken explanation.
* Overall doc structure.

cc @bartgras @shuesc1

PR #3037, opened 2021-10-19
kfp: Update Service account explanation in mult-user access

zijianjoy:
Addressing comments in https://github.com/kubeflow/website/pull/3022.

Thank you @bartgras and @Bobgy for the information and comments!


content/en/docs/components/pipelines/sdk/connect-api.md
```diff
@@ -165,6 +166,9 @@ Choose your use-case from one of the options below:
                 audience: pipelines.kubeflow.org      
   ```
 
+  Note that this example uses `default-editor` in `my-namespace` as the service account identity, but you can configure
+  to use any service account that runs in your Pod.
```
Bobgy, line 170: We are probably missing the explanation of what permissions the service account needs.

https://github.com/kubeflow/pipelines/issues/3513#issuecomment-721784849 is best documentation we have for now

PR #3078, opened 2021-11-25
use submodule for docsy & improve website styling

thesuperzapper:
This PR might look very large, but it's mostly just deleting docsy code, and instead using [git-submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) so we don't have to store all the docsy code in our repo.

### Changes:
- Updates Hugo to `0.89.4`
- Updates Docsy to commit [`3ae94a0573a8c14f61c832b6e7fdcc46af62187c`](https://github.com/google/docsy/commit/3ae94a0573a8c14f61c832b6e7fdcc46af62187c)
- Adds "folding buttons" to the sidebar menues (picture 1)
- Improves styling on "feedback" (picture 2)

### Picture 1 (folding buttons)

Screen Shot 2021-11-26 at 1 57 36 am

### Picture 2 (feedback)

Screen Shot 2021-11-26 at 1 59 07 am


thesuperzapper:
@Bobgy @james-jwu @zijianjoy can you guys take a look at this PR?

NOTE: It removes an incredible number of files because it moves to git-submodules for the docsy code (rather than cloning docsy into our repo).

NOTE: To help review it, I recommend looking at the individual commits rather than the full GitHub diff (which is 5000+ files).

assets/scss/_nav.scss
```diff
@@ -116,7 +116,128 @@
...
+
+}
+
+@media (hover: hover) and (pointer: fine) {
+
+    nav.foldable-nav {
+
+        .ul-1 .with-child > label:hover:before {
+            color: $primary;
+            transform: rotate(30deg);
```
Arhell, line 233: I think this effect needs to be removed, it is not needed
**transform**
On hover, you can already see what changes color

thesuperzapper: This change actually was not from me, it was from the docsy update (needed because we were fully overwriting the docsy `_nav.scss` file).

I have now moved these overrides into `_styles_project.scss` (where docsy actually recommends all custom staying is put).

thesuperzapper:
In the latest commits I have:
- removed the need to override `_nav.scss` (see [this comment](https://github.com/kubeflow/website/pull/3078#discussion_r757168745))
- Made the "search" always display in the sidebar (even on desktop screens)
- Hidden the search in the navbar

### Sidebar (with search)
Screen Shot 2021-11-26 at 11 36 40 am

zijianjoy:
It can be helpful to describe how to keep the docsy submodule up-to-date, in this documentation section: https://github.com/kubeflow/website#working-with-the-theme

assets/scss/_nav.scss
zijianjoy: I guess the question is how about we override the default behavior, to remove `transform` animation in Kubeflow website? Because the default behavior is to rotate the pointer when being hovered over, which is a bit strange. (If the sub-section is already expanded, the pointer will rotate backward by 30 degree)

thesuperzapper: @zijianjoy take that up on the [`google/docsy` repo](https://github.com/google/docsy) if you want, but its unrelated to this PR.

Here is [the exact upstream docsy code](https://github.com/google/docsy/blob/3ae94a0573a8c14f61c832b6e7fdcc46af62187c/assets/scss/_nav.scss#L206-L210).

thesuperzapper:
> It can be helpful to describe how to keep the docsy submodule up-to-date, in this documentation section: https://github.com/kubeflow/website#working-with-the-theme

@zijianjoy I have added a guide to update the docsy theme submodule.

Is there anything else this PR needs before it can be merged?

thesuperzapper:
@zijianjoy or @Bobgy I just rebased this, it would be amazing to get this merged, do you think you will have time to review it soon?

Bobgy:
/lgtm
Defer to @zijianjoy for final approval

zijianjoy (commented):
/approve

Thank you so much for the doc improvement!

thesuperzapper:
@zijianjoy thanks heaps, I am very excited about this PR getting merged!

PR #3079, opened 2021-11-26
Revert "Update getting-started.md (#2934)"

thesuperzapper:
This reverts commit af22175be74938e6290d2fb2abf24c7150e71951.

This is because #2934 renames `AuthorizationPolicy` to `AuthenticationPolicy`, which is obviously not correct, as a `AuthenticationPolicy` resource kind does not exist.

PR #3084, opened 2021-11-30
Use AuthorizationPolicy in instructions

benjamintanweihao:
The instructions reference `AuthenticationPolicy` which I couldn't find anywhere. However, there's AuthorizationPolicy`. I've tested this on my cluster and it's working.


thesuperzapper:
We reverted the PR which caused this issue in PR: https://github.com/kubeflow/website/pull/3079

PR #3089, opened 2021-12-06
Delete v2-compatibility.md

chensun:
Part of https://github.com/kubeflow/pipelines/issues/6829


Arhell (commented):
/assign @zijianjoy

content/en/docs/components/pipelines/sdk-v2/v2-compatibility.md
```diff
@@ -1,158 +0,0 @@
...
-title = "Introducing Kubeflow Pipelines SDK v2"
-description = "Overview of how to get started with Kubeflow Pipelines SDK v2"
-weight = 10
-+++
-
-{{% beta-status
-  feedbacklink="https://github.com/kubeflow/pipelines/issues" %}}
-
-The Kubeflow Pipelines SDK provides a set of Python packages that you can use to specify and run your machine learning (ML) workflow as a pipeline. Version 2 of the SDK adds support for tracking pipeline runs and artifacts using ML Metadata. Starting with Kubeflow Pipelines 1.6, you can build and run pipelines in v2 compatibility mode.
-
```
zijianjoy, line 11: Since user can still use V2 compatible in pre-1.9.0 SDK, and we haven't officially released V2 yet, can we call out the supported SDK versions for V2 compatible, instead of removing the whole page here?

zijianjoy: If that is the case, should we remove all other pages under `sdk-v2` as well? https://www.kubeflow.org/docs/components/pipelines/sdk-v2/. Deleting only one page but not other pages seems to leave the kubeflow doc in a confusing state.

zijianjoy:
Please note that there are multiple places in the kubeflow doc which reference this page. https://github.com/kubeflow/website/search?q=%22Learn+more+about+Pipelines+SDK+v2%22.

If we are to remove this page, we should update the reference as well. But I would suggest updating the existing doc to say that the V2 mode is not available on KFP backend, until V2 mode is released.

PR #3090, opened 2021-12-07
Updated K3ai version and re-freshed docs

alefesta:
K3ai is now v1 and has been migrated to the k3ai-project ..So I've updated the docs accordingly.
Also, K3ai does support the full components of Kubeflow so I think I'll open up another PR to add other capabilities of K3ai to the docs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "markdownlint.config": {
+        "MD028": false,
+        "MD025": {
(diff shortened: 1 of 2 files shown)


alefesta:
also please notice that I'm the middle of migrating my old account @alfsuse  to this one @alefesta  so I'm the original author of this doc.

Arhell (commented):
/ok-to-test

alfsuse (approved):

Issue #3109, opened 2022-01-05
Broken presubmit when syncing submodule


zijianjoy:
When creating a PR, the presubmit for Netlify preview failed with following messages:

```

2:26:11 PM: ​
2:26:11 PM: ────────────────────────────────────────────────────────────────
2:26:11 PM:   1. build.command from netlify.toml                            
2:26:11 PM: ────────────────────────────────────────────────────────────────
2:26:11 PM: ​
2:26:11 PM: $ git submodule update --init --recursive && hugo --gc --minify
2:26:11 PM: Start building sites …
2:26:11 PM: hugo v0.89.4-AB01BA6E+extended linux/amd64 BuildDate=2021-11-17T08:24:09Z VendorInfo=gohugoio
2:26:23 PM: Error: Error building site: TOCSS: failed to transform "scss/main.scss" (text/x-scss): SCSS processing failed: file "stdin", line 6, col 1: File to import not found or unreadable: ../vendor/bootstrap/scss/bootstrap.
2:26:23 PM: Total in 12316 ms
2:26:23 PM: ​
2:26:23 PM: ────────────────────────────────────────────────────────────────
2:26:23 PM:   "build.command" failed                                        
2:26:23 PM: ────────────────────────────────────────────────────────────────
2:26:23 PM: ​
2:26:23 PM:   Error message
2:26:23 PM:   Command failed with exit code 255: git submodule update --init --recursive && hugo --gc --minify
2:26:23 PM: ​
2:26:23 PM:   Error location
2:26:23 PM:   In build.command from netlify.toml:
2:26:23 PM:   git submodule update --init --recursive && hugo --gc --minify
2:26:23 PM: ​
2:26:23 PM:   Resolved config
2:26:23 PM:   build:
2:26:23 PM:     command: git submodule update --init --recursive && hugo --gc --minify
2:26:23 PM:     commandOrigin: config
2:26:23 PM:     environment:
2:26:23 PM:       - REVIEW_ID
2:26:23 PM:       - HUGO_VERSION
2:26:23 PM:       - NODE_VERSION
2:26:23 PM:     publish: /opt/build/repo/public
2:26:23 PM:     publishOrigin: config
```

Line of execution: https://github.com/kubeflow/website/blob/master/netlify.toml#L3

Log: https://app.netlify.com/sites/competent-brattain-de2d6d/deploys/61d61ad91cc55f000878efde

cc @thesuperzapper

thesuperzapper:
[...]

PR #3114, opened 2022-01-09
fix netlify build command

thesuperzapper:
As raised in https://github.com/kubeflow/website/issues/3109, the Netlify pre-submit tests are failing because they can't see `../vendor/bootstrap/scss/bootstrap`.

Docsy itself has bootstrap as a submodule, and for some reason `git submodule update --init --recursive` is not actually recursively downloading the submodules from the `./themes/docsy` folder.

The command [recomended by the docsy docs](https://www.docsy.dev/docs/deployment/#deployment-with-netlify) is `cd themes/docsy && git submodule update -f --init && cd ../.. && hugo`, so I have used that, but have also kept the production flags `--gc --minify` to improve site performance.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
   publish = "public"
-  command = "git submodule update --init --recursive && hugo --gc --minify"
+  command = "cd themes/docsy && git submodule update -f --init && cd ../.. && hugo --gc --minify"
 
 [context.deploy-preview.environment]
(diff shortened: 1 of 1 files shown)


zijianjoy:
/lgtm
/approve

Thank you for the fix! Interesting to learn that `--recursive` flag doesn't work, as the instruction here is working locally to me: https://github.com/kubeflow/website#run-local-hugo-server.

PR #3156, opened 2022-02-17
gke: Documentation for Config Controller

zijianjoy:
Corresponding Config Controller change: https://github.com/kubeflow/gcp-blueprints/pull/347.
Partial: https://github.com/kubeflow/gcp-blueprints/issues/341

With the migration to Config Controller, user doesn't need to manually manage or upgrade their Management cluster. This is one-time cost for setting up Management cluster. The integration with Kubeflow project is also simplified.


Arhell (commented):
lgtm

PR #3179, opened 2022-03-04
charmed kubeflow: updates command for creating controller

DnPlas:
This PR updates the command for creating a controller. It is necessary to specify the agent version as newer ones are not supported. This is an important step that shouldn't be missed as some critical issues could be raised.


ca-scribner (approved):

DomFleischmann (approved):

Arhell (commented):
/lgtm

PR #3199, opened 2022-03-16
Charmed Kubeflow: Remove workaround option for creating controller

DomFleischmann:
Remove the workaround introduced in #3179 which bootstrapped an older version of a juju controller due to 2.9.25 not working correctly with charmed kubeflow. With 2.9.26 this is now fixed and the workaround is not needed anymore.


DnPlas (approved):
/lgtm

PR #3220, opened 2022-04-12
Removed configuration with AWS credentials section

judyheflin:
Removed section "Configure containers to use AWS credentials" on AWS Kubeflow Pipelines page as per dev team request.


mbaijal (approved):
LGTM

PR #3242, opened 2022-05-02
Connecting to Kubeflow Pipelines using the SDK client from within multi-user cluster is not working

### LINKED ISSUE #3241: Connecting to Kubeflow Pipelines using the SDK client from within multi-user cluster is not working
By following [Connecting to Kubeflow Pipelines using the SDK client in multi-user-mode](https://www.kubeflow.org/docs/components/pipelines/sdk/connect-api/#multi-user-mode) docs to setup a **Jupyter notebook** I understand that setting the `PodDefault` with the sample values would work (adjusted to use the desired namespace) when running the suggested python code:

```python
import kfp
client = kfp.Client(host='http://ml-pipeline-ui.kubeflow:80')
print(client.list_experiments())
```

But this is not the case, this raises a `ApiException: (403)`: it seems that by setting the `host` parameter the `kfp` client doesn't fetch the in-cluster token, therefore the snippet should be more like:

```python
import kfp
client = kfp.Client()
print(client.list_experiments())
```

GerardoGR:
Closes #3241


Arhell (commented):
/ok-to-test

PR #3251, opened 2022-05-19
Update mlrun.md

mohamadmansourX:
The URL is targeted towards a landing page for the MLRun Serving.

After the release of version 0.10.x, there is no more landing index.html page as such : https://docs.mlrun.org/en/latest/serving/index.html

The last one was in version 0.9.x: https://docs.mlrun.org/en/v0.9.3/serving/index.html

So I'm suggesting to changing it to: https://docs.mlrun.org/en/latest/serving/build-graph-model-serving.html


Arhell (commented):
/lgtm

PR #3256, opened 2022-05-24
[IKS] Update cluster creation instructions

Tomcli:
The command line instructions now recommend to specify the `PUBLIC_VLAN_ID` and `PRIVATE_VLAN_ID` since iks 1.21+ version. Hence we need to update the instructions to avoid errors.


yhwang (approved):
/lgtm

PR #3274, opened 2022-06-08
Add new Updated Kubeflow Architecture Diagram

akgraner:
Adding updated image so I can updated the started page with the new image. 

Kubeflow_architecture_22

I think I've added it correctly, but added it in the comments as well.


Arhell (commented):
/assign @johnugeorge

PR #3316, opened 2022-08-05
Bump KFP version

gkcalat:


content/en/docs/distributions/gke/deploy/deploy-cli.md
```diff
@@ -187,7 +187,7 @@ Follow these steps to verify the deployment:
    your `kubectl` credentials to point to the new cluster:
 
     ```bash
-    gcloud container clusters get-credentials "${KF_NAME}" --zone "${ZONE}" --project "${KF_PROJECT}"
+    gcloud container clusters get-credentials "${KF_NAME}" --zone "${REGION}" --project "${KF_PROJECT}"
```
zijianjoy, line 190: Since the flag is `--zone`, we should use `${ZONE}` for this flag.

layouts/shortcodes/pipelines/latest-version.html
```diff
@@ -1 +1 @@
-1.8.3
+1.8.4
```
zijianjoy, line 1: NIT: KFP version here is for `KFP standalone deployment`, while `gke distribution` is for `full fledged kubeflow deployment`. So it will be easier if we separate this PR into two PRs:
1. This pipelines version file.
2. The rest of files under `gke/` path in this PR.

But it is only for reference, we don't need to take action to split this PR here.

content/en/docs/distributions/gke/deploy/deploy-cli.md
zijianjoy: https://cloud.google.com/sdk/gcloud/reference/container/clusters/get-credentials

layouts/shortcodes/pipelines/latest-version.html
gkcalat: SGTM. Thanks!

content/en/docs/distributions/gke/deploy/deploy-cli.md
gkcalat: Fixed

zijianjoy:
/lgtm
/approve

Thank you Ablai for updating the documentations!

PR #3317, opened 2022-08-08
add kfp v2 quickstart

connor-mccarthy:
Some outstanding content, such as cross-section links, to be added once merged into the `pipelines-v2-docs` branch.

PR #3320, opened 2022-08-11
add kfp v2 authoring/components

connor-mccarthy:


connor-mccarthy:
/cc @chensun
for content review

connor-mccarthy:
/cc @zichuan-scott-xu
for content review

content/en/docs/components/pipelines/author-a-pipeline/components.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Compile (save) a component
+Once you've written a component, you may wish to write the component definition to YAML for future use or submission for execution. This can be done via the KFP SDK DSL compiler:
+
+```python
+from kfp import compiler
+
+compiler.Compiler().compile(pipeline_func=addition_component, package_path='addition_component.yaml')
+```
+
```
zichuan-scott-xu, line 204: The previous doc contains a section of submitting a pipeline run via a client. I think it may help to keep that section for the reader to try creating their components themselves. WDYT? 
e.g. `client.create_run_from_pipeline_func()`

content/en/docs/components/pipelines/author-a-pipeline/components.md
```diff
@@ -0,0 +1,222 @@
...
+Often times, the output artifact from one task is an input artifact to another task. In this case, the artifact can be easily accessed from the upstream task using `my_task.outputs['artifact_name']`. Additionally, the artifact is already registered in MLMD.
+
+But what if you wish to use an existing artifact that is not generated by a task in the current pipeline? Or wish to use as an artifact an external resource that was not generated by a pipeline at all? In either case, you would use an importer component to load an artifact from its URI.
+
+You do not need to write an importer component; it can be imported from the `dsl` module and used directly:
+
+```python
+from kfp import dsl
+
+@dsl.pipeline():
```
zichuan-scott-xu, line 184: nit: remove `:`

content/en/docs/components/pipelines/author-a-pipeline/components.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+The following is an example of a lightweight component that trains a model on an existing input `Dataset` artifact for `num_epochs` epochs, then saves the output `Model` artifact.
+
+```python
+from kfp import dsl
+
+@dsl.component(
+    base_image='python:3.7',
+    packages_to_install=['tensorflow']
```
zichuan-scott-xu, line 55: nit: may consider specifying a version of packages to install (`tensorflow==1.3.0`) and mentioning that these are pip installed.

content/en/docs/components/pipelines/author-a-pipeline/components.md
zichuan-scott-xu: Following on this, both of the above sample `train_model` components in the previous section have two inputs and an output. Maybe we can showcase how to call these components when submitting a run, i.e. only passing the inputs to the components but not the annotated outputs. This may be confusing to the beginners, but it could also be placed in the documentation about data passing.

zichuan-scott-xu (commented):
Thanks @connor-mccarthy ! The documentation looks very clear and concise to me and I just have some small suggestions.

content/en/docs/components/pipelines/author-a-pipeline/components.md
connor-mccarthy: > The previous doc contains a section of submitting a pipeline run via a client.
I think it may make sense to include this in the "Submit a Pipeline Run" authoring section. It is also covered in the "Quickstart" section (PR #3317). WDYT?

> Maybe we can showcase how to call these components when submitting a run, i.e. only passing the inputs to the components but not the annotated outputs.

Good point on this and we definitely want to document this. I was planning on covering this in an authoring section called "Tasks" which describes how to instantiate and use tasks (PR #3322). WDYT?

content/en/docs/components/pipelines/author-a-pipeline/components.md
connor-mccarthy: Good catch

content/en/docs/components/pipelines/author-a-pipeline/components.md
zichuan-scott-xu: Thanks for clarifying! I agree that we should separate content like this and create a more linear reading experiences.

content/en/docs/components/pipelines/author-a-pipeline/components.md
```diff
@@ -0,0 +1,225 @@
...
+To create a lightweight component, you must:
+1. Define a standalone function.
+
+    A standalone Python function is a function that does not reference any symbols defined outside of its scope. This means the function must define all objects it uses within the scope of the function and must include all import statements within the function body.
+
+2. Include type annotations for the function parameters and return values.
+
+    Type annotations indicate what the component inputs and outputs are and tells the KFP lightweight component executor how to serialize and deserialize the data as it is passed within a pipeline. This also (optionally) allows the KFP DSL compiler to type check your pipeline.
+
+    Valid parameter annotations include `str`, `int`, `float`, `bool`, `dict`, `list`, `OutputPath`, `InputPath`, `Input[<Artifact>]`, and `Output[<Artifact>]`.
```
chensun, line 35: We also support `InputUri` and `OutputUri`, but didn't advertise these as we prefer `Input` and `Output` which covers both path and uri.

content/en/docs/components/pipelines/author-a-pipeline/components.md
```diff
@@ -0,0 +1,225 @@
...
+    # train for num_epochs
+    my_model.fit(x, y, epochs=num_epochs)
+    
+    # save the Model artifact
+    my_model.save(model.path)
+```
+
+Notice the `base_image` argument to the `@kfp.dsl.component` decorator. Despite not having the word "container" in its name, lightweight components are still executed as a container at runtime. The `@kfp.dsl.component` decorator mereley provides a convient Pythonic interface to defining this container image, command, and arguments. [`python:3.7`][python-docker-image] is the default image, but can be changed to any image accessible to the executing backend, as long as the image has a Python interpreter available as `python3`. Packages in `packages_to_install` will be pip installed at container runtime.
+
+**When to use?** Lightweight components should be used any time your component implementation can be written as a standalone Python function and does not require an abundance of source code.
```
chensun, line 86: Lightweight component is preferred for quick demo, notebook use. 
For production usage, more importantly components to be shared and reused, we recommend containerize the implementation even if it'w written by Python.

https://docs.google.com/presentation/d/1uHArEC7eMgZMUdy_V-h2mHTiU1N5Z5JA48-OEbkGs3g/edit?resourcekey=0-HrhC7BsFkEGurXi0c02fOA#slide=id.gb7fca233c8_0_73p

content/en/docs/components/pipelines/author-a-pipeline/components.md
connor-mccarthy: I know we have these placeholders, but do we have an annotation to indicate these? Or are they only accessible via the `artifact.uri` notation in container components?

content/en/docs/components/pipelines/author-a-pipeline/components.md
connor-mccarthy: > can be written as a standalone Python function and does not require an abundance of source code.

Does this capture the underlying rationale for why lightweight components are preferred for quick demo / notebook use? Or is there another reason? My sense is that if there is very minimal source code and the component function is simple that there isn't an advantage to containerizing it (and may actually be a disadvantage b/c it (a) adds a step to the component creation process and (2) adds another artifact required to use the component (private image + component YAML v just component YAML). Is there an advantage I'm not thinking of?

Either way, the suggested usage is helpful and I will add to this file.

content/en/docs/components/pipelines/author-a-pipeline/components.md
chensun: We have those available to use as placeholders in v1 component yaml: https://github.com/kubeflow/pipelines/blob/f17e9625dbe0cfad1e7a7e87fa7cccc79e08c2ac/sdk/python/kfp/v2/compiler_cli_tests/test_data/two_step_pipeline.py#L53

but not as annotations--change was prepared but discarded: https://github.com/kubeflow/pipelines/pull/5634

So I think maybe we can skip these in the doc as well, for now.

chensun (approved):
/lgtm
/approve

Thanks!

PR #3326, opened 2022-08-17
add kfp v2 compile a pipeline

connor-mccarthy:

### DIFF
--- a/content/en/docs/components/pipelines/compile-a-pipeline.md
+++ b/content/en/docs/components/pipelines/compile-a-pipeline.md
@@ -1,0 +1,123 @@
++++
+title = "Compile a Pipeline"
+description = "Compile a pipeline definition to YAML"
+weight = 5
++++
+
+While pipelines and components are authored in Python, they can be compiled to intermediate representation (IR) YAML ([example][compiled-output-example]).
+
+A YAML pipeline/component definition preserves a static representation of your pipeline/component. This YAML can be submitted to the KFP backend for execution or deserialized by the KFP SDK for integration into another pipeline.
+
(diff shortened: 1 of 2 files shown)


connor-mccarthy:
/cc @chensun
for content review

content/en/docs/components/pipelines/compile-a-pipeline.md
```diff
@@ -0,0 +1,123 @@
...
+
+```python
+cmplr.compile(addition_component, package_path='addition_component.yaml')
+```
+
+The `Compiler.compile` method accepts a few optional additional parameters:
+
+**`pipeline_name`** (`string`)
+
+Sets the name of the pipeline (or component). This is written to IR as the  `pipelineInfo.name` field. Will override the `name` passed to the `@dsl.pipeline` decorator.
```
chensun, line 45: nit: mention this name is used a pipeline context and explains a bit what is pipeline context?

connor-mccarthy: By pipeline context, are you referring to the pipeline context manager we use to create pipeline spec? If so, isn't that largely an implementation detail and not something we need to explain to authors? If not, can let me know what you are referring to?

chensun: No, pipeline context is maybe a MLMD concept. It identifies a number of runs originated from the same pipeline definition. I think it provides a query scope for some features like caching and importer. Desmond and/or Jialu may provide better explanation on this.

chensun (commented):
/lgtm
/approve

Thanks!

PR #3327, opened 2022-08-17
add kfp v2 run a pipeline

connor-mccarthy:


connor-mccarthy:
/cc @chensun
for content review

content/en/docs/components/pipelines/run-a-pipeline.md
```diff
@@ -0,0 +1,80 @@
++++
+title = "Run a Pipeline"
+description = "Execute a pipeline on the KFP backend"
+weight = 6
++++
+
+The KFP SDK offers three ways to run a pipeline.
+
+## 1. Run from the KFP Dashboard
```
chensun, line 9: If KFP dashboard the official term to reference that UI? I often refer to it as KFP Pipeline templates page. Not sure if we have a formal name for it.

content/en/docs/components/pipelines/run-a-pipeline.md
```diff
@@ -0,0 +1,80 @@
...
+The first and easiest way to run a pipeline is by submitting it via the KFP dashboard.
```
chensun, line 10: Probably the easiest way is to run one of the preloaded pipeline samples that come with the deployment.

content/en/docs/components/pipelines/run-a-pipeline.md
connor-mccarthy: I think we refer to the FE in several ways. I think that that the v2 docs is a good opportunity to unify around one term we prefer. What do you prefer?

content/en/docs/components/pipelines/run-a-pipeline.md
connor-mccarthy: Agreed, though with this page I intended to lay out general instructions for running _any_ pipeline, not just _a_ pipeline. I think it makes sense to include at least what is here, but do you think it's worth including how to run one of the preloaded pipelines also? It's slightly captured from step 4 and 5, but not explicitly. WDYT?

chensun: Make sense, preloaded samples are special, we can skip it in this context.

content/en/docs/components/pipelines/run-a-pipeline.md
chensun: I don't have a strong preference. Dashboard sounds good to me. Like to see others' opinion. 
/cc @zijianjoy

connor-mccarthy: Merging, but we can change this easily if other opinions emerge

connor-mccarthy:
@chensun, thank you for the lgtm. Can you approve? I am not an OWNER.

PR #3330, opened 2022-08-19
add kfp v2 component i/o

connor-mccarthy:

### DIFF
--- a/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
+++ b/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
@@ -69,7 +69,7 @@ Ultimately, each authoring style creates a component definitied by an `image`, `
 There is one additional type of input, `PipelineTaskFinalStatus`, which allows access to the metadata of one task from within another. This input is a special case, as it is neither a parameter nor an artifact. Use of this input is covered in [Authoring: Pipelines][pipelines].
 
 ### Input parameters
-Input parameters are declared when you use a `str`, `int`, `float`, `bool`, `dict` or `list` type annotation. The data passed to parameters typed with `dict` or `list` may only container JSON-serializable Python primitives. Inner types to `dict` and `list`, such as `Dict[str, str]` or `List[int]` are not respected. `Union` types are not permitted.
+Input parameters are declared when you use a `str`, `int`, `float`, `bool`, `dict` or `list` type annotation. The data passed to parameters typed with `dict` or `list` may only container JSON-serializable Python primitives. `Union` types are not permitted.
 
 In the example `create_dataset` component, `initial_text` is an input parameter. In `augment_dataset`, `text` and `num` are input parameters.
 
--- a/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
+++ b/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
@@ -203,11 +203,11 @@ Let's introduce two more components for sake of demonstrating passing parameters
 from typing import NamedTuple
 
 @dsl.component
-def named_tuple(an_id: int) -> NamedTuple('Output', [('name', str), ('id', int)]):
+def named_tuple(an_id: int) -> NamedTuple('Outputs', [('name', str), ('id', int)]):
     """Lightweight Python component with a NamedTuple output."""
     from typing import NamedTuple
-    output = NamedTuple('Output', [('name', str), ('id', int)])
-    return output('my_dataset', an_id)
+    outputs = NamedTuple('Outputs', [('name', str), ('id', int)])
+    return outputs('my_dataset', an_id)
 
 @dsl.container_component
 def identity_container(integer: int, output_int: OutputPath(int)):
--- a/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
+++ b/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
@@ -1,5 +1,5 @@
 +++
-title = "Task I/O"
+title = "Component I/O"
 description = "Use parameter/artifact inputs and outputs"
 weight = 3                 
 +++
--- a/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
+++ b/content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
@@ -105,8 +105,8 @@ In our example, `augment_dataset` has a one integer output.
 You may also specify multiple output parameters by using these annotations within a `typing.NamedTuple` as follows:
 
 ```python
-from kfp import dsl
 import typing
+from kfp import dsl
 
 @dsl.component
 def my_component() -> typing.NamedTuple('Outputs', [('name', str), ('id', int)]):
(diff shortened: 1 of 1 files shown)


connor-mccarthy:
/cc @chensun
for content review

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
++++
+title = "Task I/O"
```
chensun, line 2: Component I/O ?

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Inputs
+Component inputs are specified by the component function's signature. This applies for all authoring approaches: [lightweight Python components][lightweight-python-component], [containerized Python components][containerized-python-component], and [container components][container-component].
+
+Ultimately, each authoring style creates a component definitied by an `image`, `command`, and `args`. When you use an input, it is represented as a placeholder in the `command` or `args` and is interpolated at component runtime.
+
+There is one additional type of input, `PipelineTaskFinalStatus`, which allows access to the metadata of one task from within another. This input is a special case, as it is neither a parameter nor an artifact. Use of this input is covered in [Authoring: Pipelines][pipelines].
+
+### Input parameters
+Input parameters are declared when you use a `str`, `int`, `float`, `bool`, `dict` or `list` type annotation. The data passed to parameters typed with `dict` or `list` may only container JSON-serializable Python primitives. Inner types to `dict` and `list`, such as `Dict[str, str]` or `List[int]` are not respected. `Union` types are not permitted.
```
chensun, line 72: > Inner types to `dict` and `list`, such as `Dict[str, str]` or `List[int]` are not respected.

IIRC, in some cases they are even required.
https://github.com/kubeflow/pipelines/blob/efb96135033fc6e6e55078d33814c45a98566e68/sdk/python/kfp/compiler/test_data/pipelines/pipeline_with_loops.py#L23

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
...
+
+Let's introduce two more components for sake of demonstrating passing parameters between components:
+```python
+from typing import NamedTuple
+
+@dsl.component
+def named_tuple(an_id: int) -> NamedTuple('Output', [('name', str), ('id', int)]):
+    """Lightweight Python component with a NamedTuple output."""
+    from typing import NamedTuple
+    output = NamedTuple('Output', [('name', str), ('id', int)])
```
chensun, line 209: nit: convention is `NamedTuple('Outputs', ...)` as there are multiple.

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
connor-mccarthy: Good though. That's probably the correct name here. I will update.

In general, it's sometimes a bit ambiguous whether something ought to be called a component or a task in the docs. For example, we often talk about "component run time", but this is more accurate as "task run time", IMO. Usually it's fairly inconsequential.

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
connor-mccarthy: You are correct. Updated.

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+This pipeline uses a container component `create_dataset` to construct an initial `Dataset` artifact containing `initial_text`. Then, the downstream lightweight Python component `augment_dataset` appends `text` repeated `num` times to the dataset and saves it as a new dataset.
+
+## Inputs
+Component inputs are specified by the component function's signature. This applies for all authoring approaches: [lightweight Python components][lightweight-python-component], [containerized Python components][containerized-python-component], and [container components][container-component].
+
+Ultimately, each authoring style creates a component definitied by an `image`, `command`, and `args`. When you use an input, it is represented as a placeholder in the `command` or `args` and is interpolated at component runtime.
+
+There is one additional type of input, `PipelineTaskFinalStatus`, which allows access to the metadata of one task from within another. This input is a special case, as it is neither a parameter nor an artifact. Use of this input is covered in [Authoring: Pipelines][pipelines].
```
chensun, line 69: Technically, `PipelineTaskFinalStatus` is a struct typed parameter with system provided value. And it's only usable in an exit handler scenario. 
But feel free to refer to other doc for more details on this.

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
...
+In our example, `augment_dataset` has a one integer output.
+
+You may also specify multiple output parameters by using these annotations within a `typing.NamedTuple` as follows:
+
+```python
+from kfp import dsl
+from typing import NamedTuple
+
+@dsl.component
+def my_component() -> NamedTuple('Output', [('name', str), ('id', int)]):
```
chensun, line 112: nit: Output -> Outputs

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
```diff
@@ -0,0 +1,349 @@
...
+    from typing import NamedTuple
+
+    output = NamedTuple('Output', [('name', str), ('id', int)])
```
chensun, line 115: this should be `collections.namedtuple`, right?

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
connor-mccarthy: Done

content/en/docs/components/pipelines/author-a-pipeline/inputs-outputs-data-passing.md
connor-mccarthy: Both work, but I prefer `typing.NamedTuple`. Rationale:

All of the examples that I've seem use `typing.NamedTuple` in the annotation then `collections.namedtuple` as the true return type. We request this in our [docs](https://www.kubeflow.org/docs/components/pipelines/sdk/python-function-components/#building-python-function-based-components), too. From my testing, the `typing.NamedTuple` annotation is required, but the return return type can be a `collections` or a `typing` named tuple.

I suspect the examples return a `collections` namedtuple because it's more minimal, not requiring annotations in the `typing.NamedTuple` class factory function, but I think that the confusion of using two different types of named tuples outweighs this "benefit". Furthermore, I'm not convinced that the minimal quality of `collections.namedtuple` is a benefit, as I think using types in both function calls is better for emphasizing the strongly typed and strongly coupled nature of the named tuple class instances. I think these docs are a good opportunity to present a more internally consistent authoring approach. WDYT?

connor-mccarthy:
/hold
until add content about pipeline as task

chensun (commented):
/lgtm
/approve

Thanks!

PR #3363, opened 2022-09-15
fix pipelines file location

connor-mccarthy:
Fixes a file that was missed during the recent migration.


chensun (approved):
/lgtm
/approve

PR #3368, opened 2022-09-26
add hint in case passlib is not installed

romeokienzler:


Arhell (commented):
/ok-to-test

varodrig:
/close due to content no longer valid; the page and content don't exist.
cc @romeokienzler @Arhell 
@hbelmiro

PR #3399, opened 2022-11-08
update release docs (dependency table + headings)

thesuperzapper:
This PR resolves the following issues:

- Previously it was not clear what the dependencies on the release pages actually referred to. They are now explicitly stated as being ONLY related to the `kubeflow/manifests`.
- Previously it was not clear that the dependencies on the release pages were not necessarily "hard" requirements (see [this `kubeflow-discuss` thread with a confused user](https://groups.google.com/g/kubeflow-discuss/c/NaFNdy3UixE/m/YsBav1K2BwAJ)). There is now a disclaimer that other versions might work, but are not validated by the Manifests WG. 
- Previously, the headers for each release version were slightly confusing.
- Dependencies were not listed for Kubeflow 1.5

## Screenshot of the updated "Kubeflow 1.6.1" dependency table:

Screenshot 2022-11-08 at 14 49 09

### DIFF
--- a/content/en/docs/releases/kubeflow-0.6.md
+++ b/content/en/docs/releases/kubeflow-0.6.md
@@ -4,7 +4,7 @@ description = "Information about the Kubeflow 0.6 release"
 weight = 106
 +++
 
-## 0.6.1
+## Kubeflow 0.6.1
 
 <div class="table-responsive">
 <table class="table table-bordered">
@@ -48,7 +48,7 @@ weight = 106
 </table>
 </div>
 
-## 0.6.0
+## Kubeflow 0.6.0
 
 <div class="table-responsive">
 <table class="table table-bordered">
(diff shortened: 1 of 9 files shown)


thesuperzapper:
@zijianjoy and @annajung I appreciate a quick look at this PR, it resolves a confusion that users are having around the "dependencies" table.

annajung:
Thanks for adding additional clarification to the dependency table, changes look good to me
Holding for a few days for @zijianjoy to review as well. Feel free to unhold 

/lgtm
/hold

zijianjoy:
/lgtm
/approve

Thank you for the change!


NIT: I think large table is easier to edit if we use this syntax in markdown file: https://www.markdownguide.org/extended-syntax/#tables. But I don't have strong opinion.

thesuperzapper:
@zijianjoy thanks, the reason I can't use the markdown syntax is that the table styling won't be applied, but more importantly, I wouldn't be able to do things like `rowspan`, which allows me to make a cell take up multiple rows.

/unhold

PR #3416, opened 2022-12-14
Pass in `pipeline_fn` instead of `"pipeline.yaml"` to `client.create_run_from_pipeline_func()`

paulcrowley:


varodrig:
/close due to content no longer valid; the page and content don't exist. cc @hbelmiro 
@Arhell @paulcrowley

PR #3427, opened 2023-01-13
Switched the case of the heading

sen-sam:
Switched the case of the heading to make it consistent with the rest of the documentation.


Arhell (commented):
/ok-to-test

varodrig:
/close due to PR no longer valid; the page don't exist, but content updated.
cc @hbelmiro @Arhell @sen-sam

PR #3432, opened 2023-01-21
[Training] Add the guide on how to use Scheduler Plugins with coscheduling

tenzen-y:
Signed-off-by: Yuki Iwai <yuki.iwai.tz@gmail.com>

I added the guide on how to use Scheduler Plugins with coscheduling.


johnugeorge:
Thanks for updating docs

/lgtm
/approve

PR #3434, opened 2023-01-23
Add create kubeflow-user-example-com ns

strangiato:
When following the current instructions the install throws the following error message:

```
Error from server (NotFound): error when creating "STDIN": namespaces "kubeflow-user-example-com" not found
```

The kubeflow-user-example-com ns should be created before the install to prevent this error.


Arhell (commented):
Please sign CLA

varodrig:
/close 
due to content no longer valid; the page and content don't exist.
cc @hbelmiro @Arhell @strangiato

PR #3435, opened 2023-01-25
[Katib] Change the Default Resume Policy

andreyvelich:
Related: https://github.com/kubeflow/katib/pull/2102.
We changed the default resume policy to `Never` for Katib Experiments.
Also, I used commit for APIs links so we always have correct reference.

/assign @johnugeorge @tenzen-y


terrytangyuan (approved):
/lgtm

PR #3436, opened 2023-01-25
[Katib] Limit verbs when use a new CRD in Trials

andreyvelich:
After this PR: https://github.com/kubeflow/katib/pull/2091, we use scoped verbs while adding a new CRD to the Katib Controller RBAC.

/assign @johnugeorge @tenzen-y


tenzen-y (commented):
Thanks for updates!
/lgtm

PR #3462, opened 2023-03-14
Remove "Pipeline metrics" page

jlyaoyuli:
In KFP v2 beta0 and above, we are deprecating the `runMetrics `. To avoid the misunderstanding, we will need to remove the section here.


content/en/docs/components/pipelines/v1/sdk/pipelines-metrics.md
```diff
@@ -120,23 +120,6 @@ Refer to the [full example](https://github.com/kubeflow/pipelines/blob/sdk/relea
 * `numberValue` must be a numeric value.
 * `format` can only be `PERCENTAGE`, `RAW`, or not set.
```
zijianjoy, line 122: We will probably need to remove more than just the UI section. 

@chensun Can we remove the whole page here because we are going to remove V1 support of runMetrics in https://github.com/kubeflow/pipelines/pull/8963?

zijianjoy: Note: User can always navigate to older version of kubeflow.org to refer to this page, if they are using older version of KFP.

chensun: Removing this would only affect version `master` right?
image

I wonder if we should wait until Kubeflow 1.7 is out, because this feature is technically available in Kubeflow 1.7, and removing it from master would likely "remove" it from v1.7 doc.

zijianjoy: You are right about removing it from master version. The old versions will still have this instruction. 

If we don't plan to support this feature in KFPv2, it is probably better to remove it sooner than later. We are receiving github issues asking about it, and we have to keep telling them not to use it. It is better to prevent users from depending on a soon-be-deprecated feature from the documentation.

chensun: That's reasonable. I'm fine with removing either this section or the whole page now.

chensun (commented):
/lgtm

chensun (approved):
/lgtm

strangiato:
Is there something replacing runMetrics that we should be migrating to?  Is there documentation for how to utilize that feature?

PR #3488, opened 2023-04-04
add kfp v2 keywords to pipelines documentation

connor-mccarthy:


connor-mccarthy:
@zijianjoy, do you have a sense of whether this will improve the search relevance of KFP v2 documentation? Currently the [existing (soon-to-be old) KFP v2 documentation](https://www.kubeflow.org/docs/components/pipelines/v1/sdk-v2/v2-compatibility/) ranks highly. We'd like these new docs to rank as highly as possible.

zijianjoy:
@connor-mccarthy We can follow similar pattern like Distribution in https://www.kubeflow.org/docs/distributions/aws/. We clean up the old doc's content but keep a redirect link to the new V2 doc location.

In that way you don't need to add keyword in all v2 pages.

connor-mccarthy:
@zijianjoy, can you explain a bit more? My sense is that we want to keep both the v1 and v2 docs. We can make the v2 docs primary and link to v1, but would this affect the search ranking?

zijianjoy:
My understanding is that this documentation is about `V2 compatible mode`, not `V2 mode`: https://www.kubeflow.org/docs/components/pipelines/v1/sdk-v2/v2-compatibility/.

Therefore, I think deleting all or most the content in v2 compatible mode is preferred. Because we don't want to guide users to a maze where we have v1, v2, v2 compatible. The current content is a heavy learning curve for users. If there is anything we want to keep in V2 compatible mode, let's try to move them to the new v2 doc. 

Above is only my thought process, I will let @chensun to make the decision.

connor-mccarthy:
Want to note that we also want to demote the v1 SDK v2 namespace docs (not compatible mode), but probably cannot delete them, @zijianjoy.

chensun (approved):
/lgtm
/approve

PR #3496, opened 2023-04-13
add platform-specific features docs

connor-mccarthy:


content/en/docs/components/pipelines/v2/platform-specific-features.md
```diff
@@ -0,0 +1,119 @@
...
+
+For cases where features are not portable across platforms, users may author pipelines with platform-specific functionality via KFP SDK platform-specific plugin libraries.
+
+In general, platform-specific plugin libraries provide functions that act on tasks similarly to [task-level configuration methods][task-level-config-methods] provided by the KFP SDK directly. Platform-specific plugin libraries may also provide pre-baked components.
+
+
+
+<!-- TODO: add docs on how to create a platform-specific authoring library -->
+
+## Example: Read/write to a Kubernetes PVC using kfp-kubernetes
```
zijianjoy, line 18: This example should suffice as a standalone page like "Read/Write to a Persistent Volume" to increase searchability. It is because people will search for PVC/PV first before they realize that Platform specific flag is for PVC read/write.

connor-mccarthy: What do you think about enumerating the Kubernetes-specific features on this page for easier in-website search/discoverability?

My sense we want to emphasize the platform-specific pattern in the general case through this specific example, without promoting this one example to a top-level content item. IMO, it would be confusing to omit the other Kubernetes-specific features and my sense is we don't want to include examples for all of them, as it would be duplicative of the [`kfp-kubernetes` reference docs](https://kfp-kubernetes.readthedocs.io/en/kfp-kubernetes-0.0.1/). WDYT?

zijianjoy: We can do it later because it is out-of-scope of this PR, but we should make the functionality itself as title of a documentation page to improve search/discoverability. When a user wants to use Persistent Volume, they won't look it up in this kubernetes-specific page first, they will browse the table of context. Therefore using page title to describe actions that user desire is the common practice.

PR #3530, opened 2023-06-20
Notebooks: Rename "Notebook" to "Workbench" in web documentation

tzstoyanov:
Kubeflow notebooks were initially created from JupyterLab Notebook. That functionality evolved, now other web-based development environments can be used as well - VSCode, RStudio and others can be added in future. The term "notebook" is a bit misleading and does not reflect all recent changes and enhancements. That's why a proposal was made to replace it with the term "workbench" - issue 6981.
This patch renames "notebook" to "workbench" in the Kubeflow web documentation.

issue: https://github.com/kubeflow/kubeflow/issues/6981


Arhell (commented):
/ok-to-test

thesuperzapper:
@juliusvonkohout @tzstoyanov I strongly believe that we should NOT do a rename unless there is a technical change, as that would just be confusing to users who already know about "Kubeflow Notebooks", which is effectively a brand at this point rather than a description.

/hold

varodrig:
@thesuperzapper  @juliusvonkohout @tzstoyanov can you provide an update on this PR?
should we close it?

thesuperzapper:
> @thesuperzapper @juliusvonkohout @tzstoyanov can you provide an update on this PR? should we close it?

@varodrig yep, especially with the [Notebooks 2.0](https://github.com/kubeflow/notebooks/issues/85) work we will eventually use the term "Kubeflow Workspaces", after a few releases, but may never fully drop the "Kubeflow Notebooks" branding given its industry prevalence. 

/close

PR #3533, opened 2023-06-21
kfp: Add version compatibility matrix doc

chensun:
Closes: https://github.com/kubeflow/pipelines/issues/8728

### DIFF
--- a/content/en/docs/components/pipelines/v2/version-compatibility.md
+++ b/content/en/docs/components/pipelines/v2/version-compatibility.md
@@ -6,15 +6,25 @@ weight = 80
 
 {{% kfp-v2-keywords %}}
 
-The following table presents a comprehensive overview of the version compatibility between Kubeflow Pipelines (KFP) Runtime and KFP SDK.
+The following table presents a comprehensive overview of the version compatibility between the Kubeflow Pipelines (KFP) Runtime and the KFP SDK.
 
 | KFP Runtime | KFP SDK | Notes |
 |---|---|---|
 | v2.0.* | v2.0.* | Active development. Support for certain features may be staged between the Runtime and the SDK. |
-| v2.0.* | v1.8.* | Backward compatibility maintained. No v2 features support. |
-| v1.8.* | v1.8.* | Maintenance mode. Full compatibility for features available from SDK v1 namespace. Deprecated support for features from SDK v2 namespace. |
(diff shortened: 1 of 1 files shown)


chensun:
Preview: https://deploy-preview-3533--competent-brattain-de2d6d.netlify.app/docs/components/pipelines/v2/version-compatibility/

content/en/docs/components/pipelines/v2/version-compatibility.md
```diff
@@ -0,0 +1,27 @@
...
++++
+
+{{% kfp-v2-keywords %}}
+
+The following table presents a comprehensive overview of the version compatibility between Kubeflow Pipelines (KFP) Runtime and KFP SDK.
+
+| KFP Runtime | KFP SDK | Notes |
+|---|---|---|
+| v2.0.* | v2.0.* | Active development. Support for certain features may be staged between the Runtime and the SDK. |
+| v2.0.* | v1.8.* | Backward compatibility maintained. No v2 features support. |
```
connor-mccarthy, line 14: It's not entirely clear to me what is meant by backward compatibility here. It seems it could be within the KFP Runtime and KFP SDK individually, between them, or something else. Is there a way we can clarify this a bit?

Also, can you help me understand what is meant by "No v2 features support." This is confusing to me since features from the v1.8 SDK's `kfp.v2` namespace are supported by the v2.0 backend. I think I'm thinking about this the wrong way and perhaps other readers will too.

connor-mccarthy (commented):
Thank you for this!

content/en/docs/components/pipelines/v2/version-compatibility.md
chensun: Added some clarification notes, see if it helps. 

> This is confusing to me since features from the v1.8 SDK's kfp.v2 namespace are supported by the v2.0 backend.

They're not officially supported by v2.0 backend. Specifically, If they were compiled by v2 compiler, they would be IR in JSON format, and we disallow JSON in v2.0 backend. If they were compiled by v1 compiler using v2-compatible mode, they are Argo in YAML format, which is technically working on either v1 or v2 backend. But we have declared end of support for v2-compatible mode a long time ago. 
To avoid further confusing the users with all these details, I think simply put "they're not supported" is sufficient. WDYT?

connor-mccarthy (approved):
/lgtm
/approve

PR #3534, opened 2023-06-27
kfp: improve the doc on pipeline as component use case

chensun:
Fixes https://github.com/kubeflow/pipelines/issues/8726

### DIFF
--- a/content/en/docs/components/pipelines/v2/load-and-share-components.md
+++ b/content/en/docs/components/pipelines/v2/load-and-share-components.md
@@ -25,7 +25,7 @@ def my_pipeline():
 You can also load a component directly from a URL, such as a GitHub URL:
 
 ```python
-loaded_comp = components.load_component_from_url('https://github.com/kubeflow/pipelines/blob/984d8a039d2ff105ca6b21ab26be057b9552b51d/sdk/python/test_data/pipelines/two_step_pipeline.yaml')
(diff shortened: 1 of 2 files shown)


content/en/docs/components/pipelines/v2/load-and-share-components.md
```diff
@@ -25,7 +25,7 @@ def my_pipeline():
 You can also load a component directly from a URL, such as a GitHub URL:
 
 ```python
-loaded_comp = components.load_component_from_url('https://github.com/kubeflow/pipelines/blob/984d8a039d2ff105ca6b21ab26be057b9552b51d/sdk/python/test_data/pipelines/two_step_pipeline.yaml')
+loaded_comp = components.load_component_from_url('https://raw.githubusercontent.com/kubeflow/pipelines/2.0.0/sdk/python/test_data/components/add_numbers.yaml')
```
chensun, line 28: Note: it has to be the `https://raw.githubusercontent.com` url, otherwise the loading code would fail.

Linchin: Maybe use a permanent link for the raw file too? You can do that by accessing the permanent link of this file, then choose raw.

Linchin: Since the version tag is included in the URL, it should be stable.

PR #3419, opened 2023-01-03
Migrate GKE docs

gkcalat:
- Added a link to the external website
- Added ignoreFiles with the list of migrated docs 
- Updated description for the landing page of Kubeflow of Google Cloud

Issue #3074, opened 2021-11-23
TRACKING - remove distribution owned docs from website


thesuperzapper:
_This issue was spawned from the discussions raised in [PR #3004](https://github.com/kubeflow/website/pull/3004)._

---

Some time ago, the community agreed that distributions would ["be owned and maintained outside of Kubeflow"](https://github.com/kubeflow/community/blob/master/proposals/kubeflow-distributions.md#ownership--development).

We have already [removed distribution-owned YAML](https://github.com/kubeflow/manifests/issues/1904) from `kubeflow/manifests`, we should now also remove the [distribution-owned docs](https://www.kubeflow.org/docs/distributions/) from `kubeflow.org`.

_NOTE: It is important that we get started on this soon to minimize disruption of the [“Certified Kubeflow”](https://github.com/kubeflow/community/blob/master/proposals/kubeflow-conformance-program-proposal.md#certified-kubeflow-for-kubeflow-distribution) program. 
As once this program is implemented, distributions will need to be certified if they wish to be linked on `kubeflow.org`._

## Proposed Plan:

1. Update the ["Installing Kubeflow"](https://www.kubeflow.org/docs/started/installing-kubeflow/) table with links to each distribution's nominated "external website":
    - (Most active distributions already have their own "external website" or GitHub repo that can serve this purpose)
2. Remove the [distribution-owned docs](https://www.kubeflow.org/docs/distributions/) from kubeflow.org:
    - (Target __February 1st, 2022 (70 days)__ for removing any remaining distribution docs)
    - (We can prevent 404 errors by redirecting existing docs pages to their "external websites")
    - (As ["Kubeflow on AWS"](https://www.kubeflow.org/docs/distributions/aws/) and ["Kubeflow on Azure"](https://www.kubeflow.org/docs/distributions/azure/) are abandoned, and thus unlikely to set up an "external website", we can move the still-relevant information into the [./docs/ folder](https://github.com/kubeflow/manifests/tree/master/docs) of the `kubeflow/manifests` repo, and update the `kubeflow/manifests` [installation guide](https://github.com/kubeflow/manifests#installation) with links to this)

## Tracking Table:

| Name | Docs Path | OWNERS | PR to Remove | Status |
| --- | --- | --- | --- | --- |
| Kubeflow on AWS | [docs/distributions/aws](https://github.com/kubeflow/website/tree/master/content/en/docs/distributions/aws) | @Jeffwan @PatrickXYS |  #3233   |  ✅  |
[...]

PR #3548, opened 2023-07-13
remove shadowed gke distribution docs

thesuperzapper:
This PR removes the files that contained the old GKE docs, these pages are no longer accessible because all pages starting with [`/docs/distributions/gke/*` are redirected to the official GKE Docs](https://github.com/kubeflow/website/blob/386621ea6c862ff2a37bac2f061d894ee1724f71/content/en/_redirects#L205) after PR #3419.

Part of https://github.com/kubeflow/website/issues/3074


thesuperzapper:
@zijianjoy @james-jwu  this PR removes the GKE docs from the website (as the previous PR which added the redirect forgot to remove the actual files).

thesuperzapper:
@zijianjoy @james-jwu can you please LGTM / approve? 

This is an important repo cleanup, and these docs are not even accessible since we automatically redirect `https://www.kubeflow.org/docs/distributions/gke/*` to `https://googlecloudplatform.github.io/kubeflow-gke-docs/docs/`.

---

Also, you may want to change the redirect to always point at your homepage `https://googlecloudplatform.github.io/kubeflow-gke-docs/`, because there is no guarantee that your docs will always use the same structure as the old ones.

For example, right now, if you go to `https://www.kubeflow.org/docs/distributions/gke/EXAMPLE_PATH`, you get redirected to `https://googlecloudplatform.github.io/kubeflow-gke-docs/docs/EXAMPLE_PATH` (which is obviously a 404, but just as an example):

https://github.com/kubeflow/website/blob/c1a824356897434bd646ade21b18eb09a09ebf65/content/en/_redirects#L205

PR #3555, opened 2023-07-21
move istio docs from `external-addons` to `multi-tenancy` component

thesuperzapper:
This PR moves the `istio` docs from `external-addons` to the `multi-tenancy` component section.

This is a better location for these docs as istio is clearly not an external addon for Kubeflow, but is part of the multi-user feature.

### DIFF
--- a/content/en/_redirects
+++ b/content/en/_redirects
@@ -186,7 +186,6 @@ docs/started/requirements/                    /docs/started/getting-started/
 /docs/started/kubeflow-overview  /docs/started/architecture
 
 # move istio from external-add-ons to multi-tenancy component
-/docs/external-add-ons/istio/     /docs/components/multi-tenancy/istio
 /docs/external-add-ons/istio/*    /docs/components/multi-tenancy/istio
 
 # ===============
--- a/content/en/_redirects
+++ b/content/en/_redirects
@@ -185,6 +185,9 @@ docs/started/requirements/                    /docs/started/getting-started/
 /docs/about/kubeflow             /docs/started/introduction
 /docs/started/kubeflow-overview  /docs/started/architecture
 
+# move istio from external-add-ons to multi-tenancy component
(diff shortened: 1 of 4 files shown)


thesuperzapper:
@zijianjoy I would appreciate a review on this.

I think this small refactor will significantly reduce confusion, as `istio` is really not the same as the other `external-add-ons` and is mostly a part of the `multi-tenancy` "component" (which is really more of a feature... but I digress).

jbottum:
I agree with the general direction of moving istio into multi-tenancy and out of add-ons.     I did not look at all of the redirects in this part of the PR.  Are these all part of the requested Istio change? https://github.com/kubeflow/website/blob/8b4307c4a619276318333925b8d9ffbe2f65574b/content/en/_redirects.

thesuperzapper:
@jbottum, there should only be 2 changes to the redirects, which is what I see in https://github.com/kubeflow/website/pull/3555/files:

```
# move istio from external-add-ons to multi-tenancy component
/docs/external-add-ons/istio/     /docs/components/multi-tenancy/istio
/docs/external-add-ons/istio/*    /docs/components/multi-tenancy/istio
```

thesuperzapper:
@jbottum I did some testing, and you are right that we only need the single `/docs/external-add-ons/istio/* --> /docs/components/multi-tenancy/istio` redirect, as the `*` matches nothing as well.

thesuperzapper:
@james-jwu can we please merge this PR? Users are thinking that Istio is an optional part of Kubeflow (https://github.com/kubeflow/website/issues/3536#issuecomment-1737836946) because it currently sits under the `"external add-ons"` section.

This PR moves it under `"multi-tenancy"` (which is where it is used), and should reduce confusion.

zijianjoy:
/lgtm
/approve

Thank you @thesuperzapper for the change! I agree that istio is part of multi-tenancy dependency.

PR #3557, opened 2023-07-21
remove unused files

thesuperzapper:
This PR removes files that are no longer referenced anywhere on the site.

You can check the Netelify preview to verify the site looks correct.


thesuperzapper:
@jbottum @zijianjoy this is a pretty straightforward cleanup, but you will probably just have to trust that I was careful to note remove any files which are actually used!

PR #3559, opened 2023-07-22
update root OWNERS (remove inactive, add @thesuperzapper)

thesuperzapper:
This PR updates the root OWNERS to reflect the currently active and highly trusted maintainers of the website.

#### Regarding the `approvers`, the following actions are taken:

- Remove anyone who has not approved a PR in many years (or who has publicly left the project)
- Adds @thesuperzapper:
    -  _(NOTE: if adding me is too controversial with the remaining approvers, I am happy to make a separate PR for further discussion)_
- This leaves the following members:
    - @chensun
    - @james-jwu
    - @thesuperzapper
    - @zijianjoy

#### Regarding the `reviewers`, the following actions are taken:

- Aligns them exactly with the `approvers`:
    - This is NOT a reflection on anyone who was removed, it's just that there is no practical reason to have root `reviewers` who are separate from the `approvers`.
    - This ensures PRs that make changes to something that ONLY a root approver can approve are automatically assigned to the root approvers for review. 
    - _(NOTE: root `reviewers` are only assigned by the bot when a file does not have a more specific OWNER reviewer)_

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,18 +1,10 @@
 approvers:
-  - animeshsingh
   - chensun
   - james-jwu
-  - knkski
-  - shannonbradshaw
+  - thesuperzapper
   - zijianjoy
 reviewers:
-  - 8bitmp3
-  - aronchick
-  - berndverst
-  - dansanche
-  - dsdinter
-  - Jeffwan
-  - jinchihe
-  - nickchase
-  - pdmack
-  - terrytangyuan
+  - chensun
+  - james-jwu
+  - thesuperzapper
+  - zijianjoy


thesuperzapper:
@zijianjoy @chensun @james-jwu as you are currently the only active root owners, one of you will need to approve this.

Happy to discuss this further in the next community meeting.

terrytangyuan (commented):
I don't understand the intention of this change. What is the definition of "approver" and "reviewer" here?

OWNERS
```diff
@@ -1,18 +1,10 @@
 approvers:
-  - animeshsingh
   - chensun
   - james-jwu
-  - knkski
-  - shannonbradshaw
+  - thesuperzapper
   - zijianjoy
 reviewers:
```
terrytangyuan, line 6: We need a clear definition of this. I feel like this should be the same as approvers at root level since we already have repo/component level OWNERs.

OWNERS
```diff
@@ -1,18 +1,10 @@
 approvers:
```
terrytangyuan, line 1: This should be org owners/governing board, IMO.

chensun:
I suggest we don't add new approvers at this point until we finalize the governance structure. 

/hold

OWNERS
thesuperzapper: I am a little confused about what you are saying here.

This PR is specifically about the root OWNERS for the repo, which are a "fallback" if a more specific reviewer/approver is not defined for a PR.

The main issue is that the websites' structure almost always requires a root approver due to configs being at relatively top-level folders (e.g. redirects, and git-submodules for the docsy theme).

OWNERS
thesuperzapper: For many repos, that would be sensible, but for the website, many standard changes, including updating redirects and applying security updates require a top-level approver given how the folders are required to be structured.

PR #3565, opened 2023-07-28
update kfp sdk "dex authentication" example

thesuperzapper:
This PR updates the `"Connect the Pipelines SDK to Kubeflow Pipelines"` dex example, to be more robust and use a "class constructor" pattern.

This makes it easier for users to automatically re-authenticate when their session expires, as they can simply call the `get_kfp_client()` method on the already initialized `KFPClientManager` class.

Additionally, it allows the creation of `kfp.Client()` instances that ignore self-signed SSL certificate errors, which is useful for development environments.

### DIFF
--- a/content/en/docs/components/pipelines/v1/sdk/connect-api.md
+++ b/content/en/docs/components/pipelines/v1/sdk/connect-api.md
@@ -152,146 +152,186 @@ and give them access to submit pipelines in their own namespace.
 <summary>Click to expand</summary>
 <hr>
 
-The process to authenticate the Pipelines SDK from outside the cluster in multi-user mode will vary by distribution:
+When running the Kubeflow Pipelines SDK from outside a multi-user Kubeflow cluster (e.g. on a laptop), the `kfp.Client()` will need to be initialized with credentials.
 
-* [Kubeflow on Google Cloud](/docs/distributions/gke/pipelines/authentication-sdk/#connecting-to-kubeflow-pipelines-in-a-full-kubeflow-deployment)
-* [Kubeflow on AWS](/docs/distributions/aws/pipeline/#authenticate-kubeflow-pipeline-using-sdk-outside-cluster)
-* [Kubeflow on Azure](https://awslabs.github.io/kubeflow-manifests/docs/component-guides/pipelines/)
-* [Kubeflow on IBM Cloud](/docs/distributions/ibm/pipelines/#2-authenticating-multi-user-kubeflow-pipelines-with-the-sdk)
+{{% alert title="Tip" color="info" %}}
+The process required will vary depending on which [distribution of Kubeflow](/docs/started/installing-kubeflow/#packaged-distributions-of-kubeflow) is being used.
+{{% /alert %}}
 
 ### Example for Dex
 
-For the deployments that use [Dex](https://dexidp.io/) as their identity provider, this example demonstrates how to authenticate the Pipelines SDK from outside the cluster.
+This example is for distributions of Kubeflow which use [Dex](https://dexidp.io/) as their identity provider and have `staticPasswords` or `ldap` configured.
 
-__Step 1:__ expose your `istio-ingressgateway` service locally (if your Kubeflow Istio gateway is not already exposed on a domain)
+##### Step 1:
+
+Expose your `istio-ingressgateway` service locally (if your Kubeflow Istio gateway is not already exposed on a public domain).
 
 ```bash
 # `svc/istio-ingressgateway` may be called something else, or use different ports
 kubectl port-forward --namespace istio-system svc/istio-ingressgateway 8080:80
 ```
 
-__Step 2:__ this Python code defines a `get_istio_auth_session()` function that returns a session cookie by authenticating with dex
+##### Step 2:
+
+This Python code defines a `KFPClientManager()` class which creates authenticated `kfp.Client()` instances when `get_kfp_client()` is called.
 
 ```python
 import re
-import requests
 from urllib.parse import urlsplit
 
-def get_istio_auth_session(url: str, username: str, password: str) -> dict:
+import kfp
+import requests
+import urllib3
+
+
+class KFPClientManager:
     """
-    Determine if the specified URL is secured by Dex and try to obtain a session cookie.
-    WARNING: only Dex `staticPasswords` and `LDAP` authentication are currently supported
-             (we default default to using `staticPasswords` if both are enabled)
-
-    :param url: Kubeflow server URL, including protocol
-    :param username: Dex `staticPasswords` or `LDAP` username
-    :param password: Dex `staticPasswords` or `LDAP` password
-    :return: auth session information
+    A class that creates `kfp.Client` instances with Dex authentication.
     """
-    # define the default return object
-    auth_session = {
-        "endpoint_url": url,    # KF endpoint URL
-        "redirect_url": None,   # KF redirect URL, if applicable
-        "dex_login_url": None,  # Dex login URL (for POST of credentials)
-        "is_secured": None,     # True if KF endpoint is secured
-        "session_cookie": None  # Resulting session cookies in the form "key1=value1; key2=value2"
-    }
 
-    # use a persistent session (for cookies)
-    with requests.Session() as s:
+    def __init__(
+        self,
+        api_url: str,
+        dex_username: str,
+        dex_password: str,
+        dex_auth_type: str = "local",
+        skip_tls_verify: bool = False,
+    ):
+        """
+        Initialize the KfpClient
 
-        ################
-        # Determine if Endpoint is Secured
-        ################
-        resp = s.get(url, allow_redirects=True)
+        :param api_url: the Kubeflow Pipelines API URL
+        :param skip_tls_verify: if True, skip TLS verification
+        :param dex_username: the Dex username
+        :param dex_password: the Dex password
+        :param dex_auth_type: the auth type to use if Dex has multiple enabled, one of: ['ldap', 'local']
+        """
+        self._api_url = api_url
+        self._skip_tls_verify = skip_tls_verify
+        self._dex_username = dex_username
+        self._dex_password = dex_password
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@kubeflow/wg-pipeline-leads this small documentation update should significantly improve users' understanding of how to connect the KFP SDK to a Dex cluster (a common point of confusion).

@zijianjoy @chensun you are probably the best people to review.

thesuperzapper:
@zijianjoy this is the PR with the updates to the "dex auth" example I showed today in the community meeting.

thesuperzapper:
@zijianjoy just wanted to highlight this PR again, as it will significantly improve the experience of users trying to connect with Kubeflow via dex auth.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
```diff
@@ -152,146 +152,186 @@ and give them access to submit pipelines in their own namespace.
 <summary>Click to expand</summary>
 <hr>
 
-The process to authenticate the Pipelines SDK from outside the cluster in multi-user mode will vary by distribution:
+When running the Kubeflow Pipelines SDK from outside a multi-user Kubeflow cluster (e.g. on a laptop), the `kfp.Client()` will need to be initialized with credentials.
 
-* [Kubeflow on Google Cloud](/docs/distributions/gke/pipelines/authentication-sdk/#connecting-to-kubeflow-pipelines-in-a-full-kubeflow-deployment)
-* [Kubeflow on AWS](/docs/distributions/aws/pipeline/#authenticate-kubeflow-pipeline-using-sdk-outside-cluster)
-* [Kubeflow on Azure](https://awslabs.github.io/kubeflow-manifests/docs/component-guides/pipelines/)
-* [Kubeflow on IBM Cloud](/docs/distributions/ibm/pipelines/#2-authenticating-multi-user-kubeflow-pipelines-with-the-sdk)
```
zijianjoy, line 160: Understanding that distribution section has been moved out of the `kubeflow.org`, I am leaning towards keeping this redirection using `updated links to each distribution website`. We can move this highlighted part to the bottom of this page around `## Next Steps` section, so the generic connection approach is higher priority on the page.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
```diff
@@ -152,146 +152,186 @@ and give them access to submit pipelines in their own namespace.
...
-        # else, we need to be redirected to the actual login page
+
+        # if we are at `/auth/xxxx/login` path, then we are at the login page
+        if re.search(r"/auth/.*/login$", url_obj.path):
+            dex_login_url = url_obj.geturl()
         else:
-            # this GET should redirect us to the `/auth/xxxx/login` path
-            resp = s.get(redirect_url_obj.geturl(), allow_redirects=True)
+            # otherwise, we need to follow a redirect to the login page
+            resp = s.get(url_obj.geturl(), allow_redirects=True)
```
zijianjoy, line 260: I think `allow_redirects=True` has been used earlier in `resp = s.get(self._api_url, allow_redirects=True)`, in what situation do we need to manually redirect here? Can we use an code `301` from response to determine it is a `redirect` response?

content/en/docs/components/pipelines/v1/sdk/connect-api.md
```diff
@@ -152,146 +152,186 @@ and give them access to submit pipelines in their own namespace.
...
-    username=KUBEFLOW_USERNAME,
-    password=KUBEFLOW_PASSWORD
+        return patched_kfp_client(
+            host=self._api_url,
+            cookies=session_cookies,
+        )
+
+    def get_kfp_client(self) -> kfp.Client:
+        """Get a newly authenticated Kubeflow Pipelines client."""
+        return self._create_kfp_client()
```
zijianjoy, line 312: NIT: Looks like this func is only calling `_create_kfp_client()`, and the func name `get_kfp_client()` which suggests that it is retrieving the same kfp.Client is actually generating new session cookie and a new kfp client every time being called. Maybe consider making `_create_kfp_client()` public directly?

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: @zijianjoy the problem is deciding which distributions get to be listed here, (and which page we would even link to).

I think the "tip" which links back to the "installing kubeflow" page should be sufficient:

Screenshot 2023-08-09 at 12 16 40

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: The problem is that multiple redirects can happen (and we don't know how many, because it depends on the user's dex configs, specifically, if they have multiple auth providers, or just one).

So using  `allow_redirects` makes it cleaner.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: I agree that `get_kfp_client` could be read either way, which is why the docs say "newly authenticated" and there is a tip in the following example which says `TIP: long-lived sessions might need to get a new client when their session expires`.

For reference, I think the class is more readable with a single (clearly public) method that is intended for the user.

However, could rename the current `get_kfp_client()` to `create_kfp_client()` but I actually think that is misleading, I will leave it up to you to decide.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
zijianjoy: I am not against using `allow_redirects=True`, but the `else` condition here is trying to call `resp = s.get(url_obj.geturl(), allow_redirects=True)` again, while `resp = s.get(self._api_url, allow_redirects=True)` earlier should have handled the redirect automatically.

If multiple manual `s.get()` is needed, it sounds like a `for loop` is needed, instead of `if/else`.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
zijianjoy: I would suggest to move this change to another PR, so I can help merging the `dex authentication` portion quicker, which aligns with the title of this PR (update kfp sdk "dex authentication" example)

content/en/docs/components/pipelines/v1/sdk/connect-api.md
zijianjoy: It seems that renaming to `create_kfp_client()` is clear to me. Would you like to explain why is it misleading?

content/en/docs/components/pipelines/v1/sdk/connect-api.md
chensun: +1 on what @zijianjoy said. @thesuperzapper please focus on the area you're trying to update without disturbing the rest.

chensun (changes requested):
@thesuperzapper The code is too heavy as a example IMO. Have we discussed implementing the dex support within KFP SDK client?

/cc @connor-mccarthy

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: @zijianjoy actually, I think we are slightly misunderstanding what's happening here (its been a while since I wrote this).

First, we do a get which may redirect us to a "choose auth type" page, if we land on that page, the following code rewrites the patch to the "chosen" auth type:

```python
        # if we are at `/auth?=xxxx` path, we need to select an auth type
        url_obj = urlsplit(resp.url)
        if re.search(r"/auth$", url_obj.path):
            url_obj = url_obj._replace(
                path=re.sub(r"/auth$", f"/auth/{self._dex_auth_type}", url_obj.path)
            )
```

After this, we need to perform another get to be redirected to the actual login page.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: I think this update (to remove these links) is relatively important, as the links are to distribution docs that no longer exist as part of https://github.com/kubeflow/website/issues/3074.

Furthermore, as part of https://github.com/kubeflow/website/issues/3074, we try not have any distribution-specific docs on the main website.

_PS: The reason we are including the dex example, is because that is the auth that the upstream `kubeflow/manifests` uses._

thesuperzapper:
> @thesuperzapper The code is too heavy as a example IMO. Have we discussed implementing the dex support within KFP SDK client?
> 
> /cc @connor-mccarthy

I think the new code is actually less heavy, and this is why I am proposing it.

But regarding having the code on the website at all, I am strongly in favor, as it is critical for users to understand how to connect to KFP's API from outside the cluster.

And because the `kubeflow/manifests` use dex for auth, it just makes sense to demonstrate that kind of auth.

content/en/docs/components/pipelines/v1/sdk/connect-api.md
thesuperzapper: @zijianjoy I agree, let's rename, I have done this in https://github.com/kubeflow/website/pull/3565/commits/c930d58164810ef67fcf5ed638028e84f179c225

content/en/docs/components/pipelines/v1/sdk/connect-api.md
zijianjoy: Thanks for the explanation, in that case, shouldn't we make `get call` immediately after this block?

```
  # if we are at `/auth?=xxxx` path, we need to select an auth type
        url_obj = urlsplit(resp.url)
        if re.search(r"/auth$", url_obj.path):
            url_obj = url_obj._replace(
                path=re.sub(r"/auth$", f"/auth/{self._dex_auth_type}", url_obj.path)
            )
            resp = s.get(url_obj.geturl(), allow_redirects=True)
```

thesuperzapper:
Closing as we have already updated this section.

PR #3571, opened 2023-08-03
Add the FAQs to address the deploying issue on kind cluster.

dashanji:
I have used the kind to build a kubernetes cluster and encounter such error `mysqld: Can't read dir of '/etc/mysql/conf.d/' (OS errno 13 - Permission denied)` which blocks the installation of standalone kubeflow. Also, I can't find the related issue or discussion. As a result, I think it's better to add FAQs to help coming people fix the same issue.


Arhell (commented):
/ok-to-test

varodrig:
@hbelmiro  the page doesn't longer exist but it seems we could make this update on this page 
https://www.kubeflow.org/docs/components/pipelines/legacy-v1/installation/localcluster-deployment/
cc @dashanji  to follow up

hbelmiro:
/close
Due to inactivity.

PR #3572, opened 2023-08-09
redirect distribution-specific docs to external websites

thesuperzapper:
closes https://github.com/kubeflow/website/issues/3074

This PR is the final part of https://github.com/kubeflow/website/issues/3074, and does the following:
- redirects the pages for distributions (which previously had a place on the sidebar) to their respective external websites
- replaces the `Distributions` section from the sidebar with a link to the "Installing Kubeflow" page


thesuperzapper:
@zijianjoy @jbottum I would appreciate a review on this one, so we can finally get https://github.com/kubeflow/website/issues/3074 closed!

jbottum:
We need to improve our website.  There is a lot of changes in this PR.   We probably need a call to walk through this.   Can it be done in 15 minutes during a Community call ?

thesuperzapper:
@jbottum This is a very small change in the scheme of things, as we are just implementing the plan we have been working towards for over a year in https://github.com/kubeflow/website/issues/3074.

Without this change, it gives unnecessary preference to the small number of distributions which retain their single-page entry on the sidebar (which is the whole idea behind https://github.com/kubeflow/website/issues/3074).

For the redirect links, I have used the same ones which the distribution owners provided in their table entry the [installing kubeflow page](https://www.kubeflow.org/docs/started/installing-kubeflow/#active-distributions), so they are definitely correct.

thesuperzapper:
@zijianjoy as discussed, I have updated the PR to only redirect the distribution-specific docs to their websites, but leave the `Distributions` section on the sidebar, with a link to the `Installing Kubeflow` page.

This will also let us include conformance testing docs under the `Distributions` section in the future, if we want.

zijianjoy:
/lgtm
/approve

Thank you @thesuperzapper !

PR #3574, opened 2023-08-10
move "Arrikto Enterprise Kubeflow" to legacy distributions

thesuperzapper:
This PR moves "Arrikto Enterprise Kubeflow" to the legacy distributions section.

While it has not technically been 6 months of inactivity (only about 4, since the [last release on `2023-03-31`](https://docs.arrikto.com/Changelog.html#version-2-0-2-aurora)), as the most recent version only included Kubeflow 1.5.0 (which was released on `2022-06-15`), I think its fair to say this distribution is no longer active.

I am sad to have to do this.

### DIFF
--- a/content/en/docs/started/installing-kubeflow.md
+++ b/content/en/docs/started/installing-kubeflow.md
@@ -109,23 +109,6 @@ The following table lists <b>active distributions</b> that have <b>had a recent 
         </td>
       </tr>
       <tr>
-        <td>Arrikto Enterprise Kubeflow</td>
-        <td>Arrikto</td>
-        <td>
-          ◦ Amazon Elastic Kubernetes Service (EKS)
-          <br>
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@cvenets can you confirm if "Arrikto Enterprise Kubeflow" is being actively developed, and if not, approve this PR?

/cc @jbottum @zijianjoy

Arhell (commented):
/lgtm

thesuperzapper:
@akgraner just want to confirm if you have any contact with former Arrikto people who can confirm this change.

Otherwise, I think we can merge it safely.

thesuperzapper:
@zijianjoy I think we can safely merge this given the lack of responses, and the fact that the company which backs this distribution does not appear to exist anymore.

akgraner:
Yes, it can be removed.

chensun (approved):

PR #3578, opened 2023-08-16
fix(pipelines): Update .output['output-key'] to plural

junggil:
.output['<output-key>'] should be .outputs['<output-key>'] (plural) instead in case of multiple outputs.


connor-mccarthy (approved):
/lgtm
/approve

PR #3588, opened 2023-09-21
Remove AutoML Calendar

andreyvelich:
Related: https://github.com/kubeflow/community/pull/652.

We moved AutoML and Training WG event to Kubeflow calendar.
I also added AutoML and Training Slack channels.

cc @kubeflow/wg-training-leads @tenzen-y @thesuperzapper

PR #3589, opened 2023-09-22
update slack channel links

thesuperzapper:
This PR ensures we are linking to the public Slack channels, rather than the internal working group ones.
So we don't have people asking questions in the WG channels.

- `#wg-automl` replaced with `#kubeflow-katib`
- `#wg-training` replaced with `#kubeflow-training-operator`

It also updates the name of the channel for KFServing to KServe, and reflects that it is now a separate project:

- `#kubeflow-kfserving` renamed to `#kserve`

---

The problematic links were introduced in https://github.com/kubeflow/website/pull/3588


terrytangyuan (approved):
/lgtm

andreyvelich:
@thesuperzapper There are not internal Slack channels.
#wg-automl and #wg-training is the updated channels where we ask our users to join.
#kubeflow-katib and #kubeflow-training-operator is the legacy channels.

andreyvelich:
@thesuperzapper Are you going to submit PR to update it back ?

thesuperzapper:
@andreyvelich , that is not correct, when we made those channels we were clear that the users should discuss in the `#kubeflow-xxxx` channels, and the working group members should discuss in `#wg-xxxx`.

It's also a lot less confusing, because users don't think about the working groups they think about the app they're using.

Furthermore, if you compare the member counts of each of these channels, the `#kubeflow-xxxx` ones have at least 2x as many members, so I would highly recommend you continue using those.

thesuperzapper:
@andreyvelich firstly, the reason that channel is listed in your charter is because it's where you're meant to discuss working group matters .

You're not really allowed to have private channels, as the working groups are meant to operate in the open, where possible.

I would suggest that you either make your main channel open or you can just start using the official channel. (After you make your private channels open , we could rename the existing `#wg-xxxx` channels to `#kubeflow-xxxx`, if you would like, but that seems unnecessary given your private channels probably have less members).

andreyvelich:
>You're not really allowed to have private channels, as the working groups are meant to operate in the open, where possible.

Why this statement is true ? I don't think we discussed it before.

>(After you make your private channels open , we could rename the existing #wg-xxxx channels to #kubeflow-xxxx, if you would like, but that seems unnecessary given your private channels probably have less members).

I don't understand why we can't keep channels as `wg-training`, etc. ?

For example, for Manifests we have [`wg-manifests` ](https://kubeflow.slack.com/archives/C01EY3L525N) and for Notebooks we have [`wg-notebooks` ](https://kubeflow.slack.com/archives/C01DJECE8H5) channel.
We should be consistent across different Kubeflow components to not confuse users.

cc @juliusvonkohout @kimwnasptd

thesuperzapper:
The main channels for users to ask questions about Notebooks and Pipelines are

- `#kubeflow-pipelines` (2999 members)
- `#kubeflow-notebooks` (1289 members)

The `#wg-notebooks` channel is used by the working group to discuss working group matters.

Also, I'm not sure that the pipelines working group actually uses slack for this purpose so they don't have a channel as far as I can see.

andreyvelich:
I think, the main reason is why these channels have more users because they were created earlier (5 years ago) than WGs were established (3 years ago).
For example, Notebooks working group maintain 9 components. Do we need 9 Slack channels for users ?
Decrease number of channels will be better I think, since users don't need to know about all existing Kubeflow Slack channels.

I am happy to discuss it during our Kubeflow Community Call to provide consistent across Kubeflow components.
cc @jbottum @james-jwu @zijianjoy

PR #3591, opened 2023-09-25
remove reference to now-deprecated "Python based visualizations"

gregsheremeta:
Python based visualizations were deprecated in b2483a1. Yet there was language left over that referred to them as the "new" way to do visualizations. This was highly confusing.

Fix some typos too.


Arhell (commented):
/ok-to-test

PR #3628, opened 2023-11-17
Netlify build command cleanup

chalin:
This PR contains suggested improvements to the site build scripts and config.

- Simplifies the `netlify.toml` build commands by using the NPM scripts. E.g., the production build command goes from:
  ```sh
  cd themes/docsy && git submodule update -f --init && npm install && cd ../.. && hugo --gc --minify
  ```
  to more the more concise yet equivalent:
  ```sh
  npm run build:production
  ```
- Factors out common env var in `netlify.toml`
- Adds NPM scripts that simplify building and serving the site

/cc @nate-double-u


netlify.toml
```diff
@@ -1,16 +1,6 @@
...
-  command = "cd themes/docsy && git submodule update -f --init && npm install && cd ../.. && hugo --gc --minify"
+  command = "npm run build:preview"
 
-[context.deploy-preview.environment]
-  HUGO_VERSION = "0.119.0"
-  NODE_VERSION = "18"
-
-[context.production.environment]
-  HUGO_VERSION = "0.119.0"
-  HUGO_ENV = "production"
```
chalin, line 11: Setting `HUGO_ENV ` is no longer required.

Arhell (commented):
/ok-to-test

chalin:
Anything left to be done here? /cc @thesuperzapper

netlify.toml
```diff
@@ -1,16 +1,6 @@
...
-  NODE_VERSION = "18"
-
-[context.branch-deploy.environment]
-  HUGO_VERSION = "0.119.0"
-  NODE_VERSION = "18"
```
thesuperzapper, line 16: @chalin Won't removing this prevent the old branches like https://github.com/kubeflow/kubeflow/tree/v1.7-branch, from being deployed, because we achieve this by using a CNAME record from:

- `v1-7-branch.kubeflow.org` -> `v1-7-branch--competent-brattain-de2d6d.netlify.app`

I don't have access to the Netelify dashboard, so I can't see where those are being deployed from.

netlify.toml
```diff
@@ -1,16 +1,6 @@
...
\ No newline at end of file
+[context.production]
+  command = "npm run build:production"
```
thesuperzapper, line 6: Are you sure that using `context.production` will still work?

I don't have access to the Netelify dashboard, so I can check, but based on Hugo's docs: https://gohugo.io/hosting-and-deployment/hosting-on-netlify/ they seem to suggest that `context.production.environment` is the right heading.

@james-jwu or @zijianjoy probably have access to Netelify?

thesuperzapper:
@chalin Also, all our PR/historical branches are getting indexed by Google, we should fix that at the same time as this PR. 

The goals would be:

1. The main `www.kubeflow.org` site should be indexed
2. All PR `deploy-preview-XXXX--competent-brattain-de2d6d.netlify.app` should NOT be indexed
3. All other `v1-7-branch.kubeflow.org` sites should be NOT be indexed:
     - _(these are just CNAME records pointing to the branch domains like `v1-7-branch--competent-brattain-de2d6d.netlify.app`)_

I believe your changes here achieve 2, because you are setting `-e dev` in the hugo command, and because this is not "production", docsy adds `<meta>` no index tags.

We need to be careful about 1. Are you 100% confident that not setting `-e production` or `HUGO_ENV=production` is safe?

To achieve 3, we could set the `HUGO_ENV` from `[context.branch-deploy.environment]` to dev, but it will probably propagate faster if we use a `robots.txt` disallow on those domains (otherwise, the `<meta>` tags will take until Google next indexes each page).

chalin:
I've created a separate issue to track the full extent of the indexing issue, since more work will need to be done outside of the scope of this PR:

- #3645

netlify.toml
chalin: @thesuperzapper - from my understanding of how Netlify works, unless you back port this change to an older v1.x branch, the builds of other branches won't be affected.

For example, v1.7 has it's own `netlify.toml` file, https://github.com/kubeflow/website/blob/v1.7-branch/netlify.toml, with it's own env var values and commands:
https://github.com/kubeflow/website/blob/b48a664062876fdc54aee9c831d0032924734a1e/netlify.toml#L1-L16

But, I could be wrong.

netlify.toml
chalin: I'm using similar setups for the OTel and gRPC websites, e.g., see https://github.com/open-telemetry/opentelemetry.io/blob/5df8d080fe6997d1ee514b9f50f6880155276e5b/netlify.toml#L1-L6:

```toml
[build]
publish = "public"
command = "npm run seq -- build:preview diff:check"

[context.production]
command = "npm run seq -- build:production diff:check"
```

The `.environment` section is for setting environment variables only, not for setting other fields like `command`.

chalin:
> I believe your changes here achieve 2, because you are setting `-e dev` in the hugo command, and because this is not "production", docsy adds `<meta>` no index tags.

Correct.

> We need to be careful about 1. Are you 100% confident that not setting `-e production` or `HUGO_ENV=production` is safe?

For reassurance, see https://discourse.gohugo.io/t/what-does-setting-hugo-env-to-production-do/24669/2. Since the site uses `hugo.IsProduction`, we're good.

> To achieve 3, we could set the `HUGO_ENV` from `[context.branch-deploy.environment]` to dev, but it will probably propagate faster if we use a `robots.txt` disallow on those domains (otherwise, the `<meta>` tags will take until Google next indexes each page).

I'll answer this in #3645.

package.json
```diff
@@ -1,8 +1,29 @@
...
+    "get:submodule": "npm run _get:${GET:-submodule}",
+    "postget:submodule": "git submodule",
+    "prepare": "npm run get:submodule && npm run _prepare:docsy",
+    "serve": "npm run _serve:hugo",
+    "test": "echo 'TODO: setup link checking'"
+  },
   "devDependencies": {
     "autoprefixer": "^10.4.0",
-    "postcss": "^8.4.1",
+    "hugo-extended": "^0.119.0",
```
thesuperzapper, line 20: @chalin are you sure that this is the correct way to set the Hugo version on Netelify?

Because the Netelify website suggests using `HUGO_VERSION`, which is what we were previously doing.

netlify.toml
thesuperzapper: @james-jwu or @zijianjoy do you have access to Netelify for the Kubeflow website, so we can confirm how it is configured?

We probably also need to give access in some way to the CNCF for continuity purposes.

package.json
chalin: This is what we're using on other sites such as OTel: https://github.com/open-telemetry/opentelemetry.io/blob/299ad7bb24488e39eb17d43393dd05f513321253/package.json#L89

The advantage of using `hugo-extended` is that you'll never have to worry again about using the right version of Hugo even if (at some point in the future) you work on branches of older versions of the docs since NPM install and the NPM build commands will naturally use the version of Hugo needed for that specific branch/commit.

thesuperzapper:
I believe what we are doing is now fine (e.g. we already use hugo extended as this is the default for Netlify), please reopen a new PR if there are still issues.

/close

PR #3635, opened 2023-11-25
Rewrite "Central Dashboard" docs and Move "Multi-Tenancy" section

thesuperzapper:
This PR is a significant rewrite of the "Central Dashboard" docs, as they have not been updated in many years.

In addition to the rewrite, I have moved the "Multi-Tenancy" section under "Central-Dashboard" as it's not really an independent component, and the content is much better suited under the "Central-Dashboard" component.

I have also made @kimwnasptd and I the owners of this section, to reflect that the central dashboard is owned by @kubeflow/wg-notebooks-leads, and there was no previous OWNERS file.


thesuperzapper:
@zijianjoy as there is no existing owners file for the central dashboard, we will need a root approver.

zijianjoy:
How do you think about moving `central dashboard` and `multi-tenancy` to a separate categories than `Component`? For example: We can add a new section called `Core Concepts` or `Fundamentals`, then put `central dashboard` and `multi-tenancy` as children. It is because I don't think `multi-tenancy` is fully a child concept under `central dashboard`, thus I try to find a proper location to host them.

thesuperzapper:
@zijianjoy While I agree that multi-tenancy probably should live in a concepts section (or really  just anywhere not under components) I think that the central dashboard is not a concept, and is definitely a component.

But the bigger problem is that the multi-tenancy docs just aren't very good, and are honestly likely to confuse the reader. So giving it its own section is probably bad.

It's also the only section under components that doesn't give any actionable information about Kubeflow.

I think there are a few options:

- remove the multi-tenancy section
- put the multi-tenancy section under "about"
- create a new "concepts" section, which only has multi-tenancy (with the possibility of new sections being added and the multi-tenancy being improved)

thesuperzapper:
@zijianjoy I have updated the PR in https://github.com/kubeflow/website/pull/3635/commits/1ebe583873486d07e760f5a670d9934d825366b9, and moved the "multi-tenancy" section under a new "concepts" section.

I put the following warning on it:

Screenshot 2023-11-27 at 20 16 44

content/en/docs/components/central-dash/profiles.md
```diff
@@ -0,0 +1,238 @@
++++
+title = "Profiles and Namespaces"
+description = "About Kubeflow Profiles and Namespaces for multi-user isolation"
+weight = 20
++++
+
+## What is a Kubeflow Profile?
```
zijianjoy, line 7: It seems to me that this page should be under `multi-tenancy` instead of `Central Dashboard`. How do you think?

thesuperzapper: @zijianjoy I have two strong reasons to keep profiles under the central dashboard section:

1. It's very related to the central dashboard (parts of the guide even have screenshots showing how to manage contributors using the web-ui in the dashboard)
2. We should keep the new "concepts" section about actual concepts, not about how to actually use the components (and I think the profiles page is probably one of the most important pieces of platform-usage docs we have, especially after the update).

zijianjoy: Although I have preference on the followings, I feel that merging the current PR is going to be helpful, and we can figure out how to adjust in the future:

- Users can manage kubeflow profile via central dashboard right, but they can also manage via yaml. So central dashboard and yaml are only the interfaces of managing multi-tenancy functionality.
- I feel that `concepts` section can document usage guide as well.

zijianjoy:
/lgtm
/approve

Thank you @thesuperzapper for the change!

PR #3641, opened 2023-11-30
move `Kubeflow on AWS` to legacy distributions

thesuperzapper:
As indicated by AWS in this thread https://github.com/awslabs/kubeflow-manifests/issues/794, the `Kubeflow on AWS` distribution is no longer actively maintained.

This PR updates the "installing kubeflow" table to reflect this fact.

### DIFF
--- a/content/en/docs/started/installing-kubeflow.md
+++ b/content/en/docs/started/installing-kubeflow.md
@@ -43,19 +43,6 @@ The following table lists <b>active distributions</b> that have <b>had a recent 
     </thead>
     <tbody>
       <tr>
-        <td>Kubeflow on AWS</td>
-        <td>Amazon Web Services</td>
-        <td>
-          Amazon Elastic Kubernetes Service (EKS)
-        </td>
-        <td>
-          <a href="https://awslabs.github.io/kubeflow-manifests">Website</a>
-        </td>
-        <td>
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@surajkota if you can confirm the `Kubeflow on AWS` distribution is no longer maintained then @zijianjoy can you please merge this PR.

This is important to avoid confusion for new users who are using AWS, to prevent them from adopting an inactive distribution.

thesuperzapper:
@jbottum @zijianjoy I have separated the style changes that were in this PR into a separate one https://github.com/kubeflow/website/pull/3643 (which I think we should merge, as they help new users).

This PR is not only for moving AWS to legacy, so will allow us to give AWS more time to confirm if they plan to continue supporting their distribution.

jbottum:
@thesuperzapper I think we need to consider that the last AWS release was for 1.7 and has a date of Sept 23 per this link, https://github.com/awslabs/kubeflow-manifests/releases, which I believe is within the 6 month window that you have defined as an active distribution.

varodrig:
> #3693

is merged already, however I don't see the Legacy distributions on the web page. should we closed this PR? @andreyvelich

thesuperzapper:
We can close it, because we are actively keeping older distributions on the website for now, mainly to highlight that they are old so people know to move to another distribution.

PR #3652, opened 2023-12-15
Remove KFP v2-compatible docs

chensun:
v2-compatible docs are causing confusions to users: https://github.com/kubeflow/pipelines/issues/10287
Given v2 is out for 6 months and we have v2 docs updated, we should remove docs on the experimental v2-compatible mode.


connor-mccarthy:
/lgtm
/hold to let @chensun submit

I like this choice 👍🏼 . Thank you!

PR #3667, opened 2024-01-30
GSOC 2024 proposal

rareddy:
Add project ideas for Kubeflow's application to Google Summer of Code 2024, aiming to participate as an organization and recruit student contributors

PR #3669, opened 2024-02-01
Adding link to events page from top nav

rareddy:
Adding "events" page to top nav and also replacing Kubeflow Summit with GSoC reference.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -36,7 +36,7 @@ ignoreFiles = []
   [[menu.main]]
     name = "Events"
     weight = -102
-    pre = "<i class='fas fa-book pr-2'></i>"
+    pre = "<i class='fas fa-calendar pr-2'></i>"
     url = "/events/"
   [[menu.main]]
     name = "Docs"
--- a/config.toml
+++ b/config.toml
@@ -28,17 +28,18 @@ ignoreFiles = []
 ###############################################################################
 [menu]
   [[menu.main]]
-    name = "Google Summer Of Code 2024 "
+    name = "Kubecon EU 2024 "
     weight = -1000
     pre = "<i class='fas fa-calendar pr-2' style='color: #FFC107'></i>"
-    url = "/events/gsoc-2024/"
+    post = "<br><span class='badge badge-warning'>March 9-12th</span> <span class='badge badge-warning'>PARIS, FRANCE</span> "
+    url = "https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/"
   [[menu.main]]
     name = "Events"
-    weight = -101
+    weight = -102
     pre = "<i class='fas fa-book pr-2'></i>"
     url = "/events/"
   [[menu.main]]
-    name = "Documentation"
+    name = "Docs"
     weight = -101
     pre = "<i class='fas fa-book pr-2'></i>"
     url = "/docs/"
--- a/config.toml
+++ b/config.toml
@@ -28,11 +28,15 @@ ignoreFiles = []
 ###############################################################################
 [menu]
   [[menu.main]]
-    name = "Kubeflow Summit 2023 "
+    name = "Google Summer Of Code 2024 "
     weight = -1000
     pre = "<i class='fas fa-calendar pr-2' style='color: #FFC107'></i>"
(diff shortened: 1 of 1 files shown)


rareddy:
@thesuperzapper the navigation changes you are recommending in #3667 for the "events" for your review.

Arhell (commented):
/ok-to-test

config.toml
```diff
@@ -28,11 +28,15 @@ ignoreFiles = []
...
-    post = "<br><span class='badge badge-warning'>Oct 6th</span> <span class='badge badge-warning'>Irving, Texas</span> <span class='badge badge-warning'>Virtual</span>"
-    url = "/events/kubeflow-summit-2023/"
+    url = "/events/gsoc-2024/"
+  [[menu.main]]
+    name = "Events"
+    weight = -101
+    pre = "<i class='fas fa-book pr-2'></i>"
+    url = "/events/"
   [[menu.main]]
     name = "Documentation"
```
thesuperzapper, line 41: Because this top bar now takes up more space, let's make this `Docs`, rather than `Documentation`.

config.toml
```diff
@@ -28,11 +28,15 @@ ignoreFiles = []
 ###############################################################################
 [menu]
   [[menu.main]]
-    name = "Kubeflow Summit 2023 "
+    name = "Google Summer Of Code 2024 "
     weight = -1000
     pre = "<i class='fas fa-calendar pr-2' style='color: #FFC107'></i>"
-    post = "<br><span class='badge badge-warning'>Oct 6th</span> <span class='badge badge-warning'>Irving, Texas</span> <span class='badge badge-warning'>Virtual</span>"
-    url = "/events/kubeflow-summit-2023/"
+    url = "/events/gsoc-2024/"
```
thesuperzapper, line 34: Personally, I think the "highlighted events" should be for actual in-person events that typical end-users can attend, and it's unlikely that GSOC will be relevant to most end-users of Kubeflow.

So let's just comment out the existing "Kubeflow Summit 2023" for now, so we can use it as a template for the next event.

If you are able, we could also make a page for Kubecon EU 2024, which Kubeflow is going to have a presence at, and make that the "highlighted" event.  (FYI, it's on `19 – 22 MARCH, 2024`, but I need to confirm who has the info about what we are doing there before we list it as a page).

config.toml
```diff
@@ -28,11 +28,15 @@ ignoreFiles = []
...
+  [[menu.main]]
+    name = "Events"
+    weight = -101
```
thesuperzapper, line 37: I think we can make "Events" be positioned before "Docs", so make this weight `-102`.

thesuperzapper (changes requested):
@rareddy I suggested some changes.

config.toml
rareddy: corrected

config.toml
rareddy: Agree.

So for now I added Kubecon but going to the Kubecon page, once we have the agenda or presenters then we will add the landing page. This can be a nice placement until then?

config.toml
rareddy: thanks for the suggestion, corrected it now.

thesuperzapper:
@rareddy I just saw that we have a page for Kubeflow Summit 2024 on CNCF website, which is co-located with KubeCon EU 2024:

- https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/co-located-events/kubeflow-summit/

Let's use that link and update the info to: 
- Title: "Kubeflow Summit EU / KubeCon EU"
- Date: "19 March, 2024"
- Location: "Paris, France"

Also, no need to capitalize the location.

thesuperzapper:
@rareddy sorry to suggest one more change, but after looking at it, I think it will probably look better to only have one "EU" in the title, so something like `Kubeflow Summit / KubeCon Europe`

After that, it's great, and we will need one of the root approvers to accept it.

thesuperzapper:
@james-jwu @zijianjoy can you please approve this small website update?

It replaces the old Kubeflow Summit link with the Kubeflow Summit/KubeCon Europe one (which is happening in 6 weeks).

/lgtm

thesuperzapper:
@rareddy actually we should use a calendar logo for the "events" link, probably just use the same one as the "KubeCon" link.

thesuperzapper:
Ok, @james-jwu @zijianjoy after that small update it should be ready to approve now.

/lgtm

PR #3678, opened 2024-02-14
Training: Add Distributed Training Diagrams

andreyvelich:
Related: https://github.com/kubeflow/training-operator/issues/1998

I added 2 diagrams for PyTorch distributed training with ring all-reduce algorithm and Tensorflow distributed training with PS.

It gives an idea what is the Training Operator responsibility.

Please take a look @kubeflow/wg-training-leads @diegolovison @hbelmiro @kubeflow/release-team 

/hold for review


content/en/docs/components/training/overview.md
```diff
@@ -41,11 +33,67 @@ for each ML framework:
...
+[TensorFlow Distributed APIs](https://www.tensorflow.org/guide/distributed_training) and create a
+TFJob with required number PSs, workers, and GPUs using Training Operator Python SDK.
+Then, Training Operator creates Kubernetes pods with appropriate environment variables for
+[`TF_CONFIG`](https://www.tensorflow.org/guide/distributed_training#setting_up_the_tf_config_environment_variable)
+to start distributed TensorFlow training job.
+
+Parameter server splits training data for every worker and averages model weights based on gradients
+produced by every worker.
+
+You can define various distributed strategy supported by TensorFlow in your training code, and Training
```
tenzen-y, line 94: Can we mention this document? https://www.tensorflow.org/guide/distributed_training#types_of_strategies

tenzen-y (commented):
Thank you for the great documentation!

content/en/docs/components/training/overview.md
andreyvelich: Yes, let me add those!

tenzen-y (commented):
Great work!
/lgtm

diegolovison (approved):

content/en/docs/components/training/overview.md
```diff
@@ -41,11 +33,68 @@ for each ML framework:
...
+  class="mt-3 mb-3">
+
+Training Operator is responsible to schedule the appropriate Kubernetes workloads to implement
+various distributed training strategies for different ML frameworks. The following examples show
+how Training Operator allows to run distributed PyTorch and TensorFlow on Kubernetes.
+
+### Distributed Training for PyTorch
+
+This diagram shows how Training Operator creates PyTorch workers for
+[ring all-reduce algorithm](https://tech.preferred.jp/en/blog/technologies-behind-distributed-deep-learning-allreduce/).
```
kuizhiqing, line 56: I'm OK here to put the allreduce algorithm as an example to show distributed training with PyTorch, however, I've a little bit concerns for two aspects,

* if we limit the scope to ring allreduce, horovod would be the best practice
* different and more complicate collective algorithms can be used with PyTorch by adopting NCCL or alternatives, all can be done with PyTorchJob configuration, e.g. [accelerate](https://github.com/huggingface/accelerate)

My personal suggestion would be to emphasize this as example and the operator is open to all popular framework and user cases.

kuizhiqing (commented):
Thanks for your great work.

A mino change would be better for me that the index in the diagrams can count from 0, since the index for worker and for data in memory are begin with 0. It would be OK for me if you are just keep it, it's my personal taste indeed.

content/en/docs/components/training/overview.md
tenzen-y: It's good points. I think that we can mention in the top of `##Architecture` section.

andreyvelich:
>A mino change would be better for me that the index in the diagrams can count from 0, since the index for worker and for data in memory are begin with 0. It would be OK for me if you are just keep it, it's my personal taste indeed.

I agree with you @kuizhiqing, even NCCL ranks start from 0: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html
Will change it.

content/en/docs/components/training/overview.md
andreyvelich: I agree that Horovod will be preferable implementation for ring all-reduce, but PyTorch DDP docs also can do it as it says here: https://pytorch.org/tutorials/beginner/ddp_series_theory.html#what-is-distributed-data-parallel-ddp.

>My personal suggestion would be to emphasize this as example and the operator is open to all popular framework and user cases.

Yes, I tried to emphasize it here: https://github.com/kubeflow/website/pull/3678/files#diff-768cd67df75edc45450aa16fb45d81da4b5fc0dfdcd2457c2bd81cc16a10e3e7R72-R73
@kuizhiqing Do you have any other suggestion how we can explain it ?

hbelmiro (commented):
Just a few typos @andreyvelich. The rest LGTM.

andreyvelich:
Thanks for review @diegolovison!
I addressed your comments.

tenzen-y (commented):
Thanks!
/lgtm

PR #3719, opened 2024-04-23
Training: Reorganized Training Operator Docs

andreyvelich:
Related: https://github.com/kubeflow/training-operator/issues/1998.
I created the following sections for Training Operator docs:
- Overview
- Installation 
- Getting Started
- User Guides
- Reference

A few points:

- @StefanoFioravanzo @kubeflow/wg-training-leads Any ideas on what we could add to `Why Training Operator ?` section ? Initially, we can just add some basic info.
- I didn't move CRDs to reference in this PR since we don't have time to discuss how we are going to generate them. What do you think we should do in this PR ?
- Do we need to have working example in GettingStarted page ? Would it be too complicated to consume ?

/hold for review

/assign @StefanoFioravanzo @kubeflow/wg-training-leads @hbelmiro @kuizhiqing @droctothorpe @franciscojavierarceo
Looking for your feedback!


content/en/docs/components/training/getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+
+```python
+def train_func():
+    import torch
+    import os
+
+    # Create model.
+    class Net(torch.nn.Module):
+        """Create the Pytorch model"""
+        ...
```
franciscojavierarceo, line 36: nit: my recommendation would be to populate this even with a trivial single layer NN that would actually run for this example. It helps users get started that may just be executing copy-pasta.

StefanoFioravanzo:
@andreyvelich thanks for this! 

> Any ideas on what we could add to `Why Training Operator?` section ?

Let's start with something simple and iterate in future PRs. We can start by answering questions like: 

- How does training operator simplify distributed training with respect to a more traditional approach?
- How does Kubernetes help in solving these problems?
- Why is Training Operator part of the Kubeflow ecosystem?

> I didn't move CRDs to reference in this PR since we don't have time to discuss how we are going to generate them. What do you think we should do in this PR ?

Makes sense. I'd keep the scope of this PR to the restructuring you already implemented. Let's iterate on content separtely. We can address each framework's user guide in dedicated PRs.

> Do we need to have working example in GettingStarted page ? Would it be too complicated to consume ?

Getting Started should have an end-to-end working (yet simple) example. Generally people just want to copy paste some stuff, run it, and see results. Then you typically link some more advanced tutorials or user guides at the end

content/en/docs/components/training/getting-started.md
andreyvelich: Sure, let me try to add something simple.

andreyvelich:
That makes sense @StefanoFioravanzo, I added initial ideas for `Why Training Operator ?` and also I added the AI/ML lifecycle diagram that we can re-use in various Kubeflow components to explain which stage of lifecycle each component addresses (e.g. Spark Operator, Model Registry, Katib, Notebooks, KServe).
Please let me know what do you think @kubeflow/wg-training-leads @StefanoFioravanzo ?

content/en/docs/components/training/overview.md
```diff
@@ -22,136 +24,55 @@ Training Operator implements centralized Kubernetes controller to orchestrate di
...
-the appropriate environment variables for `torchrun`.
-
-### Distributed Training for TensorFlow
-
-This diagram shows how Training Operator creates TensorFlow parameter server (PS) and workers for
-[PS distributed training](https://www.tensorflow.org/tutorials/distribute/parameter_server_training).
-
-<img src="/docs/components/training/images/distributed-tfjob.drawio.svg"
-  alt="Distributed TFJob"
+<img src="/docs/components/training/images/ml-lifecycle-training-operator.drawio.svg"
```
franciscojavierarceo, line 41: I love this diagram. Based on my [comment here](https://github.com/feast-dev/feast/issues/4139), it may make sense to also feature serving along with model serving (though I note that there are some architectural choices to be had on that direction). Regardless, I think it's worth stating explicitly.

andreyvelich:
@franciscojavierarceo I added working example, does it look good ?

content/en/docs/components/training/overview.md
andreyvelich: That makes sense, do we want to update the diagram once we discuss the architecture for Feature Serving ?
For this diagram, I just took diagram that we worked together with @ronaldpetty @zanetworker for CNCF WG AI WhitePaper
https://www.cncf.io/wp-content/uploads/2024/03/cloud_native_ai24_031424a-2.pdf

franciscojavierarceo: Oh nice, looks like that was very recent. @ronaldpetty @zanetworker let me know if you have any thoughts/opinions there. Would love for feature serving to be included in this as I think it is becoming increasingly more important.

andreyvelich: @franciscojavierarceo Not sure how to make it simple in this diagram. What do you think about this:
Screenshot 2024-04-24 at 23 20 02

franciscojavierarceo: This may be overkill...and it's technically missing the measurement required to actually repeat the process (e.g., a click for a recommendation engine).

Screenshot 2024-04-24 at 11 55 19 PM

andreyvelich: That makes sense, thanks @franciscojavierarceo!
A few questions:
1. Isn't the feature store also use for Model Training and Optimization ?
2. Should we split the Model Development, Iteration, and Optimization with:
Model Experimentation and Development (Kubeflow Notebooks) --> Model Optimization and Hyperparameter  Tuning (Katib) (similar to this flow: https://www.kubeflow.org/docs/started/architecture/#introducing-the-ml-workflow).
Since  I also want to use this diagram in other docs: Kubeflow Notebooks, Kubeflow Katib.

andreyvelich: That looks good, I would also split Model Experimentation with Chose ML Algorithms + Code Model and HP Tuning + Architecture Search.
That will allow us to explain users on which stage Kubeflow Notebooks are used on which Kubeflow Katib is used

franciscojavierarceo: Will send you a message on slack

content/en/docs/components/training/overview.md
```diff
@@ -22,136 +24,55 @@ Training Operator implements centralized Kubernetes controller to orchestrate di
...
-    resources_per_worker={"gpu": "2"},
-)
-```
+| ML Framework | Custom Resource                                                  |
+| ------------ | ---------------------------------------------------------------- |
+| PyTorch      | [PyTorchJob](/docs/components/training/user-guides/pytorch/)     |
+| TensorFlow   | [TFJob](/docs/components/training/user-guides/tensorflow/)       |
+| XGBoost      | [XGBoostJob](/docs/components/training/user-guides/xgboost/)     |
+| MPI          | [MPIJob](/docs/components/training/user-guides/mpi/)             |
+| PaddlePaddle | [PaddleJob](/docs/components/training/user-guides/paddlepaddle/) |
```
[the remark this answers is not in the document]
andreyvelich: Nice catch!

andreyvelich:
This PR should be ready unless you have any other comments.
/hold cancel

hbelmiro (commented):
/lgtm

content/en/docs/components/training/overview.md
zanetworker: @franciscojavierarceo this sounds like a good discussion to have on #wg-artificial-intelligence :)

If we are looking for abstractions, I'd categorize it as Data prep, model building (training, tuning, experimentation, feature engineering,...), model serving & deployment (pull from registry, deploying inference,...), then operation & eval (monitoring,..). 

I.e., there are phases in the lifecycle (building, serving, operations and iteration), components (feature stores, registries, infra,...), and personas (highlighted in the white paper, but also here: https://tag-runtime.cncf.io/wgs/cnaiwg/glossary/#data-engineers, https://tag-runtime.cncf.io/wgs/cnaiwg/glossary/#data-engineers)

PR #3725, opened 2024-04-29
Fix Local Builds + Update: Hugo, MathJax, Mermaid

thesuperzapper:
Because Hugo `0.125.0+` does not work (which will break people's ability to develop the website locally), I have put an explanation in the README of how users can downgrade `hugo` with `brew`.

This PR changes the behavior of the site to improve performance:

- Only load MathJax when `mathjax: true` is explicitly put in the header of the page:
    - _NOTE: the only page using math is `/docs/components/katib/hyperparameter/`_
- Only load Mermaid when a `mermaid` code block is present on the page
    - _NOTE: the only page using mermaid is `/docs/components/notebooks/container-images/_
- We now use the standard [Hugo Syntax](https://gohugo.io/content-management/mathematics/) of `\(INLINE_MATH\)` rather than `$INLINE_MATH$` which breaks other pages.

This PR also makes the following updates:

- Updates Hugo to `0.124.1` (WARNING: `0.125.0` does NOT work)
- Updates MathJax to `3.2.2`
- Updates Mermaid to `10.9.0`


thesuperzapper:
@james-jwu @andreyvelich @jbottum this PR is small but important for 2 reasons:

1. Currently, if users follow the "local dev guide" hugo wont work (if they use the latest `hugo`).
2. It improves the performance of the website (only load heavy JS packages on the pages we actually use them).

I would appreciate your review.

andreyvelich (commented):
Thank you for fixing this @thesuperzapper!
/lgtm
/assign @james-jwu @jbottum @johnugeorge @terrytangyuan @zijianjoy

terrytangyuan (approved):
/lgtm

PR #3743, opened 2024-05-24
Removed reference to non existing path

rajeeshckr:
Looks like `platform-agnostic-pns` folder seems to be removed in [this](https://github.com/kubeflow/pipelines/commit/809d5766fc9ec436ff05c083e9a2ae65ad2667b7) commit

I hope `platform-agnostic` is the new one to use. 

Also, there was an error with `workflow-controller` and `metadata-envoy-deployment` using the same port causing crash loop. I modified the `workflow-controller` deployment to change the to a non conflicting port.

image


hbelmiro:
@gmfrasca can you PTAL?

chensun (approved):

PR #3763, opened 2024-06-13
Enabled the Checklinks Netlify plugin

hbelmiro:
This PR adds the [Checklinks](https://www.netlify.com/integrations/community-built/checklinks-build-plugin/) Netlify plugin for checking broken links during the CI.

I mentioned that in the [Kubeflow Community Meeting on 2024-06-18](https://docs.google.com/document/d/1Wdxt1xedAj7qF_Rjmxy1R0NRdfv7UWs-r2PItewxHpE/edit#heading=h.aba10q1p5i5r).

Checklinks fails the Netlify build when broken links are detected.
To avoid breaking the build due to the existing broken links, I added them to a TODO list so we can fix them later.
See a [screencast](https://github.com/kubeflow/website/assets/11776454/01cf5cf0-3d8a-49dd-b32e-89506c4a71d9) of the plugin in action.

cc @jbottum @andreyvelich

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,6 +2,42 @@
   publish = "public"
   command = "cd themes/docsy && git submodule update -f --init && npm install && cd ../.. && hugo --gc --minify"
 
+[[context.deploy-preview.plugins]]
+package = "netlify-plugin-checklinks"
+
+  [context.deploy-preview.plugins.inputs]
+
+  # You can mark some check as todo, which will execute the check, but allow failures.
+  # todoPatterns is an array of strings you can match against failing reports
+  todoPatterns = [
+    "public/docs/components/central-dash/",
+    "public/docs/components/notebooks/",
+
+    "public/docs/components/pipelines/",
+    "public/docs/gke/pipelines/authentication-pipelines",
+    "public/docs/gke/pipelines/upgrade",
+
+    "public/docs/components/katib/",
+    "public/docs/components/kfserving/",
+    "public/docs/components/training/",
+
+    "public/docs/distributions/gke/",
+    "public/docs/gke/deploy/oauth-setup",
+
+    "public/docs/distributions/aws/",
+    "public/docs/distributions/ibm/",
+
+    "public/fonts/glyphicons",
+    "public/docs/experiment",
+    "public/docs/suggestion",
+    "public/docs/trial",
+    "public/docs/examples/shared-resources",
+    "public/docs/started/getting-started",
+    "public/docs/other-guides",
+    "public/docs/external-add-ons/istio",
+    "public/404.html",
+  ]
+
 [context.deploy-preview.environment]
   HUGO_VERSION = "0.124.1"
   NODE_VERSION = "18"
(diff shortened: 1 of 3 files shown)


thesuperzapper:
@hbelmiro I agree with this change, but can you make two changes for safety:

1. Pin a specific version of the plugin in our `package.json` (Netlify says they will use the version specified in `devDependencies`) as this reduces supply chain risk:
      - You can run `npm install -D netlify-plugin-checklinks` in the root of the repo and commit the updates.
3. Move the plugin to only run on the `deploy-preview` context, so we dont break updates to old release branches, or have a build fail if we force a merge on master with broken links:
      - You can do this by replacing `[[plugins]]` in your `netlify.toml` with `[[context.deploy-preview.plugins]]`

thesuperzapper (changes requested):
@hbelmiro I agree with this change, but can you make two changes for safety:

1. Pin a specific version of the plugin in our `package.json` (Netlify says they will use the version specified in `devDependencies`) as this reduces supply chain risk:
      - You can run `npm install -D netlify-plugin-checklinks` in the root of the repo and commit the updates.
2. Move the plugin to only run on the `deploy-preview` context, so we dont break updates to old release branches, or have a build fail if we force a merge on master with broken links:
      - You can do this by replacing `[[plugins]]` in your `netlify.toml` with `[[context.deploy-preview.plugins]]`

hbelmiro:
> @hbelmiro I agree with this change, but can you make two changes for safety:
> 
>     1. Pin a specific version of the plugin in our `package.json` (Netlify says they will use the version specified in `devDependencies`) as this reduces supply chain risk:
>        
>        * You can run `npm install -D netlify-plugin-checklinks` in the root of the repo and commit the updates.
> 
>     2. Move the plugin to only run on the `deploy-preview` context, so we dont break updates to old release branches, or have a build fail if we force a merge on master with broken links:
>        
>        * You can do this by replacing `[[plugins]]` in your `netlify.toml` with `[[context.deploy-preview.plugins]]`

@thesuperzapper I addressed your suggestions.
Thank you.

netlify.toml
```diff
@@ -2,6 +2,42 @@
...
+
+    "public/fonts/glyphicons",
+    "public/docs/experiment",
+    "public/docs/suggestion",
+    "public/docs/trial",
+    "public/docs/examples/shared-resources",
+    "public/docs/started/getting-started",
+    "public/docs/other-guides",
+    "public/docs/external-add-ons/istio",
+    "public/404.html",
```
thesuperzapper, line 38: @hbelmiro my only question is what does ignoring `public/404.html` do?

hbelmiro: I don't remember why I added that.
I just removed it and confirmed it's not needed.

thesuperzapper:
@james-jwu this should be good to merge now (can you approve it), it will be great to check links automatically and avoid 404s.

@hbelmiro I assume you will make follow up PRs to remove the whitelist of links?

/lgtm

hbelmiro:
> @hbelmiro I assume you will make follow up PRs to remove the whitelist of links?

I'll do my best. I may not know the proper links for the components I'm unfamiliar with.
I plan to fix at least those in the Pipelines component and open issues for the other ones. The idea is to have the whitelist empty, or close to it.

hbelmiro:
@thesuperzapper ^

PR #3780, opened 2024-06-24
Update Kubeflow diagrams with Spark Operator and Model Registry

andreyvelich:
I updated Kubeflow diagrams with Spark Operator and Model Registry, since we already have these components on Kubeflow website.

/assign @kubeflow/wg-data-leads @kubeflow/kubeflow-steering-committee @vikas-saxena02


tarilabs:
Thank you so much @andreyvelich 

In `content/en/docs/started/images/kubeflow-intro-diagram.drawio.svg`
I would suggest rename the in-parenthesis from:
_(ML Metadata)_
to:
_(Model Metadata)_

so to keep the generic scope intended by the parenthesis, and avoid potential confusion with the "MLMD" (library/project).

wdyt ?

thesuperzapper:
@andreyvelich I agree that we should say `"Model Metadata"` rather than `"ML Metadata"` (which is the name of an actual project that we use as a dependency: https://github.com/google/ml-metadata

Also, can you fix the alignment problem on the "Feast" square, it drives me crazy:


Screenshot 2024-06-24 at 15 53 44

vikas-saxena02:
great work @andreyvelich you just depicted my precise thoughts here. 

/approve

andreyvelich:
Sounds good, updated.
/assign @tarilabs @thesuperzapper

tarilabs (approved):
/lgtm

thank you so much

Issue #3766, opened 2024-06-16
Incorrect links in V2 / data types / artifacts documentation page


ichbinjakes:
The links in these two paragraphs for "Python Components" and "Container Components" are for the wrong pages. Additionally, it is unclear from the text if python components refers to lightweight AND containerised python components.

https://github.com/kubeflow/website/blob/6be88bd593f4069183731bd5732aab5069986c1b/content/en/docs/components/pipelines/v2/data-types/artifacts.md?plain=1#L78-L84

hbelmiro:
> it is unclear from the text if python components refers to lightweight AND containerised python components.

@ichbinjakes I don't understand what is the problematic text. Can you please highlight the statement that leaves the above unclear.

rimolive:
This is where the KFP docs describes the concept of a component.

https://www.kubeflow.org/docs/components/pipelines/concepts/component/

PR #3785, opened 2024-07-01
Fixed links in Pipelines > user guides > data handling > artifacts

hbelmiro:
Resolves: https://github.com/kubeflow/website/issues/3766


hbelmiro:
cc @chensun

PR #3788, opened 2024-07-02
misc updates to KF Model Registry tutorial

tarilabs:
- add suggestions of commands to dry-run connection from Notebook, this for instance was a feedback emerged interacting with users in https://github.com/kubeflow/model-registry/issues/90#issuecomment-2143788711
- refresh versions, by pinning to the latest version (and not just `main`, per previous KF community leads request)
- to make the tutorial flow easier, align model-version used for the mnist example ML model


rareddy (approved):
/lgtm
/approve

PR #3789, opened 2024-07-04
Pipelines - @hbelmiro promoted to approver and removed old members

hbelmiro:
cc @rimolive


rimolive:
/approve

Thanks @hbelmiro for all the work in the KFP docs!

PR #3801, opened 2024-07-10
MR: update getting started

isinyaaa:


isinyaaa:
cc: @tarilabs

Arhell (commented):
/ok-to-test

tarilabs (commented):
This changes looks good to me, thank you @isinyaaa but I need to wait for 0.2.3

tarilabs:
With the additional changes of: https://github.com/isinyaaa/kubeflow-website/pull/2

Tested all tutorial:
- 
Screenshot from 2024-07-15 08-37-56
- 
Screenshot from 2024-07-15 08-38-22
- 
Screenshot from 2024-07-15 08-45-39
- 
[getting-started.ipynb.zip](https://github.com/user-attachments/files/16230909/getting-started.ipynb.zip)


Will wait for https://github.com/isinyaaa/kubeflow-website/pull/2 to be merged, then we can proceed on this one I think, thanks @isinyaaa

tarilabs (commented):
https://github.com/kubeflow/website/pull/3801#issuecomment-2227812630

/lgtm

rareddy:
"get_model_version" This method does not sound right to me, IMO this should have been "get_versioned_model"  or "get_model_by_versison", this can be another PR but wanted to mention.

Are snake "_" style APIs typical in Python APIs? TBH I do not like them :) This is time to fix any APIs before we officially do a release.

tarilabs:
This definitely the right case (Python is snake case) and is the right naming, since the logical model entity name is "ModelVersion", but thank you for noticing!

rareddy:
ok, with snake style (but do not like it).  IMO high-level APIs need to be descriptive not necessarily expose as is the data model in the background. Name closer to user actions.

tarilabs:
👍 

It is the correct action, as that is the Logic Model entity: https://github.com/kubeflow/model-registry/blob/main/docs/logical_model.md#model-version
which is the same concepts for UI mocks.

We are not leaking in this API the fact the actual data behind the scenes is a MLMD context.

PR #3818, opened 2024-07-22
Generate social embed images for all pages

thesuperzapper:
On social media like Slack, LinkedIn, Facebook, etc. the Kubeflow website does not display an embedded image.

For example:

Screenshot 2024-07-21 at 19 11 24

This PR makes it so that each page has a unique image (with the title of the page) to show on social media sites.
This is really important for people wanting to share about Kubeflow publicly.

For example, here are some of the generated images:

social 62fafe114d1c92485c6088c3efd27ce2

social 8fc3e9355cc74412ab45fdd6a3332bad

As new pages are added, this system will automatically generate new social images.

PR #3821, opened 2024-07-22
Fix baseURL in Netlify deployments (needed to fix `twitter:image`)

thesuperzapper:
This PR is a follow up to https://github.com/kubeflow/website/pull/3818

It seems like [Twitter wants full URLs](https://devcommunity.x.com/t/card-error-unable-to-render-or-no-image-read-this-first/62736)  in the `<meta name="twitter:image" content=".......">` tags, that is, including the `https://DOMAIN/` prefix. While this is what the [`PAGE.Permalink`](https://gohugo.io/methods/page/permalink/) template in Hugo is meant to do, this only works if the Hugo [`baseURL`](https://gohugo.io/methods/site/baseurl/) config is set.

This PR will set `baseURL` for:

- production deployments to `https://www.kubeflow.org/`
- deploy previews (like this PR) to [`DEPLOY_PRIME_URL`](https://docs.netlify.com/configure-builds/environment-variables/#deploy-urls-and-metadata), an environment variable provided by Netlify that corresponds to the `https://deploy-preview-PR_NUMBER--competent-brattain-de2d6d.netlify.app/` URL

I have also done the following to increase the chance that websites pick up the social images:

1. Re-ordered the `<meta>` tags so each kind is beside each other (e.g. `og:xxxx` and `twitter:xxxx` are near each other)
2. Fixed the `<script type="application/ld+json">` which was not formatted correctly before.
3. Set the `HUGO_ENV` to "development" for non-prod deployments (so that robots don't scrape the images from anything but `kubeflow.org`)

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -117,9 +117,8 @@ enable = true
   github_project_repo = "https://github.com/kubeflow/kubeflow"
 
   RSSLink = "/index.xml"
-  author = "kubeflow.org" # add your company name
-  github = "kubeflow" # add your github profile name
-  twitter = "kubeflow" # add your twitter profile
+  author = "kubeflow.org"
+  github = "kubeflow"
 
   copyright = "The Kubeflow Authors."
   privacy_policy = "https://policies.google.com/privacy"
@@ -151,6 +150,12 @@ enable = true
   # Disable MathJax by default
   # NOTE: enable it per-page with `mathjax: true` in front matter
   mathjax = false
+
+  # Social media accounts
+  [params.social]
+
+    # Twitter account (used to set `twitter:site` in the SEO partial)
+    twitter = "kubeflow"
 
   # These entries appear in the drop-down menu at the top of the website.
   [[params.versions]]
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -22,14 +22,12 @@
   {{ end -}}
 </title>
 <meta name="description" content="{{ template "partials/page-description.html" . }}">
-{{ template "_internal/opengraph.html" . -}}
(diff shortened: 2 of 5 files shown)


thesuperzapper:
@terrytangyuan this should fix the social embeds on Twitter now.

The issue was that it needs absolute URLs for the images, otherwise it fails silently. Getting that to work required a few other changes so that Hugo + Netlify actually know what the URL is.

I can't show you a preview of it working on X because the non-prod site sets `X-Robots-Tag=noindex` on the `.png` files, and X seems to respect that request.

I think we can merge this, check X works, and also monitor to ensure the site does not break (we can make a quick revert if anything is not working).

terrytangyuan (commented):
I am not familiar with the implementation so I would prefer this to be reviewed by someone else

thesuperzapper:
@james-jwu since this affects Netlify could you take a look.

I know the PR might look intimidating, but the key changes are just that Netlify knows the "base URL" the website is going to be hosted on. This allows us to use the full permalink in our social image meta tags, which is required for Twitter to show the image.

We also update the `ld+json` schema, which is just structured data used by Google to improve the index. 

Those schema updates are not technically not required to fix Twitter embeds, but we might as well fix them while we are changing the meta tags, because they are currently not formatted correctly, so are ignored by Google.

james-jwu:
The changes make sense although I can't say I understood it all :). Approving it for now. Please monitor the changes and fix any issues that may come up as a result @thesuperzapper 
/lgtm
/approve

thesuperzapper:
Seems twitter works now:

Screenshot 2024-07-23 at 15 53 00

Issue #3829, opened 2024-07-26
Fix missing google_analytics_async.html template preventing hugo to run locally.


franciscojavierarceo:
I was getting started with the website and following the instructions on [running hugo locally](https://github.com/kubeflow/website/?tab=readme-ov-file#run-local-hugo-server) and I received the following error:

```zsh
[master]% hugo server -D
Watching for changes in /Users/farceo/dev/website/{assets,content,layouts,package.json,static,themes}
Watching for config changes in /Users/farceo/dev/website/config.toml, /Users/farceo/dev/website/themes/docsy/config.yaml
Start building sites …
hugo v0.129.0+extended darwin/arm64 BuildDate=2024-07-17T13:29:16Z VendorInfo=brew

ERROR render of "section" failed: "/Users/farceo/dev/website/themes/docsy/layouts/docs/baseof.html:4:7": execute of template failed: template: docs/list.html:4:7: executing "docs/list.html" at <partial "head.html" .>: error calling partial: execute of template failed: html/template:partials/head.html:65:16: no such template "_internal/google_analytics_async.html"
ERROR render of "page" failed: "/Users/farceo/dev/website/themes/docsy/layouts/docs/baseof.html:4:7": execute of template failed: template: docs/single.html:4:7: executing "docs/single.html" at <partial "head.html" .>: error calling partial: execute of template failed: html/template:partials/head.html:65:16: no such template "_internal/google_analytics_async.html"
Built in 345 ms
Error: error building site: render: failed to render pages: render of "page" failed: "/Users/farceo/dev/website/themes/docsy/layouts/docs/baseof.html:4:7": execute of template failed: template: docs/single.html:4:7: executing "docs/single.html" at <partial "head.html" .>: error calling partial: execute of template failed: html/template:partials/head.html:65:16: no such template "_internal/google_analytics_async.html"
```

Looking at the [_internal/](https://github.com/kubeflow/website/tree/master/layouts/_internal) path I see that the `google_analytics_async.html` file doesn't exist. 

I added an empty file and then things worked just file but I wanted to confirm that this wasn't on purpose. 

The fix would be to delete the references to the template or add the file with the real template. Happy to make a fix.

thesuperzapper:
@franciscojavierarceo I don't get an issue when using the correct version of Hugo.

[...]

PR #3831, opened 2024-07-29
Issue 3829: Remove missing template preventing local build

franciscojavierarceo:
As outlined in https://github.com/kubeflow/website/issues/3829, the missing `_internal/google_analytics_async.html ` script fails the local startup of the hugo server (as outlined in the docs).

I've removed the template code reference as it's not in the repo.

Let me know if this isn't the correct thing, happy to revise or update the docs accordingly.

I was also able to get the huge server to start up locally by just adding an empty file but I suspect that's not the correct solution.

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -61,7 +61,5 @@
   {{ $enableGtagForUniversalAnalytics := not .Site.Params.disableGtagForUniversalAnalytics -}}
   {{ if (or $enableGtagForUniversalAnalytics (hasPrefix .Site.GoogleAnalytics "G-")) -}}
     {{ template "_internal/google_analytics_gtag.html" . -}}
-  {{ else -}}
-    {{ template "_internal/google_analytics_async.html" . -}}
   {{ end -}}
 {{ end -}}


layouts/partials/head.html
```diff
@@ -61,7 +61,5 @@
   {{ $enableGtagForUniversalAnalytics := not .Site.Params.disableGtagForUniversalAnalytics -}}
   {{ if (or $enableGtagForUniversalAnalytics (hasPrefix .Site.GoogleAnalytics "G-")) -}}
     {{ template "_internal/google_analytics_gtag.html" . -}}
-  {{ else -}}
-    {{ template "_internal/google_analytics_async.html" . -}}
```
terrytangyuan, line 65: Not sure if we actually need this file. @thesuperzapper Do you know?

franciscojavierarceo: Yeah I mentioned that I was uncertain about this as the fix in my issue.

thesuperzapper:
@franciscojavierarceo @terrytangyuan I think the issue is probably just that you are running a version of Hugo that is too new.

See my comment in https://github.com/kubeflow/website/issues/3829#issuecomment-2266224692.

franciscojavierarceo:
Thanks, Mathew. I'll double check. I thought I followed the instructions correctly but looks like I may have missed it. I'll confirm and close. Apologies!

thesuperzapper:
@franciscojavierarceo ok the issue was just that you had not run `git submodule update --init --recursive`, so were running an old version of the Docsy template.

/close

PR #3848, opened 2024-09-04
Block web crawlers on `v0.2-branch`

thesuperzapper:

This PR adds the [`<meta name="robots" content="noindex">` meta tag](https://developers.google.com/search/docs/crawling-indexing/block-indexing) to tell Google to stop indexing the pages from the `v0.2-branch` branch.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,6 +4,8 @@
 
 [context.deploy-preview.environment]
   HUGO_VERSION = "0.40"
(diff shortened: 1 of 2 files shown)


jlewi:
I don't think I can approve this but here goes.
/approve

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #3852, opened 2024-09-04
Block web crawlers on `v0.7-branch`

thesuperzapper:

This PR adds the [`<meta name="robots" content="noindex">` meta tag](https://developers.google.com/search/docs/crawling-indexing/block-indexing) to tell Google to stop indexing the pages from the `v0.7-branch` branch.

---

It also backports the changes from https://github.com/kubeflow/website/pull/3863 so the version selector is consistent across each archive version.

And updates the OWNERS file of this very old branch to align with [the current `OWNERS` in `master`](https://github.com/kubeflow/website/blob/89d8f792625e6c95fad4774570358ecd773830e7/OWNERS) to make it easier to approve PRs in the future without GitHub admin overrides.


thesuperzapper:
The build is actually succeeding on this one, it's just that there is a random other Netlify build which is failing.

The only one that matters is the main one, so this can be merged.

PR #3862, opened 2024-09-04
Block web crawlers on `v1.9-branch`

thesuperzapper:

This PR adds the [`<meta name="robots" content="noindex">` meta tag](https://developers.google.com/search/docs/crawling-indexing/block-indexing) to tell Google to stop indexing the pages from the `v1.9-branch` branch.

---

It also backports the changes from https://github.com/kubeflow/website/pull/3863 so the version selector is consistent across each archive version.

----

__You might be wondering, why should we do this for the 1.9 branch, isn't that the latest version?__

1. There is still no reason for google to index these pages:
      - The latest content is on the main site.
      - Google will take a long time to remove things from the index if we add this metadata at a later date.
3. The `"The site that you are currently viewing is an archived snapshot"` header is applicable:
      - It's a snapshot from the time of the 1.9 release.
      - We don't typically make any further updates to the versioned branches. 
      - It's hard for users to realize they ended up on a snapshotted version of the docs without that warning header.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -129,7 +129,7 @@ enable = true
   gcs_engine_id = "007239566369470735695:624rglujm-w"
 
   # Text label for the version menu in the top bar of the website.
-  version_menu = "Kubeflow Version"
+  version_menu = "Archive: 1.9"
 
   # The major.minor version tag for the version of the docs represented in this
   # branch of the repository. Used in the "version-banner" partial to display a
@@ -142,7 +142,7 @@ enable = true
 
   # A link to latest version of the docs. Used in the "version-banner" partial to
   # point people to the main doc site.
-  url_latest_version = "https://kubeflow.org/docs/"
+  url_latest_version = "https://www.kubeflow.org/docs/"
 
   # A variable used in various docs to determine URLs for config files etc.
   # To find occurrences, search the repo for 'params "githubbranch"'.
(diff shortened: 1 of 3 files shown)


config.toml
```diff
@@ -129,88 +129,92 @@ enable = true
...
   # NOTE: enable it per-page with `mathjax: true` in front matter
   mathjax = false
 
   # These entries appear in the drop-down menu at the top of the website.
   [[params.versions]]
-    version = "master"
+    version = "Latest"
     githubbranch = "master"
-    url = "https://master.kubeflow.org"
+    url = "https://www.kubeflow.org"
```
ederign, line 159: Out of curiosity, do the prefix change anything on google indexing?

ederign: Or did you change the other PRs for the sake of consistency?

thesuperzapper: Google prefers direct links, and only adds to the "backlinks" count for real links, not ones that are redirected (at least that's my experience).

thesuperzapper:
We will still eventually merge this once 1.10 is released.

/reopen

thesuperzapper:
@varodrig this one will need to be merged to stop google indexing the old 1.9 branch.

thesuperzapper:
closing in favor of https://github.com/kubeflow/website/pull/4120

/close

PR #3849, opened 2024-09-04
Block web crawlers on `v0.4-branch`

thesuperzapper:
Google keeps surfacing links from very old versions of the Kubeflow docs website, which is confusing to users.

This PR adds the [`<meta name="robots" content="noindex">` meta tag](https://developers.google.com/search/docs/crawling-indexing/block-indexing) to tell Google to stop indexing the pages from the `v0.4-branch` branch.

It will take probably a few months (or longer) for Google to re-index these pages, I have left `nofollow` off, so that google will re-index the whole site quicker by following links.

---

It also backports the changes from https://github.com/kubeflow/website/pull/3863 so the version selector is consistent across each archive version.

PR #3863, opened 2024-09-04
Improve version selector & rename "master" to "latest"

thesuperzapper:
In the top right of the website, we have a "version selector" that lets you view archived versions of the docs from past releases, however, the current presentation is confusing to users.

New users expect that they should browse the site on the 1.9 branch for information about 1.9 (when in reality they should browse the master branch because we are always updating it).

In reality, we treat these historical snapshot versions as "snapshot archives" and don't go back to update them with content from master.

__This PR makes the following changes to the version menu:__

- Renamed the button which you click to expand the menu:
    - For example, in this PR its `"Latest"`, but in the `v1.0-branch` snapshot it would be `"Archive: 1.0"`
- prepended `Archive: ` to each version label
- removes versions older than 0.6 from the dropdown

# Screenshots

## OLD

Screenshot 2024-09-04 at 17 58 06

## NEW

Screenshot 2024-10-15 at 09 00 03

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -133,26 +133,33 @@ enable = true
   # Google Custom Search Engine ID.
   gcs_engine_id = "007239566369470735695:624rglujm-w"
 
-  # Text label for the version menu in the top bar of the website, depends on the branch:
-  #  - master -> "Latest"
-  #  - v1.0-branch -> "v1.0"
-  version_menu = "Latest"
+  # The text label for the version menu in the top bar.
+  # The value depends on the branch:
+  #  - `master` -> "Versions"
+  #  - `v1.0-branch` -> "v1.0"
+  version_menu = "Versions"
 
-  # The major.minor version tag for the version of the docs represented in this
-  # branch of the repository. Used in the "version-banner" partial to display a
-  # version number for this doc set.
+  # The version of the docs. This is used in the "version-banner" partial.
+  # The value depends on the branch:
+  #  - `master` -> "REPLACE_WITH_VERSION_ON_ARCHIVE"
+  #  - `v1.0-branch` -> "v1.0"
   version = "REPLACE_WITH_VERSION_ON_ARCHIVE"
 
-  # Flag used in the "version-banner" partial to decide whether to display a
-  # banner on every page indicating that this is an archived version of the docs.
+  # If the "version-banner" partial should display at the top of each page.
+  # The value depends on the branch:
+  #  - `master` -> false
+  #  - `v*-branch` -> true
   archived_version = false
 
-  # A link to latest version of the docs. Used in the "version-banner" partial to
-  # point people to the main doc site.
+  # The URL of the latest version of the docs.
+  # Used in the "version-banner" partial.
   url_latest_version = "https://www.kubeflow.org/docs/"
 
-  # A variable used in various docs to determine URLs for config files etc.
-  # To find occurrences, search the repo for 'params "githubbranch"'.
+  # The GitHub branch for the current version of the docs.
+  # Used to generate the "Edit this page" link.
+  # The value depends on the branch:
+  #  - `master` -> "master"
+  #  - `v1.0-branch` -> "v1.0-branch"
   github_branch = "master"
 
   # Disable MathJax by default
@@ -214,14 +221,6 @@ enable = true
     version = "v1.0"
     githubbranch = "v1.0-branch"
     url = "https://v1-0-branch.kubeflow.org"
-  [[params.versions]]
-    version = "v0.7"
-    githubbranch = "v0.7-branch"
-    url = "https://v0-7.kubeflow.org"
-  [[params.versions]]
-    version = "v0.6"
-    githubbranch = "v0.6-branch"
-    url = "https://v0-6.kubeflow.org"
 
   # User interface configuration
   [params.ui]
--- a/config.toml
+++ b/config.toml
@@ -222,18 +222,6 @@ enable = true
     version = "Archive: 0.6"
     githubbranch = "v0.6-branch"
     url = "https://v0-6.kubeflow.org"
-  [[params.versions]]
-    version = "Archive: 0.5"
-    githubbranch = "v0.5-branch"
-    url = "https://v0-5.kubeflow.org"
-  [[params.versions]]
-    version = "Archive: 0.4"
-    githubbranch = "v0.4-branch"
-    url = "https://v0-4.kubeflow.org"
-  [[params.versions]]
-    version = "Archive: 0.3"
-    githubbranch = "v0.3-branch"
-    url = "https://v0-3.kubeflow.org"
 
   # User interface configuration
   [params.ui]
--- a/config.toml
+++ b/config.toml
@@ -133,13 +133,15 @@ enable = true
   # Google Custom Search Engine ID.
   gcs_engine_id = "007239566369470735695:624rglujm-w"
 
-  # Text label for the version menu in the top bar of the website.
-  version_menu = "Version"
+  # Text label for the version menu in the top bar of the website, depends on the branch:
+  #  - master -> "Latest"
+  #  - v1.0-branch -> "Archive: 1.0"
+  version_menu = "Latest"
 
   # The major.minor version tag for the version of the docs represented in this
   # branch of the repository. Used in the "version-banner" partial to display a
   # version number for this doc set.
-  version = "master"
+  version = "REPLACE_WITH_VERSION_ON_ARCHIVE"
 
   # Flag used in the "version-banner" partial to decide whether to display a
   # banner on every page indicating that this is an archived version of the docs.
@@ -147,7 +149,7 @@ enable = true
 
   # A link to latest version of the docs. Used in the "version-banner" partial to
   # point people to the main doc site.
-  url_latest_version = "https://kubeflow.org/docs/"
+  url_latest_version = "https://www.kubeflow.org/docs/"
 
   # A variable used in various docs to determine URLs for config files etc.
   # To find occurrences, search the repo for 'params "githubbranch"'.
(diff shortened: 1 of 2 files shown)


andreyvelich (commented):
Thank you for creating those PRs @thesuperzapper.

From my point of view, we should not add `Archive: ` message to the previous Kubeflow versions since it can cause false impression that we don't provide support for those versions.
I would suggest that we follow Kubernetes way to show versions: https://kubernetes.io/.
E.g. only leave 5 latests version of Kubeflow in the dropdown menu. 

@hbelmiro @kubeflow/kubeflow-steering-committee @StefanoFioravanzo Any thoughts?

thesuperzapper:
> Thank you for creating those PRs @thesuperzapper.
> 
> From my point of view, we should not add `Archive: ` message to the previous Kubeflow versions since it can cause false impression that we don't provide support for those versions.

@andreyvelich the idea of the word "archive" is that it's an "archive" of the docs at that version, not that we don't support that version anymore.

Also, even if this was about support for the version itself, we definitely don't support anything other than 1.9.

@andreyvelich the idea of the word "archive" is that it's an "archive" of the docs at that version, not that we don't support that version anymore.

Even if this was referring to support for the version itself, we definitely don't support anything other than 1.9.

Also, when you go to any of those old versions of the website, every page has something which explains this at the top (that you're viewing a snapshot, not the latest version), so it should be clear.

> I would suggest that we follow Kubernetes way to show versions: https://kubernetes.io/.
> E.g. only leave 5 latests version of Kubeflow in the dropdown menu.

I don't really mind, but people might want to be able to see the older versions and they won't be able to find them otherwise. 

I think we should probably keep at least all of the versions post 1.0.

hbelmiro:
What about the following?

- ~~master~~ -> 1.9
- ~~1.9~~
- 1.8
- 1.7
- 1.6
- 1.5
- 1.4
- 1.3
- 1.2
- 1.1
- 1.0

1.9 would be the latest docs version that comes from the `master` branch.
I like the idea of following what Kubernetes does and leaving all the post-1.0 versions.

andreyvelich:
>@andreyvelich the idea of the word "archive" is that it's an "archive" of the docs at that version, not that we don't support that version anymore.

@thesuperzapper What is the user value to add `archive` message to the versions that we display on the website ?

>Even if this was about support for the version itself, we definitely don't support anything other than 1.9.

Why we don't support anything older than 1.9 ? I think, that similar to Kubernetes we should provide support to the few most recent Kubeflow releases and provide critical bug fixes for that release. That will show the project stability and maturity for the end users.

>I don't really mind, but people might want to be able to see the older versions and they won't be able to find them otherwise.

We can have dedicated page for old Kubeflow versions, like in Kubernetes: https://kubernetes.io/docs/home/supported-doc-versions/#versions-older

thesuperzapper:
> > @andreyvelich the idea of the word "archive" is that it's an "archive" of the docs at that version, not that we don't support that version anymore.
> 
> @thesuperzapper What is the user value to add `archive` message to the versions that we display on the website ?

Very simply:

1. Those `v1.X-branch...` website are archives, in the sense that they are the state of the docs at the time of the release.
2. Users are best served by reading the `master`/`latest` version of the website.
3. It helps users understand quickly that this is not a "versioned docs site" like they are probably expecting.

Do you have an alternative word that we could use as a prefix? 

If you can't stand the "archive" prefix (and we can't think of an alternative), I guess as long as all the archive sites highlight at the top that you are not looking at the latest version (which will be the case after the other PRs are merged), we could drop it.

> > Even if this was about support for the version itself, we definitely don't support anything other than 1.9.
> 
> Why we don't support anything older than 1.9 ? I think, that similar to Kubernetes we should provide support to the few most recent Kubeflow releases and provide critical bug fixes for that release. That will show the project stability and maturity for the end users.

While that is nice in theory, it's a separate discussion to this as we currently don't support anything but 1.9.0.

This PR is not about the actual support status of each version, but helping users understand that they are looking at a docs "snapshot/archive".

andreyvelich:
>Users are best served by reading the master/latest version of the website.

What if some new features that we deliver are not compatible with the old version of the components ?
In that case, users who are using old version of this Kubeflow component have to read previous version of Kubeflow website.

>While that is nice in theory, it's a separate discussion to this as we currently don't support anything but 1.9.0.

This is not 100% true, at least from the @kubeflow/wg-training-leads side we can fix the critical bugs on the previous versions (e.g. three latest releases) of Training Operator if that was asked by users.

@thesuperzapper Why the Kubernetes approach doesn't work for us ?

I want to get @StefanoFioravanzo feedback on this given that he spent time to think about how we can develop the Kubeflow website given that every Kubeflow sub-project has dedicated release cycle and version.

thesuperzapper:
@andreyvelich really, we just need to make a decision on if the `"Achive: "` prefix will be present or not.

The other PRs are much more important and I don't want to let this discussion block them, we need to prevent Google from indexing the very old websites (right now, Kubeflow search results often surface results from 0.3/0.6 versions of the docs):

- https://github.com/kubeflow/website/pull/3849
- https://github.com/kubeflow/website/pull/3850
- https://github.com/kubeflow/website/pull/3851
- https://github.com/kubeflow/website/pull/3852
- https://github.com/kubeflow/website/pull/3853
- https://github.com/kubeflow/website/pull/3854
- https://github.com/kubeflow/website/pull/3855
- https://github.com/kubeflow/website/pull/3856
- https://github.com/kubeflow/website/pull/3857
- https://github.com/kubeflow/website/pull/3858
- https://github.com/kubeflow/website/pull/3859
- https://github.com/kubeflow/website/pull/3860
- https://github.com/kubeflow/website/pull/3861
- https://github.com/kubeflow/website/pull/3862

Once we agree on if the "Archive: " prefix will be present, I will update these PRs to reflect that (so all versions are consistent), but its way more important to stop them appearing in search results.

terrytangyuan:
Hmm https://github.com/kubeflow/website/pull/3849 seems merged already

thesuperzapper:
> Hmm #3849 seems merged already

@terrytangyuan Either way, its easy to update now that I have fixed building.

andreyvelich:
>The other PRs are much more important and I don't want to let this discussion block them, we need to prevent Google from indexing the very old websites (right now, Kubeflow search results often surface results from 0.3/0.6 versions of the docs):

@thesuperzapper Can you remove the version change from your PRs and only leave the change to prevent Google from indexing and we can merge those PRs ?

andreyvelich:
cc @ederign

thesuperzapper:
@andreyvelich I have updated all the other PRs with the following changes:

1. Remove the "Archive: " prefix from the "versions" menu that I added before
2. Update the root OWNERS to align with the current `master` (KSC members)
3. Updated the "version banner" to look like this:

Screenshot 2024-09-06 at 14 42 51

thesuperzapper:
We now need someone with GitHub org admin __manually merge__ the following PRs:

- A manual merge is needed because the old OWNERs are not active anymore.
     - __WARNING:__ please remember to use SQUASH merge!
- https://github.com/kubeflow/website/pull/3869
- https://github.com/kubeflow/website/pull/3868
- https://github.com/kubeflow/website/pull/3867
- https://github.com/kubeflow/website/pull/3850
- https://github.com/kubeflow/website/pull/3851
- https://github.com/kubeflow/website/pull/3852
- https://github.com/kubeflow/website/pull/3853
- https://github.com/kubeflow/website/pull/3854
- https://github.com/kubeflow/website/pull/3855
- https://github.com/kubeflow/website/pull/3856
- https://github.com/kubeflow/website/pull/3857
- https://github.com/kubeflow/website/pull/3858
- https://github.com/kubeflow/website/pull/3859
- https://github.com/kubeflow/website/pull/3860
- https://github.com/kubeflow/website/pull/3861
- https://github.com/kubeflow/website/pull/3862

thesuperzapper:
@james-jwu or @zijianjoy can you please update these DNS records, as they don't point to our modern Netlify, so will not be updated otherwise:

1. `v0-2.kubeflow.org`:
    - OLD `CNAME`: `eager-nobel-1708cf.netlify.com`
    - NEW `CNAME`: `v0-2-branch--competent-brattain-de2d6d.netlify.app`
2. `v0-3.kubeflow.org`:
    - OLD `CNAME`: `competent-jones-a15cc2.netlify.com`
    - NEW `CNAME`: `v0-3-branch--competent-brattain-de2d6d.netlify.app`
3. `v0-4.kubeflow.org`:
    - OLD `CNAME`: `infallible-hermann-cb44b2.netlify.com`
    - NEW `CNAME`: `v0-4-branch--competent-brattain-de2d6d.netlify.app`
4. `v0-5.kubeflow.org`:
    - OLD `CNAME`: `practical-neumann-e70801.netlify.com`
    - NEW `CNAME`: `v0-5-branch--competent-brattain-de2d6d.netlify.app`

andreyvelich:
@kubeflow/kubeflow-steering-committee @kubeflow/wg-pipeline-leads @kubeflow/wg-data-leads @kubeflow/wg-training-leads @kubeflow/wg-notebooks-leads @kubeflow/wg-manifests-leads Can you give +1 on the banner that @thesuperzapper proposed for the old docs ?

If that looks good to you, we can start merging the PRs on the old website branches.

andreyvelich:
@thesuperzapper Only one comment from me, that I don't think that we should add yellow banner on the 1.9 documentation, since it is the latest stable release.
As you can see, the Kubernetes doesn't have this banner on the latest v1.31 version:
- https://kubernetes.io/docs/concepts/workloads/controllers/job/
- https://v1-30.docs.kubernetes.io/docs/concepts/workloads/controllers/job/

thesuperzapper:
> @thesuperzapper Only one comment from me, that I don't think that we should add yellow banner on the 1.9 documentation, since it is the latest stable release. As you can see, the Kubernetes doesn't have this banner on the latest v1.31 version:
> 
> * https://kubernetes.io/docs/concepts/workloads/controllers/job/
> * https://v1-30.docs.kubernetes.io/docs/concepts/workloads/controllers/job/
> 
> WDYT @thesuperzapper ?

@andreyvelich on the Kubernetes docs, "1.31" is actually a link to `kubernetes.io`, which is their `master` branch.

The problem is that we have already cut a `v1.9-branch` for "1.9", so it really is a `"snapshot of the docs at the time of the 1.9 release"`. If we don't have the warning, people won't realize they are not viewing the updated version of the 1.9 docs (which live on `master`).

In the future, I think we should hold off on cutting the 1.10 branch until __just before__ we start merging docs for the 1.11 release. To resolve the current situation, we have two options:

1. OPTION 1: Keep the current `v1.9-branch` and retain the warning header.
2. OPTION 2: Delete the current `v1.9-branch` and only re-create it once we start the 1.10 release process

juliusvonkohout:
"While that is nice in theory, it's a separate discussion to this as we currently don't support anything but 1.9.0." yes there is just not much capacity beyond that. This might change if someone volunteers. We only fix critical things that are easy to fix, so in practice it was around 1 Year for 1.8/1.8.1 and with 1.9.1 released 1.8.1 will be EOL for most developers. For longer support you probably need distributions, consultants etc.

I like the term master more than latest, but I am also in favour of pushing for the master/latest version being shown by default and marking everything 2 releases away, so 1.8.1 and older as archived, End of Life or so.

tarilabs:
/lgtm

I can see the dropdown and the version banner in the linked PRs, thank you!

andreyvelich:
As we discussed during today's community call, we can manually **squash + merge** all PRs to disable indexing on old branches.
We should just have followup discussion for the PR on 1.9 branch.
@johnugeorge @terrytangyuan @james-jwu @jbottum Can you help to manually **squash + merge** these PRs if you are agree with the banner:
https://github.com/kubeflow/website/pull/3869
https://github.com/kubeflow/website/pull/3868
 #3867
https://github.com/kubeflow/website/pull/3850
https://github.com/kubeflow/website/pull/3851
https://github.com/kubeflow/website/pull/3852
https://github.com/kubeflow/website/pull/3853
https://github.com/kubeflow/website/pull/3854
https://github.com/kubeflow/website/pull/3855
https://github.com/kubeflow/website/pull/3856
https://github.com/kubeflow/website/pull/3857
https://github.com/kubeflow/website/pull/3858
https://github.com/kubeflow/website/pull/3859
https://github.com/kubeflow/website/pull/3860
https://github.com/kubeflow/website/pull/3861

james-jwu:
I've done the squash + merge.

zijianjoy:
Both the DNS configuration and the Kubeflow website hosting on Netlify are transferred to CNCF. Would you like to make a request to CNCF by submitting a service deck ticket? https://cncfservicedesk.atlassian.net/servicedesk/customer/portals

jbottum:
+1 for master rather than latest

juliusvonkohout:
> +1 for master rather than latest

This is also my opinion.

varodrig:
My opinion is to not use master, 
as github renamed master to main https://github.com/github/renaming why it was renamed https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

If we do not like latest, I vote to use **main**

PR #3881, opened 2024-09-27
[Pipelines] Specify KFP SDK version where new caching flag will be available

jgarciao:
As the changes introduced in https://github.com/kubeflow/pipelines/pull/11222 are in master but not in KFP 2.9.0 SDK, modify the _Use Caching_ page to specify when then new caching flag will be available


hbelmiro (commented):
/lgtm

DharmitD (commented):
/lgtm

PR #3842, opened 2024-08-23
Add version popout to "Packaged Distributions" table

thesuperzapper:
This PR improves the "Packaged Distributions" table by making it clearer how old/new each Kubeflow Version is. Right now, it's very confusing to new users, as they might not know that Kubeflow 1.7 is actually very old (released in March 2023), so they might start the process of adopting a long-abandoned distribution.

__The key changes this PR makes are:__

1. In the "Packaged Distributions" table under the "Kubeflow Version" column:
     - Put a hover-able icon that says how long ago that version of Kubeflow was released. (Note, this is the _Kubeflow_ released date, not the release date of the distribution).
     - ~~Highlight distributions which are up to date (within 1 minor version of the latest), and de-emphasize distributions which are 2+ minor versions out of date.~~
     - ~~Place warning icons beside versions which are very old (2 = warning, 3+ = scull)~~
2. Remove the very old link to the historical "conformance program" which was started by google but never continued.

__Please note, all of the styling is automatic:__

- when the short-codes for that distribution's version are updated everything will change __automatically__
- when we release a new version of Kubeflow, can use the `content/en/docs/started/installing-kubeflow/get_new_releases.sh` to cache that version's release date into the `content/en/docs/started/installing-kubeflow/release-info` folder of JSON (the build will fail to remind you if you try referencing a version which it does not know about)

# Screenshots

## OLD

Screenshot 2024-10-14 at 12 10 41

### Table Overview

Screenshot 2024-10-23 at 11 05 21

### On-Hover ("Latest Version")

Screenshot 2024-10-29 at 12 29 50

### On-Hover ("1 from Latest Version")

Screenshot 2024-10-29 at 12 29 20

### On-Hover ("3 from Latest Version")

Screenshot 2024-10-29 at 12 30 11

PR #3913, opened 2024-10-22
Bump Red Hat distribution to version 1.9

rimolive:
We would like to propose bump Red Hat distribution to Kubeflow 1.9 as part of updating the Distributions table.

## Why we are requesting this change?

Red Hat has been actively contributing with Kubeflow since 1.9 release in multiple Working Groups, including the new component: Model Registry.

Currently, Red Hat OpenShift AI uses the following components from Kubeflow 1.9:

* Pipelines
* Training Operator (upstream project currently)
* Model Registry
* Notebooks (in progress)

Our commitment with the community has confirmed by our initial contributions for the upcoming Kubeflow 1.10 release, and we've been working on Notebooks 2.0 (expected to be released along with Kubeflow 1.10) as well.

Components we are not considering or still evaluating:

* Central Dashboard (we implemented our own dashboard)
* Profile Controller/Volumes Web app (we use our own implementation under dashboard)
* Tensorboard Controller/Tensorboard Web app (we don't have plans to support)
* Katib (we are still evaluating adoption)

Because we don't have a clear process to evaluate if a Distribution is ready to support a given KF release, we understand that these arguments could be considered to state our current contribution level within the community. We are also happy to contribute with a concrete Conformance Testing process in which other distributions can work on a better workflow to promote their releases as well.


layouts/shortcodes/redhat/latest-version.html
```diff
@@ -1 +1 @@
-1.6.0
\ No newline at end of file
+1.9
```
thesuperzapper, line 1: The new system (https://github.com/kubeflow/website/pull/3842) requires you to specify a specific version patch (it removes the patch version from the outer table but shows the full version inside the pop-up).

Please make this 1.9.0.

thesuperzapper (changes requested):

thesuperzapper:
Also, do you have a "version matrix" page which we can link to that lists which versions of Kubeflow components are in each Open Data Hub version?

For example, we use this one on deployKF: 
- https://www.deploykf.org/releases/tool-versions/#kubeflow-ecosystem

Alternatively, please at least link to your overall platform changelog (like we do for the other distributions).

tarilabs (approved):
/lgtm

and works with me to specify micro version too

terrytangyuan (approved):
/lgtm
/approve

PR #3931, opened 2024-11-28
Update training operator installation to use SSA

astefanutti:
Latest version of the training operator will require to use server side apply to install its CRDs, as their sizes have increased beyond what client-side apply can handled.

See https://github.com/kubeflow/training-operator/pull/2332#discussion_r1860620021 for more details.

/cc @tenzen-y @andreyvelich


andreyvelich (commented):
Thanks @astefanutti!
/lgtm
/approve

PR #3976, opened 2025-01-24
Update user guide for namespace label in Katib Python SDK example

Doris-xm:
**What this PR does:**

Before creating Katib Experiments, the label for metrics collector injection `katib.kubeflow.org/metrics-collector-injection: enabled` should be presented in the namespace. 

This PR updates the user guide to inform users of the effect of this label.

**Which issue(s) this PR fixes:**
Fixes https://github.com/kubeflow/katib/issues/2499

### DIFF
--- a/content/en/docs/components/katib/getting-started.md
+++ b/content/en/docs/components/katib/getting-started.md
@@ -17,31 +17,6 @@ You need to install the following Katib components to run examples:
 
 ## Getting Started with Katib Python SDK
 
-### Adding Namespace Label for the Metrics Collector Injection
-
-Before running your hyperparameter tuning Katib Experiment with Python SDK, 
-ensure the namespace label `katib.kubeflow.org/metrics-collector-injection: enabled` 
-is present. This label enables the sidecar container injection to collect metrics during the experiment.
-
-You can configure the namespace using the following YAML:
-
-```yaml
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: <your-namespace>
-  labels:
-    katib.kubeflow.org/metrics-collector-injection: enabled
-```
-
-Or you can add the label to an existing namespace using the following command:
-
-```bash
-kubectl label namespace <your-namespace> katib.kubeflow.org/metrics-collector-injection=enabled
-```
-
-### A Simple Example
-
 You can run your first hyperparameter tuning Katib Experiment using Python SDK.
 
 In the following example we are going to maximize a simple objective function:
@@ -63,6 +38,11 @@ If you install Katib as part of Kubeflow Platform, you can open a new
 
 **Note**. If you use Katib within Kubeflow Platform to run this example, you need to use this
 namespace: `KatibClient(namespace="kubeflow-user-example-com")`.
+
+**Note**. The `kubeflow` namespace is pre-configured with the required label 
+`katib.kubeflow.org/metrics-collector-injection: enabled` for metrics collection. If you want to use pull-based metrics collector 
+in other namespaces, you can attach this label following the instructions in the 
+[Metrics Collector](/docs/components/katib/user-guides/metrics-collector/#prerequisites).
 
 ```python
 # [1] Create an objective function.
--- a/content/en/docs/components/katib/getting-started.md
+++ b/content/en/docs/components/katib/getting-started.md
@@ -17,6 +17,31 @@ You need to install the following Katib components to run examples:
 
 ## Getting Started with Katib Python SDK
 
+### Adding Namespace Label for the Metrics Collector Injection
+
+Before running your hyperparameter tuning Katib Experiment with Python SDK, 
+ensure the namespace label `katib.kubeflow.org/metrics-collector-injection: enabled` 
+is present. This label enables the sidecar container injection to collect metrics during the experiment.
+
+You can configure the namespace using the following YAML:
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: <your-namespace>
+  labels:
+    katib.kubeflow.org/metrics-collector-injection: enabled
+```
+
+Or you can add the label to an existing namespace using the following command:
+
+```bash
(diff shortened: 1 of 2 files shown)


content/en/docs/components/katib/getting-started.md
```diff
@@ -17,6 +17,31 @@ You need to install the following Katib components to run examples:
 
 ## Getting Started with Katib Python SDK
 
+### Adding Namespace Label for the Metrics Collector Injection
```
andreyvelich, line 20: I would suggest that we add this section to the Katib Metrics Collector guide for now: https://www.kubeflow.org/docs/components/katib/user-guides/metrics-collector/
The getting started example, tell users to run Experiment in `kubeflow` namespace, which has the required label in default standalone installation.

andreyvelich (commented):
Thank you for this @Doris-xm!
/assign @hbelmiro @varodrig @helenxie-bit @Electronic-Waste

content/en/docs/components/katib/getting-started.md
helenxie-bit: I noticed something strange. When I previously ran the get started example with the `kubeflow` namespace, I [encountered the "namespace lacks label" error](https://github.com/kubeflow/katib/issues/2492). However, I just tried it again, and now it works fine with the `kubeflow` namespace.

On the other hand, when I use the API to run LLM hyperparameter optimization using `kubeflow` namespace too, the "namespace lacks label" error still occurs.

@andreyvelich Do you have any idea what might be causing this issue? I’ll continue investigating as well.

andreyvelich: How did you install Katib and Training Operator when you run LLM hyperparameter optimization using `kubeflow` namespace?

helenxie-bit: I used the following commands from the installation page to install the latest changes of Katib control plane and Training Operator control plane:
`kubectl apply -k "github.com/kubeflow/katib.git/manifests/v1beta1/installs/katib-standalone?ref=master"`

`kubectl apply --server-side -k "github.com/kubeflow/training-operator.git/manifests/overlays/standalone?ref=master"`

andreyvelich: So, when you deploy Training Operator after the Katib, is might override the namespace label, since it doesn't have it: https://github.com/kubeflow/training-operator/blob/master/manifests/overlays/standalone/namespace.yaml
You should install Training Operator before Katib ideally.
Also, to use PyTorchJob CRD within Katib Trial, it needs to be exist before Katib Controller starts: https://github.com/kubeflow/katib/blob/master/pkg/controller.v1beta1/trial/trial_controller.go#L121-L132

helenxie-bit: Oh I see! Problem fixed! Thank you so much

Electronic-Waste: > I would suggest that we add this section to the Katib Metrics Collector guide for now: https://www.kubeflow.org/docs/components/katib/user-guides/metrics-collector/
The getting started example, tell users to run Experiment in kubeflow namespace, which has the required label in default standalone installation.

SGTM.

content/en/docs/components/katib/getting-started.md
```diff
@@ -17,6 +17,31 @@ You need to install the following Katib components to run examples:
...
+
+Before running your hyperparameter tuning Katib Experiment with Python SDK, 
+ensure the namespace label `katib.kubeflow.org/metrics-collector-injection: enabled` 
+is present. This label enables the sidecar container injection to collect metrics during the experiment.
+
+You can configure the namespace using the following YAML:
```
varodrig, line 26: Since the user needs to update the namespace, no create a new one. I'd update the sentence by
"You can configure the namespace by adding the following label  katib.kubeflow.org/metrics-collector-injection: enabled as is shown in the sample code."

content/en/docs/components/katib/getting-started.md
Electronic-Waste: > @andreyvelich @Electronic-Waste should we add a pre-req section on https://www.kubeflow.org/docs/components/katib/user-guides/metrics-collector/ to include this step?

SGTM. We could also declare that `kubeflow` namespace has already been configured with this label, that's why they can use `tune` function in `kubeflow` namespace without any prerequisite setup.

Such as "If you want to use pull-based metrics collector, you need to attach the label `xxx` to the target namespace (`kubeflow` namespace has built-in support for this label)"

Doris-xm: Thanks for your review. I've updated the doc:
- move the detailed guidance to Metrics Collector
- add a **Note** in Getting Started to explain `kubeflow` namespace has built-in support. I also include a link to Metrics Collector for detailed instruments.

Please let me know if you have any suggestions.

content/en/docs/components/katib/user-guides/metrics-collector.md
```diff
@@ -6,6 +6,30 @@ weight = 40
 
 This guide describes how Katib metrics collector works.
 
+## Prerequisites
+
+Before running your hyperparameter tuning Katib Experiment with Python SDK,
+ensure the namespace label `katib.kubeflow.org/metrics-collector-injection: enabled`
+is present. This label enables the sidecar container injection to collect metrics during the experiment.
```
Electronic-Waste, line 13:
```suggestion
is present. This label enables the sidecar container injection for pull-based metrics collectors to collect metrics during the experiment.
```

Electronic-Waste (commented):
Basically LGTM! Just one comment for you @Doris-xm!

Electronic-Waste (approved):
LGTM! Thanks for your great contribution @Doris-xm!

/lgtm
/assign @kubeflow/wg-automl-leads @helenxie-bit

helenxie-bit:
Thanks for the contribution!

/lgtm

andreyvelich (commented):
Thank you for this contribution @Doris-xm!
/lgtm
/approve

Issue #3707, opened 2024-04-02
So there's no concepts? Your concepts page is empty.


warmbasket:
So there's no concepts? Your concepts page is empty.

akrupien:
https://www.kubeflow.org/docs/concepts/ links only to this https://www.kubeflow.org/docs/concepts/multi-tenancy/

sftim:
This is an open source project. Help is welcome.

warmbasket:
@sftim There's no concepts, close the page in the meantime then? 🤯

varodrig:
we need help drafting a proposal of what information should be displayed in this web page.

also looking for comments from

@andreyvelich @terrytangyuan @thesuperzapper

varodrig:
cc @kubeflow/kubeflow-steering-committee @kubeflow/wg-training-leads @kubeflow/wg-pipeline-leads @kubeflow/wg-notebooks-leads @kubeflow/wg-manifests-leads @kubeflow/release-team @kubeflow/wg-data-leads @kubeflow/wg-deployment-leads

varodrig:
> I vote to remove and comeback when we have something concrete we want to share.

Currently, we have the https://www.kubeflow.org/docs/concepts/multi-tenancy/ which is out of date.

varodrig:
I propose to remove this page as well, Not sure what was planned to include here when it was created.
thoughts?
[@kubeflow/kubeflow-steering-committee](https://github.com/orgs/kubeflow/teams/kubeflow-steering-committee) [@kubeflow/wg-training-leads](https://github.com/orgs/kubeflow/teams/wg-training-leads) [@kubeflow/wg-pipeline-leads](https://github.com/orgs/kubeflow/teams/wg-pipeline-leads) [@kubeflow/wg-notebooks-leads](https://github.com/orgs/kubeflow/teams/wg-notebooks-leads) [@kubeflow/wg-manifests-leads](https://github.com/orgs/kubeflow/teams/wg-manifests-leads) [@kubeflow/release-team](https://github.com/orgs/kubeflow/teams/release-team) [@kubeflow/wg-data-leads](https://github.com/orgs/kubeflow/teams/wg-data-leads) [@kubeflow/wg-deployment-leads](https://github.com/orgs/kubeflow/teams/wg-deployment-leads)

franciscojavierarceo:
Yeah, agreed.

thesuperzapper:
I agree to remove, it's not correct anymore. 

[...]

PR #3977, opened 2025-01-27
Remove "Concepts" / "Multi-Tenancy" section

thesuperzapper:
closes https://github.com/kubeflow/website/issues/3707

As discussed, having the "Concepts" sidebar section with only "Multi-Tenancy" is not helpful to users, especially when the "Multi-Tenancy" pages are out of date and not correct.

This PR removes the `Concepts / Multi-Tenancy` pages and redirects them to [`Components / Central Dashboard / Profiles and Namespaces`](https://www.kubeflow.org/docs/components/central-dash/profiles/), as this is the most up-to-date information about how multi-tenancy works.


andreyvelich (commented):
I agree with this change. 
/lgtm
/assign @kubeflow/wg-manifests-leads @kubeflow/wg-notebooks-leads 

@thesuperzapper I even think that we should move Kubeflow projects outside of components section to give them more visibility.
So the left panel can be:
```
About
Getting Started
Kubeflow Notebooks
Kubeflow Model Registry
Kubeflow Spark Operator
Kubeflow Trainer
Kubeflow Katib
Kubeflow Pipelines
Kubeflow Dashboard
Distributions
Releases
```

We can order components by ML lifecycle (similar to this diagram: https://www.kubeflow.org/docs/started/architecture/#introducing-the-ml-lifecycle) or alphabetically. Personally, I prefer ML Lifecycle order.

Any thoughts @kubeflow/wg-pipeline-leads @kubeflow/wg-data-leads @kubeflow/kubeflow-steering-committee @StefanoFioravanzo @juliusvonkohout @franciscojavierarceo @kubeflow/wg-training-leads @kubeflow/wg-notebooks-leads ?

franciscojavierarceo:
Yeah I agree with that

thesuperzapper:
@andreyvelich let's discuss any other changes in a separate issue.

But to respond here, while I agree that we should reorder the components sidebar to ML lifecycle order, I strongly disagree with moving the components to the top level. 

It will just clutter the website and make it harder to browse, especially as we add more components.

There are also technical reasons why it would be ill-advised, mainly around the fact that the HTTP paths come from the folders, and historical redirects.

juliusvonkohout:
Yeah the multi-tenancy is more for platform administrators. Right now some components do not support hard multi-tenancy such as minio and ml-metadata for KFP. Maybe this is something we should keep somewhere or i just add it in the kubeflow/manifests readme for advanced users.

thesuperzapper:
@juliusvonkohout the reason we are proposing to remove it (and why others seem to be agreeing) is that the information is really out of date, and not correct (especially after the changes to how we use Istio in 1.9.1).

As stated, I propose we remove the sidebar section and redirect it to the "profiles" guide, as this the most practical info for end users relating to multi-user.

varodrig (approved):

andreyvelich:
@franciscojavierarceo @juliusvonkohout @kubeflow/kubeflow-steering-committee If you are ok with this change, can you /lgtm please ?

andreyvelich (commented):
Thank you @thesuperzapper!
/approve

PR #3980, opened 2025-01-31
Add Swagger UI for Model Registry

thesuperzapper:
This PR adds a __Swagger UI Page__ page for the _Kubeflow Model Registry_ REST API, similar to what we [already have](https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/) for Kubeflow Pipelines.

It is automatically generated from [this swagger yaml file](https://github.com/kubeflow/model-registry/blob/v0.2.12/api/openapi/model-registry.yaml) and will automatically point to the github tag that corresponds to the newly added `model-registry/latest-version` shortcode, which can be used to update the version.

It also updates the other pages to use the newly introduced `model-registry/latest-version` shortcode, which enables easier updating of the docs when new versions are released.

## Screenshot

REST-API-v1alpha3-Kubeflow-01-31-2025_11_02_AM


thesuperzapper:
@tarilabs @dhirajsb @ederign this is a cool addition to the model registry docs, and I think it should be easy to review.

We are adding a Swagger REST API doc based on your generated YAML, see the preview here:

- https://deploy-preview-3980--competent-brattain-de2d6d.netlify.app/docs/components/model-registry/reference/rest-api/

tarilabs (commented):
thank you @thesuperzapper ! 👏 🙏 

I confirm I see reformattings, use of version refactor in different shortcode, and the Swagger as you noted.

But for this last one I was reluctant to copy it over because, as you also notice there is no CORS support by the SwaggerUI:

> The try it out feature of Swagger UI does not work due to authentication and CORS, but it can help you construct the correct API calls.

is this the approach taken by other KF components, or wouldn't have been sufficient to link to the editor as we do in the contributing guide?

thesuperzapper:
> But for this last one I was reluctant to copy it over because, as you also notice there is no CORS support by the SwaggerUI:
> 
> > The try it out feature of Swagger UI does not work due to authentication and CORS, but it can help you construct the correct API calls.

@tarilabs all that this is trying to say is that you can't initiate requests to a kubeflow platform that is hosted on `kubeflow.example.com`, because chrome will detect that as a cross-origin scripting attack.

And even if that wasn't a problem, getting an authentication token would be very complicated for users, because there's quite a lot of variation in deployments of kubeflow.

> is this the approach taken by other KF components, or wouldn't have been sufficient to link to the editor as we do in the contributing guide?

Yes, It's [what we do](https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/) in kubeflow pipelines.

Having this page on the website is mainly to document the API, not to have people actually use the API from this page. 

It's important that we document every REST API on the official website. However, feel free to have any other links you want on this page, If they help the user.

Although probably the only helpful link would be to a Swagger UI that happens to be hosted on the app itself, so you can avoid the CORS/auth issues.

tarilabs:
yes we adopted the same strategy for the contributing guide (wrt limitation of the Swagger UI).
if this is done also for other KF Components more than happy to align, and thank you again for this PR!!

adysenrothman (approved):
this looks great!

dhirajsb:
@thesuperzapper looks good as an API document. I see the note that the API can't be directly invoked so that limits it somewhat, but it's still a good way to see what the payloads look like. Thanks!

tarilabs (approved):
- https://github.com/kubeflow/website/pull/3980#pullrequestreview-2590369540
- https://github.com/kubeflow/website/pull/3980#pullrequestreview-2609579198
- https://github.com/kubeflow/website/pull/3980#issuecomment-2656753971

/lgtm
/approve

thesuperzapper:
@andreyvelich @juliusvonkohout we need a root approver as this changes style templates.

But it has been approved by the model registry team so it's good to go.

andreyvelich:
Thank you @thesuperzapper 
/approve

PR #3984, opened 2025-02-04
Rename #kubeflow-gsoc-participants to #kubeflow-contributors

terrytangyuan:


andreyvelich (commented):
Thank you @terrytangyuan!
Please can you also rename it here: https://www.kubeflow.org/events/gsoc-2025/#steps

franciscojavierarceo (approved):

terrytangyuan:
We decided to rename it after GSOC last year.

andreyvelich:
>wont this be confusing since most KF members are not in the GSOC channel, and the GSOC people will need their own channel?

We discussed that it is ok for maintainers and GSoC student to use the same channel: `#kubeflow-contributors`, since all of these folks are Kubeflow Contributors.

thesuperzapper:
@andreyvelich what about questions that are specific to GSoC, which there are often __many__ of during the application period.

Won't that make the channel not very useful for actual contributor discussions?

andreyvelich:
> @andreyvelich what about questions that are specific to GSoC, which there are often **many** of during the application period.
> 
> Won't that make the channel not very useful for actual contributor discussions?

@thesuperzapper I don't see any problems to discuss these questions in #kubeflow-contributors channel.
Since these questions are also relevant for all new contributors who just getting started with Kubeflow ecosystem, and want to understand how to contribute. 

So folks can search in the channel for similar topics they want to discuss.

PR #3993, opened 2025-02-12
fix(trainer): Fix redirect for /components/training/

andreyvelich:
Fixes: https://github.com/kubeflow/trainer/issues/2433

This should fix this redirect: https://www.kubeflow.org/docs/components/training

/assign @kubeflow/wg-training-leads @gaocegege


andreyvelich:
cc @Electronic-Waste

Electronic-Waste (approved):
Thanks for this!

/lgtm

Issue #3934, opened 2024-11-30
Create a stale script to mark stale issues and pull requests


varodrig:
Since we currently have 79 issues open (the oldest is from 2020).
https://github.com/kubeflow/website/pulls

We need a way to standarize PRs and issues to be stale after x amount of time. This was mentioned during the community meeting last week.

The recommended implementation is to follow a similar implementation as the pipeline module does:

- [](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/stale.yml)

varodrig:
@andreyvelich @mageirakos

hbelmiro:
Sounds good to me.

andreyvelich:
Sounds good, thanks @varodrig.
Please also add message to stale PRs, similar to our Training Operator stale config: https://github.com/kubeflow/training-operator/blob/master/.github/workflows/stale.yaml#L34-L43

juliusvonkohout:
You can just copy https://github.com/kubeflow/manifests/blob/master/.github/workflows/stale.yaml for /lgtm https://github.com/kubeflow/website/pull/3995

PR #3995, opened 2025-02-13
Create stale.yaml

juliusvonkohout:
copied from kubeflow/manifests

@varodrig 
fixes https://github.com/kubeflow/website/issues/3934


juliusvonkohout:
/approve

@varodrig can you /lgtm ?

varodrig:
@juliusvonkohout can you add [start-date](https://github.com/actions/stale#start-date) as starting this year? January 1, 2025.
since there are some old PRs/issues that I'm keeping an eye on it and we need them.

Thanks.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-stale: 60
```
varodrig, line 23: @andreyvelich  suggested to use this as example https://github.com/kubeflow/trainer/blob/master/.github/workflows/github-stale.yaml

can we update days-before-stale to 90.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,48 @@
...
+        # The message that will be added as a comment to the pull requests
+        # when the stale workflow closes it automatically after being stale for too long.
+        close-pr-message: "This pull request has been automatically closed because it has not had recent  activity.You can reopen the PR if you want. \n"
+        stale-pr-label: lifecycle/stale
+        # Exclude them from being marked as stale
+        exempt-pr-labels: lifecycle/frozen,enhancement,good first issue
+        # The issues or the pull requests with a milestone will not be marked as stale automatically
+        exempt-all-milestones: true
+        # Learn more about operations: https://github.com/actions/stale#operations-per-run.
+        operations-per-run: 250
```
varodrig, line 48: @juliusvonkohout  can we include a start date for this? there are issues and PRs that I'm keeping an eye on it and we need them.
It will be something like this start-date: 2025-01-01

based on:
https://github.com/actions/stale?tab=readme-ov-file#start-date
The ISO 8601 date format is "YYYY-MM-DD"



strat

varodrig (commented):
thank you so much @juliusvonkohout  for working on this.

.github/workflows/stale.yaml
juliusvonkohout: I am against the start date, because after 90 days it shall also kick in for current issues. I also doubt that this is what you want.

juliusvonkohout: you can use /lifecycle frozen as in https://github.com/kubeflow/manifests/issues?q=is%3Aissue%20state%3Aopen%20label%3Alifecycle%2Ffrozen if you really want to override it long-term

varodrig: after reviewing this with @juliusvonkohout  we are moving forward with not including a start date since we still need these issues and PRs to be stale or close if we do not take action.

Issue #3982, opened 2025-02-01
Proposal: Move Kubeflow projects outside of components section at the website.


andreyvelich:
Ref discussion: https://github.com/kubeflow/website/pull/3977#pullrequestreview-2584862336

I propose moving Kubeflow projects out of the `components` section to enhance their visibility in the website outline. 
From my point of view, adding another level to the outline offers no clear benefit, even if more projects will join Kubeflow ecosystem in the future. Moreover, since KFP and Kubeflow Trainer already defines a `Component` concept, this change will help reduce user confusion.

```
About
Getting Started
Kubeflow Notebooks
Kubeflow Model Registry
Kubeflow Spark Operator
Kubeflow Trainer
Kubeflow Katib
Kubeflow KServe
Kubeflow Pipelines
Kubeflow Dashboard
Distributions
Releases / Kubeflow Platform Releases
``` 

Looking at other projects, every sub-project has clear representation at the website outline:
- Argo: https://argoproj.github.io/
- Ray: https://docs.ray.io/en/latest/index.html
- Knative: https://knative.dev/docs/
- HuggingFace: https://huggingface.co/docs/transformers/index

Please let us know what do you think ?
@kubeflow/wg-pipeline-leads @kubeflow/wg-data-leads @kubeflow/kubeflow-steering-committee @StefanoFioravanzo @juliusvonkohout @franciscojavierarceo @kubeflow/wg-training-leads @kubeflow/wg-notebooks-leads @varodrig @hbelmiro @jbottum @kubeflow/wg-manifests-leads @astefanutti @saileshd1402

andreyvelich:
Regarding @thesuperzapper comment: https://github.com/kubeflow/website/pull/3977#issuecomment-2625794720

>It will just clutter the website and make it harder to browse, especially as we add more components.

Why it is clutter the website ? Even today if user wants to check docs for any project, they will see all components at the left side:

![Image](https://github.com/user-attachments/assets/56d5b72a-c51b-4527-8477-537b303dd3ad)


---

>There are also technical reasons why it would be ill-advised, mainly around the fact that the HTTP paths come from the folders, and historical redirects.

Can you explain more please ? Every project will still have dedicated folder under `/docs` path.

thesuperzapper:
@andreyvelich can you clarify what your goal with this change is?

If your issue is just highlighting the components more, we can probably just change the way we display the components section on the sidebar (visually, without moving the files).

[...]

PR #3981, opened 2025-01-31
website: Add dark theme

thesuperzapper:
This PR adds a new dark-theme to the Kubeflow website.

resolves https://github.com/kubeflow/website/issues/3692
resolves https://github.com/kubeflow/website/issues/3954

# How does this work?

Because we use docsy, which has its own themes in addition to bootstrap, we now render two themes into our CSS based on the `[data-theme]` attribute on the root `<html>` element, see `assets/scss/main.scss` for more context.

This approach allows us to retain our custom docsy changes while adding a nice dark theme.

# What changed?

__NOTE:__ this PR does NOT change any of the docs content. However, some images/diagrams have been updated to look better when viewed on a dark background.

## Theme-specific styling

To make theme-specific styling possible, we have added the following:

- __CSS classes:__
    - `show_light-only` (only show element in light mode)
    - `show_dark-only` (only show element in dark mode)
- __Theme-specific SCSS variables:__
    - `assets/scss/dark/_variables.scss`
    - `assets/scss/light/_variables.scss`
- __Theme-specific SCSS styles:__
    - `assets/scss/dark/_styles.scss`
    - `assets/scss/light/_styles.scss`

__NOTE:__ the existing `assets/scss/_styles_project.scss` is still used for global styles. It is used by both themes, usually with theme-specific variables to set colors.

## Theme Switcher

We have added a theme switcher to the top right of the website, which allows users to switch between the light and dark themes. By default, we match the theme set on the user's OS.

theme_switcher

## Images

We have not created new images for the dark theme, as most looked OK on the dark background. In the future, we can use the `show_light-only` and `show_dark-only` classes to show different images based on the theme.

We have updated the following images to look better on the dark theme:

- Getting Started / Introduction
- ~Getting Started / Architecture~ (split off into https://github.com/kubeflow/website/pull/4038 and https://github.com/kubeflow/website/pull/4068)

## Iframes / Swagger

There were two cases where I used a color inversion filter to create a fake "dark mode":

1. For the Google calendar embed in the community page
2. For the Swagger UI API reference pages

They seem to work well enough, but we can revisit this in the future if those components get official dark mode support.

# Screenshots

## Home Page ([preview](https://deploy-preview-3981--competent-brattain-de2d6d.netlify.app/))

Home

## Getting Started ([preview](https://deploy-preview-3981--competent-brattain-de2d6d.netlify.app/docs/started/))

Getting Started

[...]

PR #4023, opened 2025-02-27
[WIP] website: Move Kubeflow Projects outside of Components

andreyvelich:
Fixes: https://github.com/kubeflow/website/issues/3982

As we discussed with KSC members, we propose to move Kubeflow Projects outside of Components section at the website to make it more visible for users.
Also, we would like to remove External Add-Ons section from the website, and potentially rename it to:
```
Integrations with Kubeflow Ecosystem
```

Our proposed outline is this:
```
About
Getting Started
Kubeflow Notebooks
Kubeflow Spark Operator
Kubeflow Trainer
Kubeflow Katib
Kubeflow KServe
Kubeflow Model Registry
Kubeflow Pipelines
Kubeflow Dashboard
Kubeflow Distributions 
Kubeflow Platform Releases
```

TODO:

- [ ] Fix redirects.
- [ ] Fix links in the pages.

/hold for review

/assign @kubeflow/kubeflow-steering-committee @varodrig @rimolive @hbelmiro @thesuperzapper


franciscojavierarceo:
Hopefully we can decide on projects joining Kubeflow and Feast before merging this 😄

thesuperzapper:
@andreyvelich we please do this virtually in the sidebar template and page metadata rather than by actually moving the pages, as it will make it nearly impossible to make the old http links work (we have used many of the root paths).

This will really mess with the website otherwise.

andreyvelich:
> @andreyvelich can we please do this virtually in the sidebar template and page metadata rather than by actually moving the pages, as it will make it nearly impossible to make the old http links work (we have used many of the root paths).
> 
> This will really mess with the website otherwise.

Why it is impossible ?
We just have ~ 270 pages under `component` where we should add re-directs to the new location.

thesuperzapper:
@andreyvelich the problem is that we have used most of the root paths like "pipelines/" in past, so the old redirect chains will not work, and will loop.

Also, the components should get special styling in their own part of the sidebar, and if we are doing that, we might as well just update the sidebar template to just unpack the "components" folder to the root under its own heading.

This also has the side effect of being much less disruptive to historical links and redirects.

Also because that change is much less disruptive, we can do it immediately (assuming we leave the external add-ons for a future PR).

Do you want to try doing it, or do you need some pointers?

thesuperzapper:
We can pull the "components" under their own top-level underline heading on the sidebar (similar to the "documentation" underline heading at the top of the sidebar right now)

So there would be two "sections":

- __Documentation:__
    - "about", "getting started", "integrations", "distributions", "releases" 
- __Components:__
    - "pipelines", "notebooks", etc.

andreyvelich:
>the problem is that we have used most of the root paths like "pipelines/" in past, so the old redirect chains will not work, and will loop.

Why it will loop ?
If we add links at the bottom of `_redirect` file, it will just redirect to the new page e.g.
```
/docs/components/pipelines/legacy-v1/sdk/             /docs/kubeflow-pipelines/legacy-v1/sdk/ 
```

>Also, the components should get special styling in their own part of the sidebar, and if we are doing that, we might as well just update the sidebar template to just unpack the "components" folder to the root under its own heading.
This also has the side effect of being much less disruptive to historical links and redirects.
Also because that change is much less disruptive, we can do it immediately (assuming we leave the external add-ons for a future PR).
Do you want to try doing it, or do you need some pointers?

We discussed with KSC members that we want to:
Rename "Kubeflow Components" to "Kubeflow Projects" and remove concept of "External Add-Ons".
Given that Kubeflow Pipelines also has a concept of "Component" that will help with user confusion.


>Documentation:
"about", "getting started", "integrations", "distributions", "releases"
Components:
"pipelines", "notebooks", etc.

Can we rename:
```
Documentation -> Kubeflow Overview 
Components -> Kubeflow Projects
```

cc @varodrig @kubeflow/kubeflow-steering-committee @rimolive for feedback

andreyvelich:
@thesuperzapper Any thoughts how we can move this forward ? I would love to finalize it before the KubeCon in April, so new folks will see the updated version of our website.
Shell we rename Components -> Projects visually in the UI and keep this list always open, like in Istio docs for Concepts/Tasks:
https://istio.io/latest/docs/

thesuperzapper:
@andreyvelich let's do it after we merge the dark mode PR to avoid conflicts in the styles/template.

Having the dark mode for kubecon will be very cool, we even beat kubernetes themselves.

We should do the sidebar styling in a new PR, as we won't need to move all the files like this PR tries to.

andreyvelich:
@thesuperzapper I am ok with that, but can you address the remaining comments on the PR: https://github.com/kubeflow/website/pull/3981
Also, we are waiting for the feedback from @ederign on the UI/UX point of view of this: https://github.com/kubeflow/website/pull/3981#discussion_r1990016510

andreyvelich:
@kubeflow/kubeflow-steering-committee @kubeflow/kubeflow-outreach-committee @thesuperzapper I updated this website PR to only keep the Kubeflow Projects sidebar always open.
Also, I keep the logical order of the Kubeflow Projects to follow AI Lifecycle diagram: https://www.kubeflow.org/docs/started/architecture/#kubeflow-projects-in-the-ai-lifecycle
If that looks good, let's merge it!

franciscojavierarceo (approved):
/lgtm
/approve

andreyvelich (commented):
/hold cancel

PR #4084, opened 2025-04-11
docs(pipelines): add troubleshooting tip for installation 

kaikaila:
This PR adds a troubleshooting tip under the installation section of Kubeflow Pipelines standalone deployment guide.

Some users (see kubeflow/pipelines#11757 and kubeflow/pipelines#9546) reported that using the default `env/dev` path caused persistent pod crashes, such as `proxy-agent` and `workflow-controller` failing to start. As a workaround, switching to the `env/platform-agnostic` configuration helped resolve the issue in version `2.0.0`, verified on Minikube.

This patch proposes a minimal documentation update to help users facing similar issues.

### DIFF
--- a/content/en/docs/components/pipelines/operator-guides/installation/_index.md
+++ b/content/en/docs/components/pipelines/operator-guides/installation/_index.md
@@ -24,11 +24,20 @@ You should be familiar with [Kubernetes](https://kubernetes.io/docs/home/),
      kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/env/dev?ref=$PIPELINE_VERSION"
      ```
 
+> 💡 **Troubleshooting**: If you encounter persistent pod crashes (e.g., `proxy-agent`, `workflow-controller`) after applying the default config, you may try using the `platform-agnostic` configuration instead:
+>
+> ```bash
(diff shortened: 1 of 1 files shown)


content/en/docs/components/pipelines/operator-guides/installation/_index.md
```diff
@@ -24,11 +24,19 @@ You should be familiar with [Kubernetes](https://kubernetes.io/docs/home/),
      kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/env/dev?ref=$PIPELINE_VERSION"
```
milinddethe15, line 24: I think we replace the environment installation from `dev` to `env/platform-agnostic` to make the installation compatible on any Kubernetes cluster

kaikaila: > I think we replace the environment installation from `dev` to `env/platform-agnostic` to make the installation compatible on any Kubernetes cluster

That makes sense — I originally added it as a troubleshooting tip to keep the change minimal, but I agree `platform-agnostic` might be a better default in the long run.

I'm happy to update the PR if that's the preferred direction from the maintainers. Open to hearing what others think!

Arhell (commented):
/ok-to-test

juliusvonkohout:
/lgtm
/approve

CC @hbelmiro @HumairAK if something is not right.

PR #4083, opened 2025-04-10
website: added social media icon on readme and footer

jaiakash:
Fixes #3684
Fixes #3633
Fixes #3683

Hi, I have added
- X (twitter), Slack, YouTube and KubeFlow CNCF icon in project readme
- New twitter icon, YouTube and LinkedIn on footer

## Screenshot 1:
Screenshot 2025-02-13 at 11 43 29 PM
-------------------------------------
## Screenshot 2:
image

PR #4085, opened 2025-04-12
website: properly update fontawsome + partial revert #4083

thesuperzapper:
/area website

This PR fixes the error introduced by https://github.com/kubeflow/website/pull/4083 where fontawsome is loaded multiple times on every page, and has two conflicting versions. This was a serious issue because it affects page load times, but also because the `integrity` arg was not set on the `<link>` which was added to the footer.

It also partially reverts the readme update which put an image above the title of the root readme. This hurts SEO of the repo, because the title needs to be at the top. But it also implies that this is the "kubeflow" repo (when its just the website source). Note, we already have a logo on the [`kubeflow/kubeflow` readme](https://github.com/kubeflow/kubeflow), and it sits below the title.

__WARNING:__ after this PR, everyone will need to stop doing `(cd themes/docsy/ && npm install)` when trying to run a local version of the website. Instead, they must run `npm install` at the root of the repo, as this is now where fontawsome and bootstrap are installed, rather than under the `themes/docsy` folder.

__Screenshot [from preview site](https://deploy-preview-4085--competent-brattain-de2d6d.netlify.app/) showing that the new icons still work after this PR:__

Screenshot 2025-04-12 at 15 18 40

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,3 @@
-<div align="center">
-  <img src="./assets/icons/logo.svg" width="320">
-</div>
-
 # Kubeflow Website
 
 <h4 align="center">
@@ -105,8 +101,9 @@ Follow the usual GitHub workflow of forking the repository on GitHub and then cl
 4. Install Docsy dependencies:
 
    ```bash
+   # install the node packages (at the root of the repo)
    # NOTE: ensure you have node 18 installed
-   (cd themes/docsy/ && npm install)
+   npm install
    ```
 
 5. Start your local Hugo server:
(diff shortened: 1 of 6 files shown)


thesuperzapper:
@andreyvelich as you approved https://github.com/kubeflow/website/pull/4083, can you please quickly review this one so we can fix the issues it introduced.

cc @milinddethe15 @jaiakash

andreyvelich (commented):
Thanks for updating this @thesuperzapper
/lgtm
/approve

PR #4167, opened 2025-08-04
trainer: Update Runtime Guide and Deprecation Policy

andreyvelich:
As we discussed, I added the Runtime Deprecation Policy to the runtimes.

Also, I updated the Runtime guides to make it more clear and structured.

Looking forward for your feedback!

/assign @kubeflow/kubeflow-trainer-team @astefanutti @kramaranya @franciscojavierarceo @varodrig @davidgs @rudeigerc @kubeflow/wg-training-leads

### DIFF
--- a/content/en/docs/components/trainer/operator-guides/runtime.md
+++ b/content/en/docs/components/trainer/operator-guides/runtime.md
@@ -46,7 +46,7 @@ spec:
                       image: pytorch/pytorch:2.7.1-cuda12.8-cudnn9-runtime
 ```
 
-In the Runtime specification, platform administrators define a reusable template with the
+In the runtime specification, platform administrators define a reusable template with the
 appropriate configuration for distributed training. A TrainJob references this runtime
 via the `RuntimeRef` API, which links to its `APIGroup`, `Kind` and `Name`. This allows the TrainJob
 to adopt the runtime’s configuration, enabling consistent and modular training setups.
@@ -120,7 +120,7 @@ spec:
 
 ## Framework Label Requirement
 
-Every deployed Runtime must have the label `trainer.kubeflow.org/framework` to ensure that
+Every deployed runtime must have the label `trainer.kubeflow.org/framework` to ensure that
 the Kubeflow SDK can recognize it, for example:
 
 ```yaml
@@ -156,24 +156,29 @@ The following runtimes are supported for TorchTune BuiltinTrainer:
 
 ### Runtime Deprecation Policy
 
-Since ML frameworks naturally evolve over time, Kubeflow community may decide to deprecate
+As ML frameworks evolve over time, the Kubeflow community may decide to deprecate
 certain supported ClusterTrainingRuntimes. To avoid breaking existing users, we follow a deprecation
-policy for Runtimes that are slated to be removed.
+policy before removing any runtime.
 
 A supported ClusterTrainingRuntime may be marked as deprecated and is eligible for removal starting
 from **two minor releases** after its deprecation.
 
-To ensure users are aware of Runtime deprecations, the following measures are taken:
+These measures are taken to inform users about runtime deprecation:
 
-- Kubeflow Trainer release notes include the deprecation as a breaking change.
-- The ClusterTrainingRuntime YAML is updated with a deprecation comment.
-- If a deprecated Runtime is deployed on Kubernetes cluster, Kubeflow Trainer validation webhook
-  prints a warning.
-- If a TrainJob references a deprecated Runtime, the validation webhook also prints a warning.
-- The Kubeflow SDK prints a warning when a deprecated Runtime is listed or referenced.
+- Add the following label to the deprecated runtime:
+
+  ```yaml
+  trainer.kubeflow.org/deprecated: "true"
+  ```
+
+- Document the deprecation as **a breaking change** in Kubeflow Trainer release notes.
+- Show a warning from the Kubeflow Trainer validation webhook when:
+  - A deprecated runtime is deployed on a Kubernetes cluster.
+  - A TrainJob is created which references a deprecated runtime.
+- Display a warning in the Kubeflow SDK when a deprecated runtime is listed or referenced.
 
 ## Next Steps
 
 - Learn how to configure [gang scheduling in Kubeflow Trainer](/docs/components/trainer/operator-guides/gang-scheduling).
-- Explore how to set up [MLPolicy in Runtimes](/docs/components/trainer/operator-guides/ml-policy).
-- See how to define [Job Template in Runtimes](/docs/components/trainer/operator-guides/job-template).
+- Explore how to set up [MLPolicy in runtime](/docs/components/trainer/operator-guides/ml-policy).
+- See how to define [Job Template in runtimes](/docs/components/trainer/operator-guides/job-template).


content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -99,157 +99,78 @@ spec:
...
+A supported ClusterTrainingRuntime may be marked as deprecated and is eligible for removal starting
+from **two minor releases** after its deprecation.
 
-**Required Labels:**
+To ensure users are aware of Runtime deprecations, the following measures are taken:
 
-- `trainer.kubeflow.org/trainjob-ancestor-step`: Specifies the role or step of the replicated job in the training workflow (e.g., `dataset-initializer`, `model-initializer` or `trainer`).
+- Kubeflow Trainer release notes include the deprecation as a breaking change.
+- The ClusterTrainingRuntime YAML is updated with a deprecation comment.
+- If a deprecated Runtime is created, Kubeflow Trainer validation webhook prints a warning.
```
Electronic-Waste: What's the meaning of creating deprecated Runtimes? Isn't it installed by default since it's officially supported by Kubeflow Trainer?

content/en/docs/components/trainer/operator-guides/gang-scheduling.md
```diff
@@ -0,0 +1,37 @@
...
+The `PodGroupPolicy` API supports multiple policies, known as `PodGroupPolicySources`. Each policy
+represents plugin configuration to enable gang scheduling using that specific integration. You can
+specify one of the supported policies in the `PodGroupPolicy` API to enable gang scheduling with
+supported plugins.
+
+### Coscheduling
+
+[The `Coscheduling` policy](https://pkg.go.dev/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1#CoschedulingPodGroupPolicySource)
+configures gang scheduling with
+[Coscheduling plugin](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master?tab=readme-ov-file#plugins)
```
astefanutti, line 31: Maybe it should be made more explicit that the co-scheduing plugin should be enabled?

content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -99,157 +99,78 @@ spec:
...
-              labels:
-                trainer.kubeflow.org/trainjob-ancestor-step: trainer
-            spec:
-              template:
-                spec:
-                  containers:
-                    - name: node
-                      image: ghcr.io/kubeflow/trainer/torchtune-trainer
-```
+- Learn how to configure [gang scheduling for TrainJobs](/docs/components/trainer/operator-guides/gang-scheduling).
```
Electronic-Waste:
```suggestion
- Learn how to configure [gang scheduling for Training Runtimes](/docs/components/trainer/operator-guides/gang-scheduling).
```

Maybe what you mean is `Training Runtime`

Electronic-Waste (commented):
@andreyvelich Thanks for this! I left my very initial comments for you

content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -99,157 +99,78 @@ spec:
...
+When referencing TrainingRuntime, the Kubernetes namespace must be the same as the TrainJob's namespace
+{{% /alert %}}
 
 ```YAML
-apiVersion: kubeflow.org/v2alpha1
+apiVersion: kubeflow.org/v1alpha1
 kind: TrainJob
 metadata:
   name: example-train-job
   namespace: team-a # Only accessible to the namespace for which it is defined
```
kramaranya:
```suggestion
apiVersion: trainer.kubeflow.org/v1alpha1
kind: TrainJob
metadata:
  name: example-train-job
  namespace: team-a # Only accessible to the namespace for which it is defined
```

content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -4,15 +4,17 @@ description = "How to manage Runtimes with Kubeflow Trainer"
 weight = 30
 +++
 
-## Overview
+This guide gives an overview for TrainingRuntime and ClusterTrainingRuntime.
 
-This guide explains how cluster administrators should manage `TrainingRuntime` and `ClusterTrainingRuntime`. It describes how to configure `MLPolicy`, `PodGroupPolicy`, and `Template` APIs.
+Runtimes are template configurations or blueprints that managed by platform administrators and
```
kramaranya:
```suggestion
Runtimes are template configurations or blueprints that are managed by platform administrators and
```

content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -99,157 +99,78 @@ spec:
...
-  number of processes per node (`numProcPerNode`) for PyTorch distributed workloads.
-- **MPI**: Configures distributed training using MPI. This policy allows you to specify options
-  such as the number of processes per node and MPI implementation details.
+## Supported Runtimes
 
-For a complete list of available options and detailed API fields, refer to the [Kubeflow Trainer API reference](https://pkg.go.dev/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1#MLPolicy).
+Kubeflow Trainer community maintains
+[several ClusterTrainingRuntime](https://github.com/kubeflow/trainer/tree/master/manifests/base/runtimes)
+to help AI practitioners quickly experiment with Kubeflow Trainer, and to enable
+platform administrators to extend these runtimes to fit their specific requirement.
```
kramaranya:
```suggestion
Kubeflow Trainer community maintains
[several ClusterTrainingRuntimes](https://github.com/kubeflow/trainer/tree/master/manifests/base/runtimes)
to help AI practitioners quickly experiment with Kubeflow Trainer, and to enable
platform administrators to extend these runtimes to fit their specific requirements.
```

content/en/docs/components/trainer/operator-guides/job-template.md
```diff
@@ -0,0 +1,145 @@
...
++++
+
+This guide describes how to configure
+[the `Template` API](https://pkg.go.dev/github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1#JobSetTemplateSpec)
+in the Kubeflow Trainer Runtimes.
+
+Before exploring this guide, make sure to follow [the Runtime guide](/docs/components/trainer/operator-guides/runtime)
+to understand the basics of Kubeflow Trainer Runtimes.
+
+### Template Overview
```
kramaranya:
```suggestion
## Template Overview
```

kramaranya (commented):
Thank you @andreyvelich!

content/en/docs/components/trainer/operator-guides/runtime.md
andreyvelich: I want to say that, if Runtime is marked as deprecated, the webhook prints a warning during creation.
Not sure if we have more elegant way to inform users. @tenzen-y @astefanutti thoughts ?

content/en/docs/components/trainer/operator-guides/runtime.md
andreyvelich: Actually, gang scheduling should be for TrainJob, not Runtimes.
In the future, we can use this doc to explain how to leverage PodSpecOverride to define scheduling policies per TrainJob.

Maybe in general I should: how to configure gang scheduling in Kubeflow Trainer.

content/en/docs/components/trainer/operator-guides/runtime.md
astefanutti: I think using webhook warnings is a great idea. Maybe we could add a link to https://kubernetes.io/blog/2020/09/03/warnings/#admission-webhooks to give more details on what they are and why it's a good solution.

tenzen-y: how to identify the deprecated runtimes during admission webhooks? resource name?
I think we should not restrict the runtime usage even after we remove the deprecated runtimes in the trainer repository. We should allow users to manually manage by themselves,

astefanutti: If I understand it correctly, we'll define an annotation to identify deprecated runtimes.

tenzen-y: But, the TrainingRuntime resource is immutable object. So, we can not enforce adding deprecation annotation to the existing TrainingRuntime in their cluster.

astefanutti: I imagine we can relax the constraint to allow some level of mutation, like adding annotations?

andreyvelich: >how to identify the deprecated runtimes during admission webhooks? resource name?

Yes, this is what I initially thought of (e.g. have the list of deprecated names), but do you think that maintaining label/annotation is better ?
```
trainer.kubeflow.org/deprecated: true
```

>But, the TrainingRuntime resource is immutable object

TrainingRuntime is immutable only when it is in-used by TrainJob (e.g. like how we add finalizer to the Runtime when TrainJob is running).
We should allow users to update the TrainingRuntime images when they update the KF Trainer control plane.

content/en/docs/components/trainer/operator-guides/runtime.md
```diff
@@ -99,157 +99,86 @@ spec:
...
-            spec:
-              template:
-                spec:
-                  containers:
-                    - name: node
-                      image: ghcr.io/kubeflow/trainer/torchtune-trainer
-```
+  ```yaml
+  trainer.kubeflow.org/deprecated: "true"
+  ```
```
andreyvelich, line 172: @astefanutti @tenzen-y I introduced the deprecated label to the runtimes.
Does it look good ?

content/en/docs/components/trainer/operator-guides/runtime.md
tenzen-y: Making TrainingRuntime mutable might cause the unhandling Pod updates. So, for now, we can have only deprecation notice in relase note and documentation.

After we introduced the revision mechanism, we can provide the automatic warning mechanism based on annotations.

content/en/docs/components/trainer/operator-guides/runtime.md
tenzen-y: https://github.com/kubeflow/website/pull/4167#discussion_r2266932545

content/en/docs/components/trainer/operator-guides/runtime.md
andreyvelich: >Making TrainingRuntime mutable might cause the unhandling Pod updates. So, for now, we can have only deprecation notice in relase note and documentation.

@tenzen-y In that case, how users can update the runtimes after we release the newer version of KF Trainer ?
For example, they should update the DeepSpeed runtime image to get the latest changes:
 ```yaml
image: ghcr.io/kubeflow/trainer/deepspeed-runtime:v2.1.0
```

Comment (tenzen-y) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: Currently, there is no way.

Comment (tenzen-y) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: So, I keep mentioning the importance of introducing a revision mechanism.

Comment (andreyvelich) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: Can you explain what unhandled Pod updates it can introduce if we only allow to update runtime that is unused by any TrainJob ?

Comment (astefanutti) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: It might be technically possible to prevent the Pod updates propagation only for that particular annotation no matter what, isn't?

Comment (andreyvelich) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: @astefanutti I don't understand what do you mean by preventing Pod updates ?
From my understanding if ClusterTrainingRuntime is unused (e.g. doesn't have any running TrainJob), we should allow users to update it.

Comment (astefanutti) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: @andreyvelich I was extrapolating to the case when the CTR is actually used, where we could imagine to relax the mutability and allow changes to some annotations and prevent updates propagation to pods.

Comment (andreyvelich) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: Make sense @astefanutti. I think, we discussed with @tenzen-y before that we should design the revision policy, so users can always see the original runtime (e.g. template) that was used to submit TrainJob.
I think, we can come up with the implementation later.

Comment (astefanutti) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: @andreyvelich I see, in case I did not manage to convey my point clearly enough, as a end-user I could probably accommodate the addition of the deprecation annotation on the original runtime, and I would still consider it the original version.

Comment (andreyvelich) [reply content/en/docs/components/trainer/operator-guides/runtime.md]: @astefanutti I see, however it might be tricky.
I imagine that when we release the newer version of Kubeflow Trainer, we publish newer version for all supported runtimes, for example:
```
kubeflow/trainer/deepspeed-runtime:v2.1.0
kubeflow/trainer/deepspeed-runtime:v2.2.0
```

And if we decide to deprecate DeepSpeed runtime in v2.2 version, and apply the deprecation label:
```
trainer.kubeflow.org/deprecated: "true"
```

Users have to update the image as well in their runtimes, when they run `kubectl apply -f ...`.

Does it make sense ?

astefanutti: @andreyvelich yes that makes sense, though this raises the question of the versioning of the runtimes w.r.t. the trainer / operator. I'd imagine the runtimes to be versioned independently based on the content of the runtimes as these are less coupled with the trainer itself. But I see that it makes things a bit more difficult to manage.

andreyvelich: For simplicity I would suggest that we keep all control plane components with the same version:
- controller manger
- builtin trainers
- runtimes
- initializers
- cache system

We've been following [this for Katib as well](https://github.com/kubeflow/katib/blob/release-0.18/manifests/v1beta1/installs/katib-standalone/kustomization.yaml), and it works quite well.

Users can decide whether they want to update/use some of those components as part of the KF Trainer release.

tenzen-y: > @andreyvelich I see, in case I did not manage to convey my point clearly enough, as a end-user I could probably accommodate the addition of the deprecation annotation on the original runtime, and I would still consider it the original version.

My point for annotations and labels propagation mechanism is the hidden recognition barrier.
The current labels and annotations propagation mechanism has clear ones for WHEN and WHICH.
The trainer propagates ALWAYS propagate ALL labels and annotations to underlying resources like JobSet.

But, if we introduce the internal criteria for WHEN and WHICH. The users could fall into a rabbit hole.
So, I want to keep the current those propagation mechanism.

tenzen-y: For unused CTR labels and annotations propagation, I don't disagree with that.
I wanted to mention about used CTRs.

andreyvelich: > The trainer propagates ALWAYS propagate ALL labels and annotations to underlying resources like JobSet.

@tenzen-y But we don't propagate labels from the `runtime.metadata` to the JobSet, isn't?
For example: [`trainer.kubeflow.org/framework: deepspeed` label](https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/deepspeed_distributed.yaml#L6C5-L6C46).

andreyvelich: @astefanutti @tenzen-y In any case, do we have any other blockers for this PR which explains the initial deprecation policy.

astefanutti: No other blockers on my side.

astefanutti:
/lgtm

Thanks!

content/en/docs/components/trainer/operator-guides/runtime.md
tenzen-y: > The trainer propagates ALWAYS propagate ALL labels and annotations to underlying resources like JobSet.
> @tenzen-y But we don't propagate labels from the runtime.metadata to the JobSet, isn't?
> For example: [trainer.kubeflow.org/framework: deepspeed label](https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/deepspeed_distributed.yaml#L6C5-L6C46).

Oh, I was imagined that we propagate runtime.metadata to the underlying objects. But, that is not correct: https://github.com/kubeflow/trainer/blob/f12a6d399a3dbb84d8829a5e7603ab310c45df6a/pkg/runtime/core/trainingruntime.go#L117-L132

So, I agree with relaxing the validation for the runtime.metadata.

tenzen-y:
Awesome!
/lgtm
/approve

content/en/docs/components/trainer/operator-guides/runtime.md
andreyvelich: Yeah! We only do this for `runtime.spec.template.metadata`

PR #4181, opened 2025-08-14
trainer: Fix the version in installation guide

andreyvelich:
We should appropriately set the `VERSION` env during installation.
We noticed that in some shell environment, it won't get saved without `export`

/cc @kubeflow/wg-training-leads @astefanutti @raravena80


astefanutti:
/lgtm

Thanks!

PR #4198, opened 2025-09-03
trainer: Update the get_job_logs() API

andreyvelich:
Part of: https://github.com/kubeflow/sdk/issues/75

We should update the `get_job_logs()` API in Trainer SDK.

/assign @kubeflow/kubeflow-trainer-team @kubeflow/kubeflow-sdk-team

### DIFF
--- a/content/en/docs/components/trainer/user-guides/builtin-trainer/torchtune.md
+++ b/content/en/docs/components/trainer/user-guides/builtin-trainer/torchtune.md
@@ -107,8 +107,7 @@ We can use the `get_job_logs()` API to get the TrainJob logs.
 ```python
 from kubeflow.trainer.constants import constants
 
-log_dict = client.get_job_logs(job_name, step=constants.DATASET_INITIALIZER)
-print(log_dict[constants.DATASET_INITIALIZER])
+print("\n".join(client.get_job_logs(job_name, step=constants.DATASET_INITIALIZER)))
 ```
 
 Output:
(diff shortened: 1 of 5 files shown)


content/en/docs/components/trainer/user-guides/builtin-trainer/torchtune.md
```diff
@@ -107,8 +107,7 @@ We can use the `get_job_logs()` API to get the TrainJob logs.
 ```python
 from kubeflow.trainer.constants import constants
 
-log_dict = client.get_job_logs(job_name, step=constants.DATASET_INITIALIZER)
-print(log_dict[constants.DATASET_INITIALIZER])
+print("\n".join(client.get_job_logs(job_name, step=constants.DATASET_INITIALIZER)))
```
kramaranya, line 110: Why not this way?
```python
for logline in TrainerClient().get_job_logs(job_id, step=constants.DATASET_INITIALIZER, follow=True):",
    print(logline)
```

andreyvelich: That will also work. I just try to show users different ways to print the logs.
In this doc, we just print logs without `follow=True`.

kramaranya: I agree that we should show different ways :) however on the website we only show examples with a string
Maybe on the getting started page we could add a note that you can use `follow=True`?

andreyvelich: Good point, let me change the getting started example.

kramaranya:
/lgtm
Thank you!

PR #4137, opened 2025-06-24
website: fix header and announcements

thesuperzapper:
**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] Ensure you follow best practices from our guide. [Contributing](https://github.com/kubeflow/website/blob/master/content/en/docs/about/contributing.md). 
- [x] You have included screenshots when changing the website style or adding a new page.

### Description

Partially reverts: https://github.com/kubeflow/website/pull/4136

Currently, after https://github.com/kubeflow/website/pull/4136 the header of the website is broken, it does not wrap correctly and will extend beyond the edge of the screen, causing it to not be possible to see the version or dark/light drop-downs.

This PR:
- reverts the incorrect style update which prevented wrapping of header text on smaller screens
- reduces the width of the announcements text (say "Kubeflow Summit" which is shorter)
- uses the standard format of "date", "location" for the icons below the Kubeflow Summit entry
- comments out the GSOC one, as we can only fit 3 announcements, and GSOC is already underway, so doesn't need a header announcement

#### After this PR:

Screenshot 2025-06-24 at 09 39 23

#### Before this PR:

Screenshot 2025-06-24 at 09 39 39

Screenshot 2025-06-24 at 09 38 49

### Labels

/area website

---

PR #4232, opened 2025-11-08
website: Add Kubeflow SDK Survey URL as announcement on the website

szaher:
### Description of Changes

Add Kubeflow SDK Survey to the website 
image

### Related Issues

Closes: #<issue number>

Related: #<issue number>

### Checklist

- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] Ensure you follow best practices from our [contributing guide](https://github.com/kubeflow/website/blob/master/content/en/docs/about/contributing.md).
- [ ] (for big changes) I will post screenshots of the changes in a PR comment

### DIFF
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -40,14 +40,22 @@
       <!-- =================================== -->
       <!-- === ANNOUNCEMENT BAR START === -->
       <!-- =================================== -->
+      {{- $announcementEnable := .Site.Params.announcement.enable | default false -}}
+      {{- if $announcementEnable -}}
       <div class="text-center w-100 pt-3 pt-md-2 pb-2 pb-md-2"> <!-- Adjusted padding for desktop -->
-        <span class="px-3 py-1 rounded" style="background-color: rgba(255, 255, 255, 0.15); font-size: 0.9em; font-weight: 500;">
-          <!-- Link added here. Change href="#" to your desired URL -->
-          <a href="#" class="text-white" style="text-decoration: none;">
-            ✨ <strong>Announcement:</strong> <a target="_blank" href="https://docs.google.com/forms/d/e/1FAIpQLSet_IAFQzMMDWolzFt5LI9lhzqOOStjIGHxgYqKBnVcRtDfrw/viewform?usp=dialog" class="text-white" style="text-decoration: underline;">Check out Kubeflow Unified SDK Survey!</a> ✨
-          </a>
+        <span class="px-3 py-1 rounded text-white" style="background-color: rgba(255, 255, 255, 0.15); font-size: 0.9em; font-weight: 500;">
+          {{- $icon := .Param "announcement.icon" | default "" -}}
+          {{- $text := .Param "announcement.text" | default "Announcement:" -}}
+          {{- $linkText := .Param "announcement.link_text" | default "" -}}
+          {{- $url := .Param "announcement.url" | default "#" -}}
+          {{- if $icon }}{{ $icon }} {{ end -}}<strong>{{ $text | safeHTML }}</strong>
+          {{- if and $linkText (ne $url "#") -}}
+            <a target="_blank" href="{{ $url }}" class="text-white" style="text-decoration: underline;">{{ $linkText | safeHTML }}</a>
+          {{- end -}}
+          {{- if $icon }} {{ $icon }}{{ end -}}
         </span>
       </div>
+      {{- end -}}
       <!-- =================================== -->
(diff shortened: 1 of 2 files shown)


layouts/partials/navbar.html
```diff
@@ -33,6 +33,24 @@
...
+    <div class="d-flex flex-column w-100">
+
+      <!-- =================================== -->
+      <!-- === ANNOUNCEMENT BAR START === -->
+      <!-- =================================== -->
+      <div class="text-center w-100 pt-3 pt-md-2 pb-2 pb-md-2"> <!-- Adjusted padding for desktop -->
+        <span class="px-3 py-1 rounded" style="background-color: rgba(255, 255, 255, 0.15); font-size: 0.9em; font-weight: 500;">
+          <!-- Link added here. Change href="#" to your desired URL -->
+          <a href="#" class="text-white" style="text-decoration: none;">
+            ✨ <strong>Announcement:</strong> <a target="_blank" href="https://docs.google.com/forms/d/e/1FAIpQLSet_IAFQzMMDWolzFt5LI9lhzqOOStjIGHxgYqKBnVcRtDfrw/viewform?usp=dialog">Check out Kubeflow Unified SDK Survey!</a> ✨
```
andreyvelich: Is it possible to make link same colour as Announcement ? I think, it would be better
Screenshot 2025-11-09 at 00 15 57

Something like this if that works

```
<p style="color: white;">
  ✨ <strong>Announcement:</strong> 
  <a 
    target="_blank" 
    href="https://docs.google.com/forms/d/e/1FAIpQLSet_IAFQzMMDWolzFt5LI9lhzqOOStjIGHxgYqKBnVcRtDfrw/viewform?usp=dialog" 
    style="color: white; text-decoration: underline;"
  >
    Check out Kubeflow Unified SDK Survey!
  </a> 
  ✨
</p>
```

kramaranya: Shall we add that alongside with other items in menu instead? https://github.com/kubeflow/website/blob/b2ed7df8ca05f385f2bcb5d49404c58744a489ce/config.toml#L63-L75
Similar to https://github.com/kubeflow/website/pull/4137#issue-3172628377

andreyvelich: It's fine too if that is easier.

szaher: updated the link color. 
@kramaranya we tried that too but the top link will be probably more visible to users and will be in all pages

szaher:
@andreyvelich @kramaranya 
I have updated my code to make the announcement configurable from `config.toml` so we don't have to change the code everytime we have a new announcement to make on the website

image

kramaranya:
cc @thesuperzapper

andreyvelich:
Thanks!
/lgtm
/approve

ederign:
Hi folks, I believe this PR kind of hide the the docs search bar:

Screenshot 2025-11-25 at 12 45 10 PM

Would it be okay to make this announcement only on the home page?

andreyvelich:
Good point @ederign, could we move the survey announcement to the left side where Kubeflow Summit is?
So these two messages will be on the same line.

kramaranya:
> Good point @ederign, could we move the survey announcement to the left side where Kubeflow Summit is? So these two messages will be on the same line.

Agree! We can do that as part of https://github.com/kubeflow/website/pull/4246 I believe

thesuperzapper:
@kramaranya To avoid the need to fix complex style issues, let's just make an "event" that sits to the left of the KubeCon which is actually an "announcement" for the survey. 

We should revert https://github.com/kubeflow/website/commit/dbe4e37a7441299ab7aa012e5bccde88610b6c0b because it really breaks the readably of the docs sections.

_PS: In future we can do a dismissible "above the page" announcement" sitting above the kubeflow logo event, like most other websites do, because it won't interfere with the page style._

kramaranya:
> @kramaranya To avoid the need to fix complex style issues, let's just make an "event" that sits to the left of the KubeCon which is actually an "announcement" for the survey.
> 
> We should revert [dbe4e37](https://github.com/kubeflow/website/commit/dbe4e37a7441299ab7aa012e5bccde88610b6c0b) because it really breaks the readably of the docs sections.
> 
> _PS: In future we can do a dismissible "above the page" announcement" sitting above the kubeflow logo event, like most other websites do, because it won't interfere with the page style._

I agree, I suggested this in https://github.com/kubeflow/website/pull/4232#discussion_r2508085959
@voodoovampire, would you mind adjusting https://github.com/kubeflow/website/pull/4246 to address our comments please?

PR #4250, opened 2025-11-27
website: move Kubeflow SDK survey to the menu bar

kramaranya:
I've reverted #4232 since it hides the search bar and moved Kubeflow SDK survey to the menu bar instead


kramaranya:
Screenshot 2025-11-28 at 00 32 48
cc @ederign @andreyvelich @thesuperzapper @szaher

Issue #4256, opened 2025-12-10
[Feedback] docs/components/spark-operator/user-guide/running-multiple-instances-of-the-operator.md


dpengpeng:
How to deploy the spark operator in different namespace? In this case, resources such as MutatingWebhookConfiguration, ValidatingWebhookConfiguration, and ClusterRole will conflict. How to solve this problem?

dpengpeng:
@liuqi @ewilderj @onelittlenightmusic @aronchick

Sanika123-art:
Hi @dpengpeng 👋  
Thanks for raising this question.

I’m new to Kubeflow and would like to help improve the documentation around this topic.  
I’m trying to understand the recommended approach for running Spark Operator across multiple namespaces, especially considering cluster-scoped resources like webhooks and ClusterRoles.

Once clarified, I can help update the docs to clearly explain:
- whether multiple instances are supported
- recommended deployment patterns
- limitations and workarounds

Please let me know if that would be helpful

PR #4303, opened 2026-02-08
docs: add warning about multi-namespace Spark Operator deployments

Sanika123-art:
### What this PR does

Adds an **Important** alert to the Spark Operator documentation explaining that
the operator creates cluster-scoped resources (such as webhooks and ClusterRoles),
which makes running instances in different namespaces within the same cluster
unsupported.

### Why this change is needed

Users attempting to deploy multiple Spark Operator instances may encounter
resource conflicts that are not clearly documented today. This change makes the
limitation explicit and guides users toward the recommended approach of using a
single operator namespace with different Spark job namespaces.

### Related issue

- Feedback: docs/components/spark-operator/user-guide/running-multiple-instances-of-the-operator.md (#4256)

### DIFF
--- a/content/en/docs/components/spark-operator/user-guide/running-multiple-instances-of-the-operator.md
+++ b/content/en/docs/components/spark-operator/user-guide/running-multiple-instances-of-the-operator.md
@@ -5,6 +5,20 @@ description: |
 weight: 70
 ---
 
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

PR #4328, opened 2026-02-25
feat(docs): add copy-to-clipboard button for code blocks (#4327)

yashwanth510:
Fixes #4327

All code blocks now have global copy-to-clipboard support.
tested locally on a Hugo server.

Issue #4210, opened 2025-09-29
[Feedback] docs/components/pipelines/operator-guides/installation/_index.md


De-Silva-Madhushankha:
Add “copy to clipboard” button to code snippets in installation docs

In the Kubeflow Pipelines installation document (e.g., [https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/](https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/?utm_source=chatgpt.com)
), the code snippets are not easily copyable because there is no “copy to clipboard” or “copy” button on the snippet blocks.

Expected behavior

Each code snippet block should have a “copy” button (e.g., a small icon “📋” or “Copy”) that, when clicked, copies the snippet to the clipboard.

This makes it easier for developers to copy commands accurately, reduces errors, and improves usability.

Why this is important:

Developers often have to manually select, drag, and copy code; that can introduce mistakes (e.g., extra spaces, missing lines).

Especially for installation/setup docs, quick and accurate copying is valuable.

Many other documentation frameworks support built-in “copy button” functionality for code blocks, so it’s a reasonably standard feature.

killjoycircuit:
Hi I would like to work on this issue, is this still open and available to work on?

EncHawk:
I was about to create an issue for the same, I would love to take this up.

PR #4393, opened 2026-06-09
website : Add copy to clipboard and language banner for code snippets

### LINKED ISSUE #4327: feature(Website): copy to clickboard button.
A copy to clipboard button that lets the readers of the documentation copy the commands that are mentioned.
**Why is this needed**
if the command is long, the overflow on x makes it hard to copy it all at once. Besides the ease to click a button to copy the command makes the DX better

**Page to Update or indicate if it is a new page (provide the full path):**
https://www.kubeflow.org/docs/components/pipelines/legacy-v1/installation/localcluster-deployment/#deploying-kubeflow-pipelines 
is one such place where the kubeflow installation command is too large.

**Component/Kubeflow Version:** irrelevant.

**Comments**
there is an issue that exists, but its stale and i just would like to raise this again. #4210 
Image

HeythisisSud:
### Description of Changes
What was changed:

assets/js/click-to-copy.js — Rewrote the existing (disabled) copy-to-clipboard script with a new, dependency-free implementation that:

Injects a header bar (.c2c-header) at the top of each .highlight code block containing a language label (e.g. "Terminal", "YAML", "Python") and a copy icon button.
Automatically wraps any bare pre tag elements in a .highlight container so all code blocks render consistently.
Uses the modern navigator.clipboard API with a legacy execCommand fallback for broader browser support.
Switches the button to a checkmark icon with a "Copied!" tooltip on success, reverting after 2 seconds.
assets/scss/_styles_project.scss — Added styles for the copy button feature:

Converts .td-content .highlight to a flex column so the header sits above the pre tag block.
Styles the .c2c-header bar with a language badge on the left and copy button on the right.
Full light and dark theme support using @at-root [data-theme="light/dark"] to correctly target the site's theme attribute.
Resets browser default <button> styles (appearance: none; background: none) to prevent unstyled grey button rendering.
Includes a smooth floating tooltip and keyboard accessibility (focus-visible outline).
layouts/partials/scripts.html — Enabled the click-to-copy script, which was previously gated behind a hardcoded false condition.

Design decisions:

The copy button is placed in a dedicated header bar above the code, so it never overlaps or obscures code text.
Button opacity and hover states follow the existing Kubeflow brand color (#4279f4).
Success state uses green (#10b981 light / #34d399 dark) consistent with standard UX conventions.
No external dependencies added (no Popper.js, no extra libraries).

### Related Issues

Closes: #4327 #4328 #4210 

Screenshot 2026-06-10 020249

### DIFF
--- a/assets/js/click-to-copy.js
+++ b/assets/js/click-to-copy.js
@@ -1,0 +1,105 @@
+
+(function () {
+  'use strict';
+
+  const COPY_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><rect x="9" y="9" width="13" height="13" rx="2" ry="2"></rect><path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"></path></svg>`;
+  const CHECK_ICON = `<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="3" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><polyline points="20 6 9 17 4 12"></polyline></svg>`;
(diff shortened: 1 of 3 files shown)


Arhell (commented):
/ok-to-test

PR #4416, opened 2026-07-01
community: Add meeting links to the website overview

christian-heusel:
### Description of Changes

Add meeting links to the website overview

_Note_: Even though one component in the meeting link is called "password" it's value is actually publicly available (verified by viewing the calandar in a firefox private tab).

If there is consensus that this is something we want I'll also propose the same change for the WG List in the community repo 🤗 

image

### Related Issues

_none_

### Checklist

- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] Ensure you follow best practices from our [contributing guide](https://github.com/kubeflow/website/blob/master/content/en/docs/about/contributing.md).
- [x] (for big changes) I will post screenshots of the changes in a PR comment

---

/cc @thesuperzapper


thesuperzapper (changes requested):
I think it will be more clear what you should click to register, if we just make a fourth column which says "register" on every row, similar to how we do it for the YouTube playlist.

Right now it's really unclear that the name of the meeting is actually the link to register.

thesuperzapper:
Thanks, this will make it so much easier to point people to the right place to register for each of our meetings.

/lgtm
/approve

Issue #3629, opened 2023-11-17
Upgrade to Docsy >= 0.8.0 for full support of GA4


chalin:
Docsy 0.8.0 will very likely be released before the end of the month (November). Release 0.8.0 brings in full support for Google Analytics 4.

To track progress on the release, see:

- https://github.com/google/docsy/issues/1667

/cc @nate-double-u

thesuperzapper:
As we are [currently on v0.6.0](https://github.com/kubeflow/website#docsy-theme) (after PR https://github.com/kubeflow/website/pull/3609), and we use a lot of overrides, this upgrade will be very painful.

It's possible even that just backporting the fixes might be easier.

But if you or someone else is open to helping, that would be great.

varodrig:
cc @thesuperzapper

PR #4457, opened 2026-08-15
website: Send feedback events with GA4

immanuwell:
### Description of Changes

The feedback widget still calls the old `ga()` API, while production only loads GA4's `gtag()`. Right now its a silent no-op

This sends a `page_helpful` event through `gtag()` and keeps the existing Yes and No values and UI behavior

Repro:
1. Open any docs page and click Yes or No.
2. Check `dataLayer`. No feedback event is added before this fix.
3. With this fix, one `page_helpful` event is added with the page path and value `1` or `0`.

Tests: Hugo 0.124.1 production build, browser checks for both buttons and the no analytics fallback, Netlify link check with 2421 passes and 1 tolerated external timeout.

### Related Issues

Related: #3629

### Checklist

- [x] Signed off the commit
- [x] Followed the contributing guide
- [x] Small behavior-only change, no screenshot needed

### DIFF
--- a/layouts/partials/feedback.html
+++ b/layouts/partials/feedback.html
@@ -27,16 +27,12 @@
     noButton.disabled = true;
   };
   const sendFeedback = (value) => {
-    if (typeof ga !== 'function') return;
-    const args = {
-      command: 'send',
-      hitType: 'event',
-      category: 'Helpful',
-      action: 'click',
-      label: window.location.pathname,
-      value: value
-    };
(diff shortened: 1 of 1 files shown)


jaiakash (commented):
Hi @immanuwell Should this work for deployed preview?
I tried here https://deploy-preview-4457--competent-brattain-de2d6d.netlify.app/ but still there no event registered. Can you please confirm.

PR #4462, opened 2026-08-21
website: reserve room for the fixed navbar when scrolling to an anchor

### LINKED ISSUE #4460: bug(website): in-page anchors land behind the fixed navbar
**This is a Bug Report**

**Problem:**

The navbar is `position: fixed` and 106px tall, and nothing compensates for it when the browser scrolls to a fragment. So following any in-page anchor lands the target heading **underneath** the navbar, completely hidden.

Measured on `https://www.kubeflow.org/docs/started/introduction/#what-is-kubeflow`:

```
nav.js-navbar-scroll   position: fixed   top: 0    bottom: 106
#what-is-kubeflow                        top: 0    bottom: 119
```

The heading is covered by 106px of navbar. Neither compensation property is set anywhere on the site:

```
html { scroll-padding-top: auto }   /* i.e. unset */
h2   { scroll-margin-top: 0px }
```

This is not limited to typed URLs. It happens through the site's own navigation: on `/docs/components/pipelines/overview/`, clicking **"Why Kubeflow Pipelines?"** in the right-hand table of contents puts that heading at `top: 0`, hidden by the same 106px. Every heading anchor and every table-of-contents entry in the docs behaves this way.

The reader ends up looking at the middle of a section with no visible heading, having just clicked the heading's name. It is most disorienting for anyone using magnification, where the navbar takes proportionally more of the viewport, and for anyone who navigates by heading and then looks to confirm where they landed.

**Proposed Solution:**

One declaration in `assets/scss/_styles_project.scss`, so the browser leaves room for the fixed navbar when it scrolls to a fragment:

```scss
html {
  scroll-padding-top: 106px; // height of the fixed .js-navbar-scroll
}
```

Better still, drive it from the same variable the navbar height comes from, so the two cannot drift apart. `scroll-padding-top` is the modern, single-declaration answer to this and needs no JavaScript.

**Page to Update (provide the full path):**
https://www.kubeflow.org/docs/started/introduction/#what-is-kubeflow
https://www.kubeflow.org/docs/components/pipelines/overview/ (via the right-hand table of contents)

Site-wide rather than page-specific, so the fix belongs in the site stylesheet.

**Component/Kubeflow Version:**
Website. `layouts/partials/navbar.html` in this repo, not inherited from Docsy. `scroll-padding` does not currently appear anywhere in the repository.

**Additional Information:**

Verified in Chromium at 375px, 1280px and 1920px. The navbar stays `position: fixed` at all three, so this is not viewport-specific.

This may also be what #4400 was reaching for. That issue reports the header overlapping content on the homepage, which I could not reproduce at any of those three widths, and neither could another contributor in July: at scroll-top the `h1` sits at 176px against a navbar ending at 106px, a 70px gap. Content hidden behind the navbar after following an anchor is a different mechanism and does reproduce reliably.

opensource-joe:
### Description of Changes

`.td-navbar` is `position: fixed` from the `md` breakpoint up, and nothing compensated for it, so following any in-page anchor scrolled the target heading underneath the navbar.

One declaration in `assets/scss/_variables_project.scss`, next to the existing `html.smooth-scroll` rule, scoped to the same breakpoint at which the navbar becomes fixed.

Measured on `/docs/started/introduction/#what-is-kubeflow`, before and after:

| Viewport | Navbar | Heading top, before | after |
|---|---|---|---|
| 1920 | fixed, 106px | 0.2px (covered) | 125.2px |
| 1280 | fixed, 106px | 0.2px (covered) | 125.2px |
| 992 | fixed, 106px | 0.2px (covered) | 125.2px |
| 768 | fixed, 109px | 0.2px (covered) | 125.2px |
| 375 | not fixed | 0.2px (fine) | unchanged |

Also checked the table of contents path from the issue: clicking "Why Kubeflow Pipelines?" on `/docs/components/pipelines/overview/` now lands the heading at 125.2px instead of 0.2px.

Two notes for reviewers:

- The rule is scoped to `min-width: 768px` because the navbar is only fixed there. Below that it scrolls away with the page and needs no compensation, so an unscoped rule would add dead space on mobile.
- It has to live in `_variables_project.scss`. `_docsy_main.scss` and `_styles_project.scss` are imported inside the `[data-theme=...]` blocks in `main.scss`, so an `html` rule in either would compile to `[data-theme=light] html` and never match. That file is imported before Bootstrap's mixins, hence the written-out breakpoint instead of `media-breakpoint-up(md)`.

Tested against a local build at `476e4c1` with Hugo extended 0.124.1, the version pinned in `netlify.toml`. `hugo --gc --minify` builds clean.

### Related Issues

Closes: #4460

### Checklist

- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] Ensure you follow best practices from our [contributing guide](https://github.com/kubeflow/website/blob/master/content/en/docs/about/contributing.md).
- [ ] (for big changes) I will post screenshots of the changes in a PR comment

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -6,6 +6,23 @@ html.smooth-scroll {
   scroll-behavior: smooth;
 }
 
+// `.td-navbar` is `position: fixed` only from Bootstrap's `md` breakpoint up, so
+// only reserve room for it there. Without this, following an in-page anchor or a
+// table of contents entry scrolls the target heading underneath the navbar. The
+// navbar is 109px tall between `md` and `lg` and 106px above that; the extra 1rem
+// keeps the heading off it.
+//
+// This has to live in this file. `_docsy_main.scss` and `_styles_project.scss` are
+// imported inside the `[data-theme=...]` blocks in `main.scss`, so an `html` rule
+// there would compile to `[data-theme=light] html` and never match. This file is
+// imported before Bootstrap's mixins, hence the written-out breakpoint rather than
+// `media-breakpoint-up(md)`.
+@media (min-width: 768px) {
+  html {
+    scroll-padding-top: 125px;
+  }
+}
+
 // Bootstrap container-max-widths
 // NOTE: we increase the max-widths to make the content wider compared to the Bootstrap defaults
 $container-max-widths: (


Arhell (commented):
/ok-to-test

christian-heusel (approved):

jaiakash (approved):

PR #4479, opened 2026-09-04
notebooks: (actually its workspaces) add Spark Connect operator and user guides

vikas-saxena02:
## What this does

Adds two guides covering how to run Spark from a Kubeflow Workspace using the Kubeflow SDK, so users can start a Spark Connect session from a notebook without hand-authoring `SparkApplication` manifests.

Addresses kubeflow/notebooks#1362. Scope confirmed with @christian-heusel on that issue: guides live here, no sample WorkspaceKind, the docs instead explain what to modify on an existing WorkspaceKind and why.

## Changes

Three new files, nothing modified or removed.

- `workspaces/operator-guides/spark-connect.md` — the one WorkspaceKind change an administrator has to make (binding `kubeflow-spark-edit` via `podTemplate.serviceAccount.clusterRoles`), why the narrow role rather than `kubeflow-edit`, how to verify with `kubectl auth can-i`, and image requirements.
- `workspaces/user-guides/spark-connect.md` — the `SparkClient` / `connect()` flow, version pinning, the three required `PodTemplateOverride` settings with the reasoning for each, and troubleshooting.
- `workspaces/user-guides/_index.md` — new section, since Workspaces had Operator and Contributor guides but no User guides.

## How the split landed

Worth flagging, because it isn't what I expected going in: the only administrator-side change is the RBAC binding. The driver ServiceAccount, Istio sidecar exclusion, and the executor container name are all supplied per session by the user in the `connect()` call, not configured on the WorkspaceKind. So the operator guide is short and most of the substance sits in the user guide.

On the Notebooks-side follow-ups asked for in kubeflow/notebooks#1362: the only candidate I found was a sample `jupyterlab-spark` WorkspaceKind, which @christian-heusel has already ruled out. I don't think anything else is needed in `kubeflow/notebooks` for this.

## Depends on kubeflow/community-distribution#3590

**Please hold this until that PR merges.** Both guides link to the integration test it adds, so the reference links 404 until then and the Netlify link check will flag them. Every configuration snippet here is taken from that test rather than written from scratch, so it should also merge first for the docs to describe verified behaviour.

Raising now so review can start in parallel. Happy to convert to a draft if you'd rather.

## Related

kubeflow/spark-operator#3141 replaces the Jupyter Enterprise Gateway notebooks guide with the same SDK approach, on the Spark Operator docs side. It states the platform-specific requirements inline as a stopgap; I'll swap those for links to these pages once both land.

### DIFF
--- a/content/en/docs/components/workspaces/operator-guides/spark-connect.md
+++ b/content/en/docs/components/workspaces/operator-guides/spark-connect.md
@@ -1,0 +1,96 @@
+---
+title: Enabling Spark
+description: How to let Workspace users start Spark Connect sessions
+weight: 20
+---
+
+{{< workspaces-beta-notice >}}
+
+This guide is for platform administrators. It covers the one change you must make so that
+Workspace users can start a Spark Connect session with the
+[Kubeflow SDK](https://sdk.kubeflow.org/en/latest/spark/index.html).
+
+Everything else is configured by the user at session creation time — see the
+[user guide](/docs/components/workspaces/user-guides/spark-connect/).
+
+## Prerequisites
+
+- Kubeflow Workspaces `{{% kf-workspaces-version %}}` or later
+- [Kubeflow Spark Operator](https://spark.kubeflow.org/) with the `SparkConnect` CRD
+  installed and the operator watching your user namespaces
+- The aggregated `kubeflow-spark-edit` ClusterRole present in the cluster
+
+## Grant the Workspace access to Spark
+
+Each Workspace runs under its own ServiceAccount, `ws-{WORKSPACE_NAME}`.
+
+Since Workspaces `v2.0.0-beta.0` this ServiceAccount **no longer inherits the Kubeflow
+ClusterRoles** the way the old shared `default-editor` did. Until you grant it explicitly,
+a Workspace cannot create `SparkConnect` resources.
+
+Bind the Spark role on the WorkspaceKind:
+
+```yaml
+spec:
+  podTemplate:
+    serviceAccount:
+      clusterRoles:
+        # create/get/list/delete sparkconnects and sparkapplications
+        - name: "kubeflow-spark-edit"
+```
+
+Bind `kubeflow-spark-edit`, not `kubeflow-edit`. The Spark role grants only what is needed
+to manage `SparkApplication` and `SparkConnect` resources; `kubeflow-edit` is far broader
+and reintroduces exactly the permissions the per-Workspace ServiceAccount change removed.
+
+That is the only WorkspaceKind change required.
+
+## Verify
+
+```bash
+kubectl auth can-i create sparkconnects \
+  --as="system:serviceaccount:${NAMESPACE}:ws-${WORKSPACE_NAME}" \
+  -n "${NAMESPACE}"
+```
+
+If this returns `no`, fix the binding before users try to connect. Without the permission
+`connect()` does not fail fast — it fails on a multi-minute Spark timeout, well away from
+the actual cause.
+
+## What users still have to configure
+
+Three settings are supplied by the user in the `connect()` call rather than by you, because
+they apply per session. They are documented in the
+[user guide](/docs/components/workspaces/user-guides/spark-connect/), but they are worth
+knowing about when a user reports a broken session:
+
+**The driver needs a different ServiceAccount.** `kubeflow-spark-edit` lets the Workspace
+create the `SparkConnect` resource, but the Spark driver pod separately creates and watches
+its own executor pods. That is not covered by the Spark role, so the driver runs as
+`default-editor`.
+
+**Spark pods must leave the Istio mesh.** The driver and executors exchange data on ports
+7078 and 7079. Executors register with the driver and then fail fetching broadcast blocks,
+so tasks never complete. Both pods set `sidecar.istio.io/inject: "false"`.
+
+**The executor container must be named `spark-kubernetes-executor`.** Spark submits with
+`executor.podTemplateContainerName=spark-kubernetes-executor` and looks the container up by
+that name, so a template must declare it even though the operator supplies the image.
+
+## Image requirements
+
+The Workspace image needs the Kubeflow SDK with its `spark` extra (`kubeflow[spark]`), plus
+a `pyspark-connect` release matching the Spark version the SDK provisions, which it exposes
+as `kubeflow.spark.backends.kubernetes.constants.DEFAULT_SPARK_VERSION`.
+
+Installing the extra alone is not enough: it pins a `pyspark-connect` release that does not
+currently match that constant. Build images against the constant rather than relying on the
+extra or choosing a Spark version independently — a mismatch surfaces as an opaque protocol
+error rather than a version complaint.
+
+## Reference
+
+The end-to-end setup is exercised in CI by the Workspaces Spark Connect test in
+[`kubeflow/community-distribution`](https://github.com/kubeflow/community-distribution/tree/master/tests):
+`workspaces_spark_connect_test.sh`, `workspacekind.spark.test.yaml`, and
+`spark_connect_from_workspace.py`. That test is the authoritative working example.
--- a/content/en/docs/components/workspaces/operator-guides/spark-connect.md
+++ b/content/en/docs/components/workspaces/operator-guides/spark-connect.md
@@ -15,7 +15,7 @@ Everything else is configured by the user at session creation time — see the
 
 ## Prerequisites
 
-- Kubeflow Workspaces `{{% kf-workspaces-version %}}` or later
+- Kubeflow Workspaces `{{% workspaces/kf-workspaces-version %}}` or later
 - [Kubeflow Spark Operator](https://spark.kubeflow.org/) with the `SparkConnect` CRD
   installed and the operator watching your user namespaces
 - The aggregated `kubeflow-spark-edit` ClusterRole present in the cluster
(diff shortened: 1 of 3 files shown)


vikas-saxena02:
@nabuskey @tariq-hasan @juliusvonkohout can I please get  a review on this

content/en/docs/components/workspaces/user-guides/spark-connect.md
```diff
@@ -0,0 +1,133 @@
...
+            },
+        ),
+        PodTemplateOverride(
+            role="executor",
+            template={
+                "metadata": {"labels": {"sidecar.istio.io/inject": "false"}},
+                "spec": {"containers": [{"name": "spark-kubernetes-executor"}]},
+            },
+        ),
+    ]
```
tariq-hasan, line 69: This option is not supported in SDK v0.5.

In addition, for a first-timer user, we should not expose options - those are meant for advanced users who are comfortable with k8s internals.

More to the point the entire reason we even had https://github.com/kubeflow/spark-operator/issues/3049 to begin with was to remove the poor UX that would come from making a user hand-write a service account into the connect API.

If this is targeting v0.5 then for now we can configure the [driver](https://github.com/kubeflow/sdk/blob/d9191052dfcbb82108909dc36bcc01b8d8daf40b/kubeflow/spark/types/types.py#L88) dataclass in the [connect](https://github.com/kubeflow/sdk/blob/d9191052dfcbb82108909dc36bcc01b8d8daf40b/kubeflow/spark/api/spark_client.py#L69) API to hand-write the service account.

But this portion would need to be edited again for SDK v0.6 and Spark Operator v2.6.0.

cc @andreyvelich @nabuskey

tariq-hasan: To add to what I wrote above we should not make a user (data scientist) configure Istio. That's an admin-level concern.

content/en/docs/components/workspaces/operator-guides/spark-connect.md
```diff
@@ -0,0 +1,96 @@
...
+## Image requirements
+
+The Workspace image needs the Kubeflow SDK with its `spark` extra (`kubeflow[spark]`), plus
+a `pyspark-connect` release matching the Spark version the SDK provisions, which it exposes
+as `kubeflow.spark.backends.kubernetes.constants.DEFAULT_SPARK_VERSION`.
+
+Installing the extra alone is not enough: it pins a `pyspark-connect` release that does not
+currently match that constant. Build images against the constant rather than relying on the
+extra or choosing a Spark version independently — a mismatch surfaces as an opaque protocol
+error rather than a version complaint.
```
tariq-hasan, line 89: @andreyvelich Would it make sense to have these instructions or would it ideal to have `pyspark-connect` match `DEFAULT_SPARK_VERSION` in the SDK so this workaround can be skipped?

content/en/docs/components/workspaces/user-guides/spark-connect.md
```diff
@@ -0,0 +1,133 @@
...
+pip install "pyspark-connect==${SPARK_VERSION}"
+```
+
+## Start a session
+
+```python
+from kubeflow.common.types import KubernetesBackendConfig
+from kubeflow.spark import Name, PodTemplateOverride, SparkClient
+
+client = SparkClient(backend_config=KubernetesBackendConfig(namespace="my-namespace"))
```
tariq-hasan, line 50: Since the permissions are namespaced do we need to be explicit that this is the namespace where the workspace permissions are enabled?

content/en/docs/components/workspaces/user-guides/spark-connect.md
```diff
@@ -0,0 +1,133 @@
...
+The two overrides are not optional. Each one works around something that would otherwise
+fail in a way that does not point at its cause:
+
+`serviceAccountName: default-editor` — your Workspace ServiceAccount can create the
+`SparkConnect` resource, but the driver pod separately creates and watches its own executor
+pods, which that role does not cover. Without this, executors never appear.
+
+`sidecar.istio.io/inject: "false"` — the driver and executors exchange data on ports 7078
+and 7079, and that traffic does not survive the service mesh. Executors register with the
+driver, then fail fetching broadcast blocks, so tasks hang rather than error.
```
tariq-hasan, line 82: Would this need a more detailed investigation to dissect what the actual issue might be?

content/en/docs/components/workspaces/user-guides/spark-connect.md
tariq-hasan, line 69: Assuming we drop out of Istio we should still not make a user (data scientist) configure Istio. That's an admin-level concern.

It's better for Istio to be configured in the cluster to not process injection for Spark driver and executor pods that the operator auto-labels with `sparkoperator.k8s.io/launched-by-spark-operator: "true"`.

Assuming we do go with this approach this should therefore be an instruction for the platform admin not for the SDK user.

tariq-hasan (commented):
Hi @vikas-saxena02! I have added a few comments.

content/en/docs/components/workspaces/user-guides/spark-connect.md
vikas-saxena02: well, in that case, I am happy to jump into the xt SDK call so that we can come up with a standard patternw hich everyone is happy with.

content/en/docs/components/workspaces/user-guides/spark-connect.md
vikas-saxena02: well, the idea is to start the sparkConnect session in that particular namespace. Default is `default` namespace which no one uses in actual prod like scenarios

content/en/docs/components/workspaces/user-guides/spark-connect.md
vikas-saxena02: its in my things to do list but given my limited knowledge of istio, it will take time for sure

content/en/docs/components/workspaces/user-guides/spark-connect.md
vikas-saxena02: actually this is what the community distribution does enforce by default but the spark-opertor on itself, if installed by helm charts directly, does not. Should we make it a default on the spark-operator itself for now ot do you want me to do a detailed investigation first.

vikas-saxena02: currently its sdk 0.0.4 that ships with the default notebook/worspace image

juliusvonkohout (changes requested):
I think you can shorten it a lot and ease the maintenance if you just link to https://github.com/kubeflow/community-distribution/blob/master/.github/workflows/workspaces_spark_connect_test.yaml 

We do the same for kserve.  just the high level and a link to constantly tested automation to reduce the maintenance and prevent it from becoming outdated.

content/en/docs/components/workspaces/user-guides/spark-connect.md
tariq-hasan: Actually it turns out that this is a Spark-on-K8s issue that can only be solved in upstream Spark. There is a Jira for this dating back to 2023: https://issues.apache.org/jira/browse/SPARK-42411. I have added in a comment in there describing the current issue.

On the operator side though this has been fairly long-running problem dating back to 2020 at the very least: https://github.com/kubeflow/spark-operator/pull/1050, https://github.com/kubeflow/spark-operator/issues/1124, https://github.com/kubeflow/spark-operator/issues/1136, https://github.com/kubeflow/spark-operator/issues/1148, https://github.com/kubeflow/spark-operator/issues/1344, https://github.com/kubeflow/spark-operator/pull/1439, https://github.com/kubeflow/spark-operator/issues/1920.

On the flip side I did also create an issue in the operator repo to add a flag that enables Spark pod labels/annotations to be configured by the platform admin: https://github.com/kubeflow/spark-operator/issues/3148. There's precedent for this in KFP as well: https://github.com/kubeflow/pipelines/issues/12015.

In retrospect though the mechanism by which we layer in the Istio workaround likely belongs in the [community distribution](https://github.com/kubeflow/community-distribution/blob/010e45e38649625ce4121ddd070f43c3dac0ba9c/common/istio/istio-install/base/patches/istio-sidecar-injector-patch.yaml#L148C10-L148C29).

The operator should not be owning a platform-specific policy. That would be the concern of the platform distribution I would think.

https://github.com/kubeflow/spark-operator/issues/3148 would possibly still be relevant for other use cases though - since an admin-controlled mechanism for attaching validated labels/annotations to Spark workload pods would possibly count as being useful.

content/en/docs/components/workspaces/operator-guides/spark-connect.md
tariq-hasan: The version compatibility gap is being fixed within the switch from `pyspark-connect` to `pyspark-client`: https://github.com/kubeflow/sdk/pull/795.

vikas-saxena02:
> I think you can shorten it a lot and ease the maintenance if you just link to https://github.com/kubeflow/community-distribution/blob/master/.github/workflows/workspaces_spark_connect_test.yaml
> 
> We do the same for kserve. just the high level and a link to constantly tested automation to reduce the maintenance and prevent it from becoming outdated.

@juliusvonkohout can you please share the link to the documentation for kserve and I will align it to the same.

juliusvonkohout:
https://www.kubeflow.org/docs/ecosystem/kserve/ but we actually need to clean it up. The kswrve documentation also looks heavily outdated. Feel free to also raise a PR to clean up kswrve. Another reference is https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/

